PR #231, opened 2019-06-03
Clojure/JVM API

borkdude:
- [x] add tests for EDN + JSON output
- [x] add test for preventing summary from printing
- [x] add tests for clj-kondo.core API functions
  - [x] add test for summary
- [x] move summary also into core fns?
- [x] docstrings for public API functions
- [x] check behavior or `{:output {:progress true}}` in combination with EDN or JSON output
- [x] update "running on the JVM" docs
- [x] update configuration docs, documenting output

### DIFF
--- a/src/clj_kondo/main.clj
+++ b/src/clj_kondo/main.clj
@@ -1,6 +1,7 @@
 (ns clj-kondo.main
   {:no-doc true}
   (:gen-class)
+  (:refer-clojure :exclude [run!])
   (:require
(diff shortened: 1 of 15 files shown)


src/clj_kondo/main.clj
```diff
@@ -1,58 +1,25 @@
 (ns clj-kondo.main
   {:no-doc true}
   (:gen-class)
+  (:refer-clojure :exclude [run!])
```
vemv: (Doesn't look necessary?)

PR #232, opened 2019-06-03
For your information

simon-brooke:
Not suggesting you merge this, but it might be useful to you.

The EDN format I'd first thought of doesn't allow two reports from the same tool regarding the same line, so this needs a bit of rethink yet.


borkdude:
@simon-brooke while I think the idea of a linter aggregator is good, the output to the standard format probably doesn't belong into this repo, but in the linter aggregator repo

simon-brooke:
Fair

PR #326, opened 2019-07-08
Add information on how to get the version of clj-kondo in use

danielcompton:
This makes it more likely that people will try and find the right version.


borkdude:
Thanks!

Issue #342, opened 2019-07-13
Unresolved symbol with :refer :all


l3nz:
**version**

[clj-kondo "2019.07.05-alpha"]

**platform**

JVM library, [org.clojure/clojure "1.9.0"]

**problem**

I have a require with `[compojure.core :refer :all]`(I know it's not elegant, but it's allowed). When I call any method, I get it as:

```
src/kpl/http/core.clj:321:4: info: unresolved symbol defroutes
src/kpl/http/core.clj:323:6: info: unresolved symbol GET
src/kpl/http/core.clj:326:6: info: unresolved symbol POST 
```

**expected behavior**

Either it should raise a linting error because I use a `:refer :all` as bad practice, or it should avoid complaining of undefined symbols.

borkdude:
@l3nz clj-kondo uses the `ns` form to resolve symbols, so when you use `:refer :all` it doesn't know these symbols. Given that ClojureScript doesn't support  `:refer :all` and the Clojure style guide recommends `:as` or `:refer [...]` over `:refer :all`, I recommend using those forms. Clj-kondo won't support resolving symbols using `:use` or `:refer :all`.

l3nz:
Makes sense, but in this case it should raise it as an issue (`not parsing because you are a dunce who uses :refer :all`) not fail on non-imported symbols 😄

borkdude:
@l3nz Good idea. I was going to fix this under the umbrella of "how to ns": https://github.com/borkdude/clj-kondo/issues/229.

borkdude:
@l3nz I've decided to give it a try anyway. Now it looks like this:

![Screenshot 2019-07-21 21 31 06](https://user-images.githubusercontent.com/284934/61596072-1511a980-abff-11e9-9326-87b9fb211d25.png)

You can try this with commit https://github.com/borkdude/clj-kondo/commit/92e4247eed9585969d4da348c0e90625b2ceae98.

borkdude:
TODO:
- [ ] fix order (using foo before definition should be unresolved)
- [ ] write test for cljc (using slurp in cljc should be unresolved because of cljs)

borkdude:
@l3nz Final result:

<img width="1122" alt="compojure-config" src="https://user-images.githubusercontent.com/284934/61795274-4d91cd00-ae23-11e9-982a-12869712f16c.png">

l3nz:
[...]

PR #365, opened 2019-07-23
[#342] symbols referred with refer :all are not reported as unresolved when found in cache

borkdude:


borkdude:
- [ ] linters.clj comment: ;; we could be using a function defined with in-ns, so we should be looking at the filename (or top-declaring ns)

borkdude:
TODO: 
- [ ] come up with some compression schema for file names as transit does not cache these

borkdude:
TODO: 
- [ ] using the filename isn't very reliable: when editing with flycheck/emacs, the filename becomes `<stdin>`...
Using the top-level namespace name might be better, as this also catches the same-ness of the file

PR #391, opened 2019-08-03
Correctly analyze `clojure.core/when-first` macro

volrath:
Hey @borkdude, here's the tiny fix we talked about yesterday at heartofclojure. I wanted to add some sort of test for it but I don't know where would be the best place to do so. Maybe I just add this line to `corpus/unresolved_symbol.clj`:

```clj
(when-first [a [1 2 3]] a)
```

wdyt?

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -908,7 +908,7 @@
           (analyze-like-let ctx expr)
           letfn
           (analyze-letfn ctx expr)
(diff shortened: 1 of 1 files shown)


borkdude:
@volrath I would add the test at the end of `unresolved-symbol-test` and the list of `(is (empty? ...))` cases:

https://github.com/borkdude/clj-kondo/blob/master/test/clj_kondo/main_test.clj#L1909

Since I already merged this PR (a bit too fast), I'll add the test myself.

PR #398, opened 2019-08-04
Support PATH native-image on *nix-likes

sogaiu:
Attempt to use native-image from PATH on *nix-like systems.

Tested locally on machine with native-image on PATH.  I don't have a local setup to fully test the other path -- where there is no native-image on PATH.

### DIFF
--- a/script/compile
+++ b/script/compile
@@ -1,21 +1,27 @@
 #!/usr/bin/env bash
 
-if [ -z "$GRAALVM_HOME" ]; then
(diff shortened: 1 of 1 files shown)


borkdude:
@sogaiu Maybe also update the Windows script?

sogaiu:
@borkdue I can try, but I'm even less proficient at .bat files.

borkdude:
@sogaiu OK, I'll merge it like this.

PR #401, opened 2019-08-06
Add default linter for `clojure.test.check.generators/let`

volrath:
@borkdude Not sure if this is interesting for you to add to the default configuration, but just in case :) feel free to close it if you prefer not to provide a default linter for `clojure.test.check.generators/let`

In the current implementation:

```bash
$> clj-kondo --lint - << EOF
(ns test
  (:require [clojure.test.check.generators :as gen]))
(gen/let [x 1] x) 
EOF
<stdin>:3:11: info: unresolved symbol x
linting took 24ms, errors: 0, warnings: 0
```

This PR removes the false positive.


borkdude:
@volrath Thanks!

PR #404, opened 2019-08-07
Fixbug editor integration with spacemacs

tanrax:


borkdude:
According to spacemacs users in #Slack it works for them without this line. I'll wait for confirmation that this line is really needed.

borkdude:
As several people have confirmed, this additional line wasn't needed for them, so I'll close this for now. Might someone need this in the future, we can re-open.

PR #418, opened 2019-08-10
[WIP] Propagate error locations of read errors

yuhan0:
Note: this is only a partial solution, and only displays error location for invalid keywords.
I'm not familiar enough with the codebase or implementation details to know how to proceed, perhaps some pointers would be helpful.
Or simply treat this PR as reference if it's easier for someone else to pick it up themselves :)

What I found was that the tools.reader exceptions have optional `:line` and `:col` keys attached to them, but it seems like only invalid keywords actually produce this information, due to some difference between `IndexingPushbackReader` and `PushbackReader`.

Also, read-time errors aren't all thrown from the same place, most come from clojure.tools.reader but some like "unexpected EOF" are thrown from `rewrite-clj`.

### DIFF
--- a/corpus/read_error/EOF.clj
+++ b/corpus/read_error/EOF.clj
@@ -1,0 +1,1 @@
+(ns read-error.EOF
--- a/corpus/read_error/error.clj
+++ b/corpus/read_error/error.clj
@@ -1,1 +1,0 @@
-(ns read-error.error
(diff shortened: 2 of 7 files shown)


marcomorain:
This is fixed in #534 534

borkdude:
Superseded by #534.

PR #425, opened 2019-08-20
docs: mention NixOS package

### LINKED ISSUE #414: Publish uberjar
Thanks for writing clj-kondo, I find it really useful.

I'm trying to create a package for [NixOS](https://nixos.org/), compiling it with GraalVM, and it would be helpful if for the next release you can publish the uberjar on github.

That's because on NixOS, during the package build phase, the build process has no network access (and lein cannot connect to the maven repositories to download the dependencies). There are workarounds, but the simplest solution will be to use the uberjar.

jlesquembre:
The package I wrote for nixos got merged, now it is part of the official packages :)
I just added some notes to the docs.

Closes #414

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -57,6 +57,14 @@ To upgrade:
 
     yay -S clj-kondo-bin
 
+## NixOS
+
+`clj-kondo` is available in the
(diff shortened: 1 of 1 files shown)


borkdude:
Thank you so much!

PR #467, opened 2019-09-23
Add alias / :as to analysis data

SevereOverfl0w:
This would be useful for:

- Auto-inserting the require based on existing aliases (e.g. str/ should
insert clojure.string)
- Ensuring a consistent use of aliases in a project.

### DIFF
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -27,6 +27,7 @@ The analysis output consists of a map with:
    - `:filename`, `:row`, `:col`
    - `:from`: the namespace which uses
    - `:to`: the used namespace
+   - `:as`: the alias of namespace, if used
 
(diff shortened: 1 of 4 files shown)


borkdude:
Thanks!

Can you move the alias down as an argument to `assoc-some`? That way it doesn't end up in the EDN if it's nil. Can you also update the tests so this will be tested? The tests are located here: 
https://github.com/borkdude/clj-kondo/blob/master/test/clj_kondo/analysis_test.clj.

SevereOverfl0w:
Made requested changes.

SevereOverfl0w:
@borkdude you also need to update the test that checks that there's no :as key if you don't use an alias (or remove the test, I'm not sure about it)

borkdude:
Done!

PR #488, opened 2019-10-01
More types for clojure.core

dancek:
- add a bunch of specs for simple functions
- add comments for all functions that are still missing specs

---

Any comments are welcome. I thought it a good idea to add comments for everything so you don't have to guess what functions are missing, and editing becomes a tiny bit easier.

I plan to write more specs/types, but don't expect all of them from me. :)

### DIFF
--- a/src/clj_kondo/impl/types/clojure/core.clj
+++ b/src/clj_kondo/impl/types/clojure/core.clj
@@ -16,6 +16,9 @@
 (def any->boolean {:arities {1 {:args [:any]
                                 :ret :boolean}}})
 
+(def any->any {:arities {1 {:args [:any]
+                            :ret :any}}})
+
 (def number->number {:arities {1 {:args [:number]
                                   :ret :number}}})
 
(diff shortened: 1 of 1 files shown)


borkdude:
Awesome job! 😍

The `any->any` doesn't really add any value and will just make kondo a bit slower, because as soon as there is a spec, kondo starts collection type information for the arguments. If there is no spec, this will be skipped. So I think it makes sense to remove that and replace it with a comment that a spec for that function is left out on purpose.

Another minor thing, the `{:op :rest}` for matching varargs doesn't have to be in a separate vector.

dancek:
Great, thanks for merging!

I was thinking something along these lines when writing `any->any`:

1. At this point the type config format is still unstable.
2. Thus it's smart to have common function types as variables so they're easy to see and edit.
3. It's useful to mark even those types we can't really do anything useful with.

Using `nil` for unuseful function types is good at this point (I wasn't sure if that would work). As for `any->any`, it does give the information that the function has arity 1, but something like `a->a` (returns same type, like `with-meta` or `identity`) may be better even if the config doesn't currently support checking for that.

borkdude:
@dancek `a->a` is currently possible but only in code, not in the config yet. For examples of this look at `assoc` and `into`, they return the same type as the first arg or coll if the type of the first arg was unknown. Actually this should be made more strict, since the first arg should at least be of type coll, e.g. we should not accept string there.

PR #493, opened 2019-10-02
updated intellij integration instructions

chorin1:
1. `show console` should be on `never`, otherwise the terminal pops up after every file change.
2. added a custom config usage example - should use `"`. using `' ` as the readme suggests makes clj-kondo stop working with intellij.
3. set `level` to `global` to enable the watcher for all future projects.
4. `scope` should be `current file` or `project files`.


borkdude:
Thanks!

PR #495, opened 2019-10-02
Types for arithmetic operations

dancek:
Here's more types for clojure.core.

I'm not sure if the haskell-style names for the type variables are a good idea. I'm willing to rewrite this commit if someone comes up with better names.

Also ideas for testing the correctness of these type definitions are welcome. I guess running clj-kondo against large known-good codebases would be one idea.

### DIFF
--- a/src/clj_kondo/impl/types/clojure/core.clj
+++ b/src/clj_kondo/impl/types/clojure/core.clj
@@ -16,12 +16,15 @@
 (def any->boolean {:arities {1 {:args [:any]
                                 :ret :boolean}}})
 
-;; this is set to nil on purpose, for performance reasons.
-(def any->any nil #_{:arities {1 {:args [:any]
-                                  :ret :any}}})
+;; arity-1 function that returns the same type
+(def a->a {:arities {1 {:args [:any]}}
+           :fn #(:tag (first %))})
 
 (def number->number {:arities {1 {:args [:number]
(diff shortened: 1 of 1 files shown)


borkdude:
> I'm not sure if the haskell-style names for the type variables are a good idea

Since this is not exposed, it's not a problem to me.

Clj-kondo is ran against a large number of deps (Clojure and ClojureScript + some other libs) on CI.
You can check that by clicking "Show all checks" and then for example "linux" / Details and then Performance report. Don't be confused by the name, it actually spits out a large number of findings for the mentioned libraries. I am using this to check performance regressions, but also compare newly generated linter findings with the same output on master.

When I do that for this PR, I find a new warning for:

https://github.com/clojure/clojure/blob/653b8465845a78ef7543e0a250078eea2d56b659/src/clj/clojure/string.clj#L125

The warning reads:

> clojure/string.clj:125:28: warning: Expected: number, received: character or nil.

So probably we have to change the type for the arg to int to `#{:number char}`.

I'm doing this comparison using Beyond Compare, just FYI, there might be other/free tools that you can also use for this.

dancek:
Good find! I edited the commit, this should be better.

I'll see if I can run the CircleCI scripts locally from now on. For changes that only add types there should not be any differences in warnings, so a simple diff should be good for my purposes at this point. Thanks for the tip!

PR #517, opened 2019-10-10
Error messages should point to the opening paren, not EOF

marcomorain:
Change the vendored copy of rewrite-clj to throw structured data rather
than raw Exception.

Change the analyzer to deal with the structured data rather than the raw
Exceptions, which allows us to remove some regex matching and parsing.

Change the rewrite-clj parser to wrap exceptions when it detects
unmatched delimiters. This allows us to throw multiple errors from a
single error.

Add test for same.

### DIFF
--- a/parser/clj_kondo/impl/rewrite_clj/parser/utils.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/parser/utils.clj
@@ -25,12 +25,10 @@
 (defn throw-reader
   "Throw reader exception, including line/column."
   [reader fmt & data]
-  (let [c (r/get-column-number reader)
-        l (r/get-line-number reader)]
-    (throw
-     (Exception.
-      (str (apply format fmt data)
-           " [at line " l ", column " c "]")))))
+  (let [f {:row (r/get-line-number reader)
+           :col (r/get-column-number reader)
+           :message (apply format fmt data)}]
+    (throw (ex-info "Syntax error" {:findings [f]}))))
 
 (defn read-eol
   [reader]
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1352,29 +1352,24 @@
 
 ;;;; processing of string input
 
-(defn- ->finding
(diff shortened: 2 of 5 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -1372,26 +1400,9 @@
...
-                           :filename filename
-                           :col 0
-                           :row 0
-                           :type :syntax
-                           :message (str "can't parse "
-                                         filename ", "
-                                         (or m (str e)))}))]}))
+      (if dev?
+        (throw e)
+        (->findings e filename)))
```
marcomorain: I will pull this `->findings` refactor out into a separate PR to make this diff smaller.

parser/clj_kondo/impl/rewrite_clj/parser/utils.clj
```diff
@@ -22,13 +22,19 @@
...
 
+(defn ex-finding [row col message cause-ex]
+  (ex-info message {:type :syntax
+                    :col col
+                    :row row
+                    :message (clojure.string/trim message)}
+           cause-ex))
+
 (defn throw-reader
   [reader & msg]
```
marcomorain: So it turns out that `rewrite-_clj/parser` has two `throw-reader` functions, with slightly different signatures. I make one of them call the other one in this change. It would be better to make this change in an isolated PR.

Or maybe we want to minimize changes to `rewrite-clj` and we can make the `ex-data` change in two places.

marcomorain:
This will be much neater once re-based on top of #523

marcomorain:
@borkdude this is ready for discussion now.

The main point that I'd like to work through is whether we should return a single error from `rewrite-clj` or multiple for the types of errors involved.

For a `(defn []` with no closing `)`, should there be a single error saying that the openeing `(` was unmatched, or should there be two messages, the second saying "unexpected EOF", (the second message is is the current error message produced by rewrite-clj).

marcomorain:
This approach doesn't work well, I'll need to think some more about this problem.

borkdude:
@marcomorain I'm favour of emitting two errors. So the old EOF one + one extra.

PR #523, opened 2019-10-11
Consolidate throw-reader functions

marcomorain:
rewrite-clj has two `throw-reader` functions, with slightly different
behaviour. One expectects a list of strings, the other expects a format
string and matching arguments.

Replace both functions with a single one.

Add a full regression test that has example source code that excercises
all call sites of throw-reader, to ensure that this refactor does not
change behaviour.

### DIFF
--- a/test/clj_kondo/impl/parser_test.clj
+++ b/test/clj_kondo/impl/parser_test.clj
@@ -1,7 +1,7 @@
 (ns clj-kondo.impl.parser-test
   (:require [clj-kondo.impl.parser :as parser :refer [parse-string]]
             [clj-kondo.impl.utils :as utils]
-            [clojure.test :as t :refer [deftest is]]))
+            [clojure.test :as t :refer [deftest is are]]))
 
 (deftest omit-unevals-test
   (is (zero? (count (:children (parse-string "#_#_1 2"))))))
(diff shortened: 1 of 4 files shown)


test/clj_kondo/impl/parser_test.clj
```diff
@@ -22,8 +22,31 @@
...
+  ;; the `clj-kondo.impl.rewrite-clj.parser.utils/throw-reader` function.
+  ;; This allows us to test for regressions when that function is refactored.
+  (are [source message] (= message (parse-error source))
+    "[" "Unexpected EOF. [at line 1, column 2]"
+    "[}" "Unmatched delimiter: } [at line 1, column 2]"
+    "#" "Unexpected EOF. [at line 1, column 2]"
+    ":" "unexpected EOF while reading keyword. [at line 1, column 2]"
+    "\"" "Unexpected EOF while reading string. [at line 1, column 2]"
+    "#?" ":reader-macro node expects 1 value. [at line 1, column 3]"
+    "#:" "Unexpected EOF. [at line 1, column 3]"))
```
marcomorain: It took me too long to find what source code would produce this error message :)

marcomorain: https://github.com/borkdude/clj-kondo/blob/97b3ddcdbc63e70faa526641ba066ae0db174725/parser/clj_kondo/impl/rewrite_clj/reader.clj#L76

PR #527, opened 2019-10-12
Better error messages for syntax errors

### LINKED ISSUE #514: Show a proper error message for unmatched Parentheses
`clj-kondo` warns about unmatched parentheses in the same way as Clojure core – the error referers to unexpected EOL at the end of the file, instead of highlighting the brackets that don't match.

I'd like `clj-kondo` to emit 2 errors when the parens are unbalanced.

I've not considered any - feedback welcome.

**Additional context**
I've just spent 5 minutes trying to find the reason that code would not compile.

### Example 1: Simple missing paren

```Clojure
(defn function-with-no-closing-brace []
  (inc 5)
```

Actual output:
```
server_test.clj:65:1: error: Unexpected EOF.
```

This error message gives a location at the end of the file, _not_ where the unmatched paren is located. The error message does not mention that a paren is unmatched, it says that it got an "unexpected EOF".

The red squiggle in VSCode produced from this error is equally unhelpful:

image

Expected Output:

```
server_test.clj:60:1: error:Unmatched opening-paren `(`
server_test.clj:65:1: error: Reached end-of-file while trying to read the `)` to close the opening paren on line 60 column 1
```

### Example 2: Badly nested parens of different types

```Clojure
(defn make-basic-service
  []
  (grpc-server/gen-service
    (UserServiceGrpc/getServiceDescriptor)
    UserServiceGrpc$UserServiceImplBase
    {:get-user (fn [req]
                 (if (= {:account-id (UUID/fromString "8f1e7316-293e-4946-99ee-c5ea74e01a5c")
                         :user-id (UUID/fromString "7c662690-cd61-4e37-939b-4727548d95b4")}
                        (select-keys req [:account-id :user-id]))
                   (assoc (select-keys req [:account-id :user-id]) :display-name "cookie monster")
                   (throw (-> (Status/NOT_FOUND)
                              (.withDescription "user not found")
                              .asRuntimeException))))
     :always-fail (fn [_req]
                    (throw (ex-info "a specific problem" {:status :resource-exhausted})))))
```

Output in VSCode:

image

Note that the red error highlights a closing paren, which makes it look like there are too many closing `)`, which is not the case. The problem is with the `{` on line 20.

Actual output:
```
server_test.clj:29:90: error: Unmatched delimiter: )
```

Expected Output:

```
server_test.clj:20:4: error: Unmatched opening `{`
server_test.clj:29:90:4: error: Unmatched parentheses - found a closing `)` when expecting to match the `{` opened on line 20:4
```
```

### PR DESCRIPTION
**Pull request by marcomorain**
Fixes #514

The error messages from the vendored copy of rewrite-clj are not as
helpful as they could be. I often have a hard time finding the location
of a mismatched bracket, since the error message refers to the EOF at
the end of the file, and not the opening paren.

To fix this we allow rewrite-clj to throw exceptions containing a
sequence of errors.

Changes:

- Make rewrite-clj throw ex-info instead of Exception. The ex-data
contains a :findings key, which is a sequence of findings in the format
expected by clj-kondo

- Change rewrite-clj.parser.core/*delimiters* to keep track of a tuple
of [open, close, row, col] rather than just the expected closing
delimiter. The extra three fields allow is to construct more informative
error messages when a parse error occurs.

- If an EOF is encountered when parsing a delimited list {}, [] or (),
throw an exception with two error messages - one for the opening
delimiter, saying that it is unmatched, and one for the end of file
saying that a closing delimiter was expected.

- If a mismatched delimiter is found when parsing a list ({}, [], ())
then throw two error messages - one for the opening bracket and one for
the closing bracket. In both cases, refer to both the line on which the
token is found, and the line on which the corresponding token is found.

Before:

> Unexpected EOF

After:

> Found an opening [ with no matching ]
> Expected a ] to match [ from line 1

Before:

> Unmatched delimiter: }

After:

> Mismatched bracket: found an opening [ and a closing } on line 1
> Mismatched bracket: found an opening [ on line 1 and a closing }

### DIFF
--- a/parser/clj_kondo/impl/rewrite_clj/parser/core.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/parser/core.clj
@@ -13,13 +13,20 @@
 ;; ## Base Parser
 
 (def ^:dynamic ^:private *delimiter*
+  "Record information about the pair of delimiter that are being parsed.
+  This keeps track when parsing matching pairs of `{}`, `[]`, `()`.
+  Store a tuple of 4 values: [open (char), close (char), row (int), col (int)]
+  - The first item is the opening character, `{`, `[` or `(`.
+  - The second item is the closing character, `}`, `]` or `)`.
+  - The third item is the row on which the opening bracket was found.
+  - The fourth item is the column on which the opening bracket was found."
   nil)
 
 (defn- dispatch
   [c]
-  (cond (nil? c)               :eof
-        (reader/whitespace? c) :whitespace
-        (= c *delimiter*)      :delimiter
+  (cond (nil? c)                   :eof
+        (reader/whitespace? c)     :whitespace
+        (= c (second *delimiter*)) :delimiter
         :else (get {\^ :meta      \# :sharp
                     \( :list      \[ :vector    \{ :map
                     \} :unmatched \] :unmatched \) :unmatched
@@ -38,11 +45,12 @@
 ;; # Parser Helpers
 
 (defn- parse-delim
-  [reader delimiter]
-  (reader/ignore reader)
-  (->> #(binding [*delimiter* delimiter]
-          (parse-next %))
-       (reader/read-repeatedly reader)))
+  [reader open-delimiter close-delimiter]
+  (let [{:keys [row col]} (reader/position reader :row :col)]
+    (reader/ignore reader)
+    (->> #(binding [*delimiter* [open-delimiter close-delimiter row col]]
+            (parse-next %))
+         (reader/read-repeatedly reader))))
 
 (defn- parse-printables
   [reader node-tag n & [ignore?]]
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by marcomorain** [comment]
@borkdude OK, I'm happy with this now, ready for a proper review 👌

**Comment by marcomorain** [comment]
## Before
image

## After

image

**Comment by marcomorain** [comment]
From this last image it looks like there is a bug with the location of the first error – `:get-user` is highlighted. I think I need to take the position from the `reader` _before_ the `{` is read.

**Comment by borkdude** [comment]
```
$ clj -A:clj-kondo --lint - <<< '{:foo'
<stdin>:1:1: error: Found an opening { with no matching }
<stdin>:2:1: error: Expected a } to match { from line 1
linting took 95ms, errors: 2, warnings: 0
```
Why is the second message saying line 2?

What about this case?
```
$ clj -A:clj-kondo --lint - <<< '{:foo}}'
<stdin>::: error: Mismatched bracket: found an opening null and a closing } on line 1
<stdin>:1:7: error: Mismatched bracket: found an opening null on line null and a closing }
linting took 50ms, errors: 2, warnings: 0
```

**Comment by marcomorain** [comment]
The first case there is an existing bug in `rewrite-clj` I think:

```
marc@blaster ~/dev/marcomorain/clj-kondo $ clj-kondo --version
clj-kondo v2019.10.04-alpha
marc@blaster ~/dev/marcomorain/clj-kondo $ echo '{:foo' | clj-kondo --lint -
<stdin>:2:1: error: Unexpected EOF.
linting took 4ms, errors: 1, warnings: 0
```

The second one is a bug in my code, thanks for finding that. I'll take a look now.

marcomorain:
I fixed the issue with the dangling closed paren last night, and added a test 👌

PR #529, opened 2019-10-12
EOF reporting off by one error

marcomorain:
Expose a bug when reporting a premature EOF. It looks like the parser
skips whitespace greedily and throws the error at the wrong place.


marcomorain:
@borkdude this test exposes the EOL "line 2" bug that you found.

marcomorain:
this isn't a bug really

PR #534, opened 2019-10-14
Use an indexing reader when parsing EDN

marcomorain:
By passing an indexing reader to the edn parser, line and column
information is produced in the error messages.

    clj-kondo.impl.toolsreader.v1v2v2.clojure.tools.reader.impl.errors/throw-ex

Remove the string formatting here. All of the data required by
->findings is contained in the ex-data. If we don't remove this string
formatting, then the error message shown to the user will be prepended
with `[line x, col y] `.

    clj-kondo.impl.analyzer/->findings

Now that tools.reader is throwing data, we need to update `->findings`
to be able to read that data correctly.

Add unit test to ensure that line data comes out of the parser, and an
integration to ensure that `->findings` works.

### DIFF
--- a/test/clj_kondo/impl/parser_test.clj
+++ b/test/clj_kondo/impl/parser_test.clj
@@ -31,10 +31,11 @@
     (parse-string source)
     nil
     (catch Exception e
-      (if-let [findings (:findings (ex-data e))]
-        (for [{:keys [row col message]} findings]
-          [message row col])
-        [[(.getMessage e) 0 0]]))))
+      (let [{:keys [findings line col]} (ex-data e)]
+        (if findings
+          (for [{:keys [row col message]} findings]
+            [message row col])
+          [[(.getMessage e) line col]])))))
 
 (deftest parse-string-test
   ;; This test has every syntax error that can cause rewrite-clj to throw using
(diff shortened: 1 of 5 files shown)


test/clj_kondo/impl/parser_test.clj
```diff
@@ -50,7 +51,7 @@
     ":"  [["unexpected EOF while reading keyword." 1 2]]
     "\"" [["Unexpected EOF while reading string." 1 2]]
     "#?" [[":reader-macro node expects 1 value." 1 3]]
-    "[1..1]" [["Invalid number: 1..1." 0 0]]
+    "[1..1]" [["Invalid number: 1..1." 1 4]]
```
marcomorain: @borkdude this fixes that bug I mentioned in Slack.

borkdude:
@marcomorain Great work! Since we're now changing the implementation of tools.reader, I think it makes sense to rename the folder from "inlined" to "reader" since it's longer just an inlined version of tools.reader version such and such. Also we probably need to remove the version reference. 
Maybe it makes sense to scan through the tools.reader codebase to check if there's anything worth upgrading since that version.

borkdude:
@marcomorain Also we could remove any code from the inlined tools.reader that's no longer needed by clj-kondo.

borkdude:
@marcomorain Could also do those things (myself) in a different PR if you prefer that.

marcomorain:
@borkdude I'd rather leave the moving of the namespace to you – I'm not too familiar with the way that you've vendored/inlined the different sub-projects.

Issue #533, opened 2019-10-14
CLJS namespace as constructor causes warning for unused namespace


dancek:
NOTE: I'm planning to submit a PR fixing this, and from a discussion in Slack was already pointed in the right direction. I'm just posting this bug report in case I forget (ie. if you don't hear from me in a week I'm probably not working on this).

Related to #446; a similar fix should work (and this should also only be for string namespaces).

---

**version**

master

**platform**

any

**problem**

Using an imported CLJS namespace as a constructor doesn't count as usage.

**repro**

```
echo '(ns foo (:require ["bar" :as Bar])) (Bar.)' | clj-kondo --lint -
```

**expected behavior**

No warnings.

**actual behavior**

```
<stdin>:1:20: warning: namespace bar is required but never used
```

borkdude:
For reference, this is where and how the string requires is checked:

https://github.com/borkdude/clj-kondo/blob/154dfd7f52f6362c5d8cc7ceae84c76a4d212204/src/clj_kondo/impl/namespace.clj#L209

borkdude:
Fixed with https://github.com/borkdude/clj-kondo/commit/4c52ab5b131b60fb3939afebf604918c43e601c1. Thank you @dancek.

PR #537, opened 2019-10-15
[#533] macroexpand (Obj.) to (new Obj)

dancek:
This fixes the issue #533 with CLJS when module export is a constructor that
is called in the short-hand constructor notation. It might also be
useful later for analyzing Java interop.

---

Rationale: just hacking special logic when cljs calls a function with a dot in the end didn't feel right, when Clojure always expands these calls to `new`.

However, this breaks tests at the moment. I'll fix them but just wanted to submit this code for comments.

### DIFF
--- a/corpus/cljs_ns_as_as_object.cljs
+++ b/corpus/cljs_ns_as_as_object.cljs
@@ -1,5 +1,7 @@
 (ns foo
   (:require [reagent.core :as r]
-            ["apsl-react-native-button" :as NativeButton]))
+            ["apsl-react-native-button" :as NativeButton]
+            ["constructor-export" :as ConstructorExport]))
 
 (def button (r/adapt-react-class NativeButton))
+(def obj (ConstructorExport.))
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -917,7 +917,19 @@
   (swap! calls-by-id assoc id call)
   nil)
 
-(defn analyze-constructor [ctx expr]
+(defn analyze-constructor* [ctx ctor-node children]
+  (analyze-usages2 (ctx-with-linter-disabled ctx :unresolved-symbol) ctor-node)
+  (analyze-children ctx children))
+
+(defn analyze-constructor
+  "Analyzes (new Foo ...) constructor call."
+  [ctx expr]
+  (let [[_ ctor-node & children] (:children expr)]
+    (analyze-constructor* ctx ctor-node children)))
+
+(defn analyze-dot-constructor
+  "Analyzes (Foo. ...) constructor call."
+  [ctx expr]
   (let [[ctor-node & children] (:children expr)
         ctor (:value ctor-node)
         ctor-name (name ctor)
@@ -926,8 +938,7 @@
                       symbol)
         ctor-node (with-meta (utils/token-node ctor-name)
                     (meta ctor-node))]
-    (analyze-usages2 (ctx-with-linter-disabled ctx :unresolved-symbol) ctor-node)
-    (analyze-children ctx children)))
+    (analyze-constructor* ctx ctor-node children)))
 
 (defn analyze-call
   [{:keys [:top-level? :base-lang :lang :ns :config] :as ctx}
@@ -945,7 +956,7 @@
         (resolve-name ctx ns-name full-fn-name)]
     (if (and unresolved?
              (str/ends-with? full-fn-name "."))
-      (analyze-constructor ctx expr)
+      (analyze-dot-constructor ctx expr)
       (let [[resolved-as-namespace resolved-as-name _lint-as?]
             (or (when-let
                     [[ns n]
@@ -1042,6 +1053,7 @@
               (if top-level? (analyze-require ctx expr)
                   (analyze-children ctx (next (:children expr))))
               if (analyze-if ctx expr)
+              new (analyze-constructor ctx expr)
               ;; catch-all
               (case [resolved-as-namespace resolved-as-name]
                 [schema.core fn]
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -917,6 +917,18 @@
   (swap! calls-by-id assoc id call)
   nil)
 
+(defn analyze-constructor [ctx expr]
+  (let [[ctor-node & children] (:children expr)
+        ctor (:value ctor-node)
+        ctor-name (name ctor)
+        ctor-name (-> ctor-name
+                      (subs 0 (dec (count ctor-name)))
+                      symbol)
+        ctor-node (with-meta (utils/token-node ctor-name)
+                    (meta ctor-node))]
+    (analyze-usages2 (ctx-with-linter-disabled ctx :unresolved-symbol) ctor-node)
+    (analyze-children ctx children)))
+
 (defn analyze-call
   [{:keys [:top-level? :base-lang :lang :ns :config] :as ctx}
    {:keys [:arg-count
@@ -933,7 +945,7 @@
         (resolve-name ctx ns-name full-fn-name)]
     (if (and unresolved?
              (str/ends-with? full-fn-name "."))
-      (analyze-expression** ctx (macroexpand/expand-constructor expr))
+      (analyze-constructor ctx expr)
       (let [[resolved-as-namespace resolved-as-name _lint-as?]
             (or (when-let
                     [[ns n]
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -930,172 +930,175 @@
          unresolved? :unresolved?
          clojure-excluded? :clojure-excluded?
          :as _m}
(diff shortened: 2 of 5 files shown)


dancek:
Right, I thought the tests broke because they use things like `(Exception.)`. Nope, they actually break because they use the `.` and `..` macros.

dancek:
That's actually problematic: Clojure does other macroexpansion first, and only then expands constructor calls ending with a dot. I.e. you can have macros ending with a dot with no problem.

dancek:
I guess only macroexpanding unresolved symbols ending with a dot is good enough. I'll try that.

dancek:
I'm too used to Gerrit and didn't realize you lose patchsets by force-pushing in Github. Here's my first try in case it's interesting. https://github.com/dancek/clj-kondo/commit/5346c5dce963e1614e70051544639ca4142a020a

borkdude:
@dancek Nice solution! There might be some code elsewhere in the codebase which ignores symbols ending with dots to prevent false positive unresolved symbols. I believe this is in `reg-unresolved-symbol!` in `impl/namespace.clj`. I think that needs to be changed as well.

dancek:
I changed that, and while doing that discovered a bug in my previous implementation: the expanded constructor calls would return `:unresolved? true`. I had to split the long let clause in `analyze-call` to avoid that. Now the program flow is better than in the previous version, but I'm afraid the code for `analyze-call` just became a bit less readable.

borkdude:
Can you maybe stop force-pushing but just use normal commits? I will squash them anyway when accepting the PR. This gives me the option to push a commit myself to a branch when I want to change some details. Are you OK with this?

dancek:
I'll try to remember that. Sorry about disrupting your workflow (I didn't realize). I'm just used to tooling where amending commits is the preferred way (Gerrit) so don't be surprised if I do this again...

borkdude:
No worries :). I'll go over this later today.

Have you checked https://app.circleci.com/jobs/github/borkdude/clj-kondo/6591/parallel-runs/0/steps/0-108 for any differences compared to master for regressions?

dancek:
No, but I'll check.

dancek:
There are differences. I'll try to make sense of each.
```
--- output.txt	2019-10-16 11:01:39.000000000 +0300
+++ output_expected.txt	2019-10-16 11:03:25.000000000 +0300
@@ -411,6 +411,7 @@
 cljs/pprint.cljs:162:1: warning: Unused private var cljs.pprint/get-line
 cljs/pprint.cljs:168:1: warning: Unused private var cljs.pprint/set-max-column
 cljs/pprint.cljs:172:1: warning: Unused private var cljs.pprint/get-writer
+cljs/pprint.cljs:230:1: warning: Unused private var cljs.pprint/logical-block
 cljs/pprint.cljs:230:1: warning: Unused private var cljs.pprint/map->logical-block
 cljs/pprint.cljs:230:1: warning: Unused private var cljs.pprint/->logical-block
 cljs/pprint.cljs:250:10: error: unresolved symbol buffer-blob
@@ -454,9 +455,11 @@
 cljs/pprint.cljs:840:1: warning: Unused private var cljs.pprint/level-exceeded
 cljs/pprint.cljs:881:9: warning: unused binding colnum
 cljs/pprint.cljs:881:16: warning: unused binding colinc
+cljs/pprint.cljs:945:1: warning: Unused private var cljs.pprint/arg-navigator
 cljs/pprint.cljs:945:1: warning: Unused private var cljs.pprint/->arg-navigator
 cljs/pprint.cljs:945:1: warning: Unused private var cljs.pprint/map->arg-navigator
 cljs/pprint.cljs:989:1: warning: Unused private var cljs.pprint/map->compiled-directive
+cljs/pprint.cljs:989:1: warning: Unused private var cljs.pprint/compiled-directive
 cljs/pprint.cljs:989:1: warning: Unused private var cljs.pprint/->compiled-directive
 cljs/pprint.cljs:1012:11: warning: use :else as the catch-all test expression in cond
 cljs/pprint.cljs:1037:27: warning: Missing else branch.
@@ -915,8 +918,6 @@
 clojure/browser/repl.cljs:65:4: warning: unused binding conn
 clojure/browser/repl.cljs:65:9: warning: unused binding block
 clojure/browser/repl.cljs:164:40: warning: unused binding name
-clojure/core.clj::: error: unresolved symbol ExceptionInfo
-clojure/core.clj::: error: unresolved symbol LinkedBlockingQueue
 clojure/core.clj:35:17: warning: unused binding &form
 clojure/core.clj:35:23: warning: unused binding &env
 clojure/core.clj:40:18: warning: unused binding &form
@@ -1711,10 +1712,12 @@
 parser/clj_kondo/impl/rewrite_clj/potemkin.clj:71:12: warning: unused binding arglists
 parser/clj_kondo/impl/rewrite_clj/potemkin.clj:94:12: warning: unused binding nspace
 parser/clj_kondo/impl/rewrite_clj/reader.clj:15:3: warning: Misplaced docstring.
+src/clj_kondo/impl/analyzer.clj:26:5: warning: namespace clojure.string is required but never used


 ==== Lint a single file (emulate in-editor usage)


 ==== Launch clj-kondo for each file in project (178)
+src/clj_kondo/impl/analyzer.clj:26:5: warning: namespace clojure.string is required but never used
```

I ran the checks locally with a script like
```
./script/compile
./.circleci/script/performance
grep -v "^\(linting took .*ms\|real\t\|user\t\|sys\t\)" performance.txt > output.txt
diff -u output.txt output_expected.txt
```

dancek:
1. There are constructor uses in  `cljs.pprint` that previously caused _unused private var_ warnings: [logical-block](https://github.com/clojure/clojurescript/blob/v1.10/src/main/cljs/cljs/pprint.cljs#L495), [arg-navigator](https://github.com/clojure/clojurescript/blob/v1.10/src/main/cljs/cljs/pprint.cljs#L953) and [compiled-directive](https://github.com/clojure/clojurescript/blob/v1.10/src/main/cljs/cljs/pprint.cljs#L2565).
2. `error: unresolved symbol ExceptionInfo` and `LinkedBlockingQueue` are new errors and need to be fixed before merging this.
3. `clojure.string` was unused in `clj-kondo.impl.analyzer` but it is now used in this commit, so that's expected.

borkdude:
@dancek:

1. good!
2. the missing locations for these are probably because in the expansion that is now constructed the location metadata is not propagated. I think the default behavior for interop was to not report the symbol as unresolved, since it could come from anywhere on the classpath without using an import.
3. ok

dancek:
There's an explicit import in https://github.com/clojure/clojure/blob/clojure-1.10.0/src/clj/clojure/core.clj#L4772 and I thought imports are required normally. E.g.
```
user=> (ExceptionInfo. "foo" {})
Syntax error (IllegalArgumentException) compiling new at (REPL:1:1).
Unable to resolve classname: ExceptionInfo
```

I'll look at this later today to see if I understand what's happening.

borkdude:
@dancek I'll push a little fix for 2 and explain afterwards, OK?

dancek:
OK.

dancek:
@borkdude oh, now I see. Thanks!

dancek:
Though still
```
$ echo '(import clojure.lang.ExceptionInfo) (ExceptionInfo. "foo" {})' | clj -m clj-kondo.main --lint -
<stdin>:1:38: error: unresolved symbol ExceptionInfo
```

borkdude:
@dancek Wasn't ready yet. I've now opted to skip the macro-expansion because we can just directly analyze the expression and maybe that is a little better for performance?

dancek:
Yeah, that is probably a better idea. Am I right that the reason to disable checks for unresolved symbols is that imports aren't currently fully handled? (I noticed that `(import)` gives the error while `(ns (:import))` doesn't.)

borkdude:
@dancek True, that's a different issue: https://github.com/borkdude/clj-kondo/issues/450

borkdude:
Both:

```
$ clj -A:clj-kondo --lint - <<< '(new Foo 1)'
linting took 122ms, errors: 0, warnings: 0
```

and 

```
$ clj -A:clj-kondo --lint - <<< '(Foo. 1)'
linting took 135ms, errors: 0, warnings: 0
```

now work without warnings.

borkdude:
I'm done pushing, feel free to clean up stuff and add additional test and/or report when you're ready for merge :-).

dancek:
This is now ready for squash+merge IMO. Thanks for helping me out a lot! And sorry for causing quite a lot of work for you with this...

borkdude:
@dancek My pleasure and no worries at all! I think it's good to go back and forth and learn as we go. Your commits inspired my commits and saved me time in the end.

I found one new regression:

```
cljs/repl/node.clj:263:9: warning: Expected: associative collection or nil, received: atom.
```

and I know why this happens. I'll push a small fix.

borkdude:
@dancek I went back the macro approach, since that seems to work more consistently with the type checking.

borkdude:
@dancek Checked the performance.txt for regressions and I only see improvements now for private usage, so I think this is good to go. If you can confirm, I'll squash-merge.

borkdude:
Btw, it would be very cool if we could get this diff for every PR automatically somehow :)

PR #544, opened 2019-10-24
Add clj-kondo icon in svg and png formats

lread:
In case you need to repeat for any reason: I created icon.svg
from the existing logo.svg using Inkscape on linux (Inkscape on
macOS is not a great experience yet). I then used Inkscape's
export to create the 128x128 icon.png.

I followed up with https://www.diffchecker.com/image-diff to
compare the new icon.png to the one I created for Slack. Looked
close enough to me.


borkdude:
Thanks a lot!

PR #565, opened 2019-10-31
Add failing hash-set test

nenadalm:
Hi. There seems to be a problem with `hash-set` function.


borkdude:
@nenadalm Thanks. Pushed a fix to your branch. Will merge when tests succeed.

nenadalm:
thanks.

PR #580, opened 2019-11-04
Add datomic datalog syntax checker.

whilo:
Support Datomic datalog syntax (and static semantics) check for datomic, datascript and datahike.

https://github.com/borkdude/clj-kondo/issues/563

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,6 +1,6 @@
 {:paths ["src" "parser" "resources" "inlined"]
  :deps {com.cognitect/transit-clj {:mvn/version "0.8.313"}
-        io.lambdaforge/datalog-parser {:mvn/version "0.1.0"}
+        io.lambdaforge/datalog-parser {:mvn/version "0.1.1"}
         cheshire {:mvn/version "5.8.1"}}
  :aliases {:clj-kondo
            {:main-opts ["-m" "clj-kondo.main"]
--- a/deps.edn
+++ b/deps.edn
@@ -1,5 +1,6 @@
 {:paths ["src" "parser" "resources" "inlined"]
  :deps {com.cognitect/transit-clj {:mvn/version "0.8.313"}
+        io.lambdaforge/datalog-parser {:mvn/version "0.1.0"}
         cheshire {:mvn/version "5.8.1"}}
  :aliases {:clj-kondo
            {:main-opts ["-m" "clj-kondo.main"]
--- a/project.clj
+++ b/project.clj
@@ -9,7 +9,7 @@
   :source-paths ["src" "parser" "inlined"]
   :dependencies [[org.clojure/clojure "1.9.0"]
                  [com.cognitect/transit-clj "0.8.313"]
-                 [io.lambdaforge/datalog-parser "0.1.0"]
+                 [io.lambdaforge/datalog-parser "0.1.1"]
                  [cheshire "5.8.1"]]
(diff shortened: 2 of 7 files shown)


whilo:
Unfortunately I get an error in one of my projects that I cannot reproduce with the tests atm.

borkdude:
@whilo Thanks!

- Can you move the tests to main_test.clj as do it like the rest of the linters (as an end to end test)?
- About the name of the linter: I might prefer just `:datalog` or `:datalog-syntax` instead of `:invalid-datalog`. Maybe there will be discoveries like: your datalog is valid but it has redundant variables, etc.?
- Could you post the bug + code your are linting?

borkdude:
NOTE: this branch adds about 2.5mb to the resulting binary:

This branch:
18620404 Nov  4 10:24 clj-kondo

2019.11.03:
15938732 Nov  3 20:43 clj-kondo

borkdude:
@whilo With this input:

```
$ cat /tmp/foo.clj
(ns user (:require [datahike.api :refer [q]]))
(q '[:find ?a :where [?a :foo _]] 42)
$ clj-kondo --lint /tmp/foo.clj
/tmp/foo.clj:2:4: error: No matching field found: getClass for class clojure.lang.Symbol
linting took 14ms, errors: 1, warnings: 0
```
I'm seeing an error. This is likely due to a reflective call somewhere. Maybe in the datalog parser?

whilo:
Yes, I have extended the reflection to `clojure.lang.Symbol` and `clojure.lang.Keyword` and it works now like a charm :). The only thing remaining would be to be able to infer datalog data structures outside the query scope, but I guess that is tricky to do with binding and too slow if we hook a pattern matcher into the vector analysis. Anyway, that is already a great progress, because the parser is giving fairly comprehensible error messages including free variables.

whilo:
Is binary size a big deal? I think this is happening because of the full blown parser that we have included and it might be hard to get around that. I will address the other issues.

borkdude:
@whilo I added `#?(:clj (set! *warn-on-reflection* true))` to every namespace in the parser. lein test now prints these warnings:

```
Reflection warning, datalog/parser/impl/util.cljc:57:21 - reference to field getClass can't be resolved.
Reflection warning, datalog/parser/impl/util.cljc:57:21 - reference to field isArray can't be resolved.
```

The relevant code:

```
(defn- #?@(:clj  [^Boolean seqable?]
           :cljs [^boolean seqable?])
  [x]
  (and (not (string? x))
       #?(:cljs (or (cljs.core/seqable? x)
                    (array? x))
          :clj  (or (seq? x)
                    (instance? clojure.lang.Seqable x)
                    (nil? x)
                    (instance? Iterable x)
                    (-> x .getClass .isArray)
                    (instance? java.util.Map x)))))
```
Why isn't `seqable?` from clojure.core used here?

borkdude:
> Is binary size a big deal?

2.5mb isn't that big of a deal but all these things add up in the long run.

If we can reduce it, that would be nice.

Are there ways to break out the checking logic from the parser logic in anyway and make it more light weight, since we don't need the parsing results? I would be fine with inlining the dependency (like rewrite-clj and tools.reader already are).

borkdude:
False positive unused binding:
```
(let [x '...] (q x))
```
False negative duplicate key:
```
(q '{:a 1 :a 2}
```

borkdude:
Binary size seems more reasonable now:

```
-rwxr-xr-x   1 borkdude  staff  16615564 Nov  4 13:45 clj-kondo
```

PR #601, opened 2019-11-11
Adds example for when lint-as clojure.core/-> does not work

johanwiren:
Failing example for lint-as clojure.core/->

    $ clj -A:clj-kondo --lint corpus/lint_as__GT.clj
    corpus/lint_as__GT.clj:11:25: error: lint-as--GT/arity-2 is called with 1 arg but expects 2
    corpus/lint_as__GT.clj:12:25: error: lint-as--GT/arity-2 is called with 1 arg but expects 2
    linting took 111ms, errors: 2, warnings: 0


borkdude:
I think the first call should have two arguments: `(arity-2 :y)`. I took over this example in issue #602, so I'll close this.

PR #651, opened 2019-12-13
Only complain about inconsistent ns alias when an alias is defined

### LINKED ISSUE #650: Consistent alias doesn't work when only requiring namespace without alias
v2019.11.23

Native / macOS

**problem / repro**

In `config.edn` i have specified

```clojure
:consistent-alias {:aliases {datomic.api datomic}}
```
In one namespace I have

```clojure
(ns foo
  (:require [datomic.api]))
```
it shows: `Inconsistent alias. Expected datomic instead of .`

In this case, one might be requiring a ns only to load specs, for example, without providing an alias.  Since the alias is not used, adding it would throw another warning about unused alias 🤷‍♂

In addition, I have a similar case for a `:require-macros` instead of a simple `:require` 

**expected behavior**

"No alias" shouldn't be shown as inconsistent alias.

Heliosmaster:
We should only complain about inconsistent ns aliases when the alias is actually present.

Fixes #650

### DIFF
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -65,7 +65,7 @@
 (defn lint-alias-consistency [{:keys [:findings :config
                                       :filename]} ns-name alias]
   (when-let [expected-alias (get-in config [:linters :consistent-alias :aliases ns-name])]
-    (when-not (= expected-alias alias)
+    (when (and alias (not= expected-alias alias))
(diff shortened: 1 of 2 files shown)


borkdude:
@Heliosmaster Thanks. I think it might make sense to move the checking for the alias higher up:

```
167: (when as (lint-alias-consistency ctx ns-name as))
```

Heliosmaster:
@borkdude you're right. done.

Issue #657, opened 2019-12-14
seqable->seqable becomes seqable->seq


yuhan0:
**version**
clj-kondo v2019.11.23
**platform**
macOS 10.14, native
**editor**
Emacs+flycheck

**problem**
Looking at the source for the clojure.core types, I noticed an odd `:seqable-out` type that was being applied to the return values of the following functions:
- `reverse`
- `map`
- `remove`
- `filter` 
- `keep`

via `seqable->seqable`:
- `next`
- `rest`
- `nfirst`
- `nnext`
- `butlast`

I believe all of these functions return concrete lazy-seqs, and should be specced accordingly with `:seq`. 

**repro**

```clj
(ns repro)

(subvec (map inc (range)) 0 1)
```
**expected behavior**
```
Type error: expected vector but provided seq.
```
**actual behavior**
No errors reported because `:seqable-out` could be a vector according to the type checker

Note: I'm willing to submit a PR for this

borkdude:
Changing this to `:seq` is not correct since seqable -> seqable functions only guarantee to operate on something seqable, then they call seq on that and return something which is seqable again. E.g. it would not be incorrect if such a function would return nil or a vector, however, those are implementation details that you should not rely on. See @puredanger's comments here: https://github.com/borkdude/speculative/issues/45

yuhan0:
I still feel like these changes are justified in the context of clj-kondo whose goal is to be a useful linting tool, even if it means leveraging some current implementation details?
Perhaps not for `speculative` which I guess has to take a stricter approach to follow the correct spec.

Even if some functions like `cycle` return a specific class type that avoids forcing a seq, practically speaking it's used exactly like a seq 100% of the time and could never be treated as eg. a vector. 

I realise this is judgement-based and I might be missing some additional subtleties. Feel free to close if it doesn't align with clj-kondo's objectives, thanks!

borkdude:
@yuhan0 You can try the change and post the output of script/diff here, let’s see how it behaves.

yuhan0:
Here's the raw terminal output of `script/diff` after replacing all instances of :seqable-out -> :seq, looks like it didn't make any difference:

```
Linting and writing output to /tmp/clj-kondo-diff/branch.txt
Cloning into 'clj-kondo'...
remote: Enumerating objects: 123, done.
remote: Counting objects: 100% (123/123), done.
remote: Compressing objects: 100% (79/79), done.
[...]

PR #665, opened 2019-12-16
Changes to seq-returning functions

yuhan0:
As per the discussion in #657, and open to changes/rejection.

I tried to add a :stack = #{:list :vector} type as suggested, but was forced to expand its scope so much due to the transitive requirements that it ended up being nearly useless.

Given that, perhaps `peek` and `pop` should be simply specced as `#{:list :vector}` instead of introducing a new type?


borkdude:
@yuhan0 It seems like seqable-out is no longer used, so it can also be removed from types.clj?

yuhan0:
Okay, I removed all references to seqable-out from that file - however the `tag-from-meta` function now has an unused `out?` parameter.

PR #667, opened 2019-12-18
Don't specify current directory as cache

DjebbZ:
Following the current doc produces a lot of cache directories/files everywhere in the project, where/when the linter is executed. Also, this PR removes confusing `.` characters, we don't know if they should be part of the input or if they end the sentence. For consistency's sake it also removes all ending `.`. I only did it for Cursive as I noticed no other editor integration specified the `cache` option manually.


borkdude:
Thanks!

DjebbZ:
You're more than reactive :)

PR #671, opened 2019-12-19
Add type specs for more clojure.core functions

yuhan0:
This PR adds specs for most predicates ending in `?` that are simple any->boolean type-checking functions, excluding those like `extends?`, `future-done?` etc.), and common seq functions. Also removes a reference to `:seqable-out` I missed from the last PR.

`script/diff` shows no linting or performance difference from the current master branch.

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -927,7 +927,7 @@
        (node->line (:filename ctx) not-expr
                    :warning :not-empty?
                    "use the idiom (seq x) rather than (not (empty? x))")))
-    (analyze-children ctx (:children expr))))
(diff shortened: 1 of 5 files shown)


yuhan0:
For some odd reason, trying to spec `empty?` as `seqable->boolean` or `any->boolean` results in "Insufficient Input" errors being linted in every file where `empty?` is used. 
I tried tracing it down but was unsuccessful.

borkdude:
> For some odd reason, trying to spec empty? as seqable->boolean or any->boolean results in "Insufficient Input" errors being linted in every file where empty? is used.

That might be a bug, I'll take a look sometime soon.

borkdude:
@yuhan0 I fixed the issue with empty? now.

Issue #681, opened 2019-12-31
:refer :all on clojure.test is flagged as a warning


technomancy:
**version**

2019.08.21-alpha

**platform**

Debian buster, openjdk version "1.8.0_232", lein master

**problem**

When you have `(:require [clojure.test :refer :all])` in your code, it gets flagged as a warning, despite the fact that the entire reason I added `:refer/:all` to `require` is to support using `clojure.test` seamlessly.

**repro**

    (ns kondo.repro
      (:require [clojure.test :refer :all]))

**expected behavior**

clj-kondo should not complain about this since it is perfectly acceptable code. Using `:refer` here is not a good choice because it means if you add or remove a `testing` block you would have to go make changes in the `ns` clause, and using an alias for `clojure.test` is contrary to widespread accepted style.

I can submit a PR to fix this.

borkdude:
@technomancy Hey Phil, thanks for your amazing work with leiningen.

I propose implementing it as part of this feature: https://github.com/borkdude/clj-kondo/issues/560
for which a PR is welcome. People can then decide for themselves if they want to add `clojure.test` to this configuration. If you don't agree with this rule in general, you can turn it off with `{:linters {:refer-all {:level :off}}}'.

Issue #560, opened 2019-10-28
refer :all, be able to exclude certain namespaces like alda.core


borkdude:
Some libraries, like https://cljdoc.org/d/io.djy/alda-clj/0.2.2/api/alda.core, assume that the user is going to `:refer :all` the core namespace. The vars from these namespaces form a DSL. In these cases it would be nice to inform the `:refer-:all` rule that it should not give a warning.

Example code:

``` clojure 
(require '[alda.core :refer :all])

(comment
  "Clapping Music (1972)
   for two performers
   Steve Reich
   sheet music:
   http://earreader.nl/wp-content/uploads/2016/01/SteveReich-ClappingMusic.pdf")

(def pattern
  (let [clap "o2 d+8"
        rest (pause (duration (note-length 8)))]
    [clap clap clap rest clap clap rest clap rest clap clap rest]))

(defn shift [n pattern]
  (take (count pattern) (drop n (cycle pattern))))

(play!
 (tempo! 172)

 (part "midi-percussion")

 (voice 1)
 (repeat 13
         (repeat 12 pattern))

 (voice 2)
 (for [n (range 13)]
   (repeat 12 (shift n pattern))))

(println "Press Ctrl-C to stop & exit.")

(.addShutdownHook (Runtime/getRuntime)
                  (Thread. #(stop!)))
```

Proposal:

`{:linters {:refer-all {:exclude [alda.core]}}}`

PR #683, opened 2020-01-02
Add support for customizing exclusions to :refer :all

technomancy:
As discussed previously, this should take care of #560 and #681.

### DIFF
--- a/src/clj_kondo/impl/config.clj
+++ b/src/clj_kondo/impl/config.clj
@@ -63,7 +63,8 @@
                                     :exclude {#_#_taoensso.timbre [debug]}}
               :unused-private-var {:level :warning}
               :duplicate-require {:level :warning}
-              :refer-all {:level :warning}
+              :refer-all {:level :warning
+                          :exclude #{}}
               :use {:level :warning}
               :if {:level :warning}
               :type-mismatch {:level :error}
@@ -247,6 +248,16 @@
         delayed-cfg (memoize delayed-cfg)]
     (fn [config ns-name var-name]
       (contains? (delayed-cfg config) [ns-name var-name]))))
+
+(def refer-all-excluded?
+  (let [delayed-cfg (fn [config]
+                      (let [excluded (get-in config [:linters :refer-all :exclude]
+                                             #{'clojure.test})]
+                        (set excluded)))
+        delayed-cfg (memoize delayed-cfg)]
+    (fn [config referred-all-ns]
+      (let [excluded (delayed-cfg config)]
+        (excluded referred-all-ns)))))
 
 ;;;; Scratch
 
--- a/src/clj_kondo/impl/linters.clj
+++ b/src/clj_kondo/impl/linters.clj
@@ -383,7 +383,8 @@
             :message (str "#'" var-ns "/" (:name v) " is referred but never used")
             :row row
             :col col}))))
-    (doseq [[_referred-all-ns {:keys [:referred :node]}] refer-alls]
+    (doseq [[referred-all-ns {:keys [:referred :node]}] refer-alls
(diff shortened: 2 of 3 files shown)


borkdude:
Thanks! 

Two things:

- Can you remove the rule for `clojure.test`? I want people to decide this for themselves and not make any decisions about specific namespaces.
- Can you add a unit test for the exclusions?

If you want me to take it over from here, that's ok too!

technomancy:
> Can you remove the rule for clojure.test

I really don't think this is a good idea since that case is the whole reason why `:refer/all` exists, but if that's the only way this patch is going to get applied, then I can do it.

borkdude:
Linters can be turned off (`{:linters {:refer-all {:level :off}}}`) and with this new configuration people can `:exclude [clojure.test]`. I think that gives people enough space to make their own decisions if they don't agree with the defaults. 

Small detail: `:refer :all` doesn't work with CLJS, so also not with `cljs.test`.

So yeah, I really insist removing it.

yuhan0:
Just curious, what specifically about clojure.test needs a `:refer :all`? I use it like this in all my test namespaces:

```clojure
(ns something.test
  (:require [clojure.test :as t :refer [deftest testing is are]]))
```

which satisfies clj-kondo and doesn't cause any inconvenience.

technomancy:
Oh, I see; that's a shame about CLJS; I haven't used it so I didn't know.

> Just curious, what specifically about clojure.test needs a :refer :all

The problem with the snippet above is that most test namespaces don't use `are`, and many of them don't use `testing`. So the snippet above will not lint anyway, it will trigger an unused refer warning. If I make a change to the test to add or remove a `testing` block, it's very tedious to have to make sure that the `ns` block changes as well even though I haven't changed what I'm requiring. That kind of tedium usually makes people want to just quit using the linter altogether.

yuhan0:
That's an interesting perspective - personally I tend to disregard "unused refer" warnings as being harmless, so having a standardized set of refers (which my editor auto-inserts in newly created test ns) doesn't bother me enough to have to prune them.

I recall from using Python that the linter (probably pylint) would actually flag all unused refers in a wildcard import statement `from module import *`, so you would get dozens of warnings for all the vars being brought into scope and never used. 
Not saying that's what clj-kondo should do as well, but I think part of the rationale was to discourage people from using that syntax and be more explicit about what they were importing.

borkdude:
Like @yuhan0, my default `clojure.test` libspec looks like:

```
(ns something.test
  (:require [clojure.test :as t :refer [deftest is testing]]))
```

without `are` since I almost never use it.

> That kind of tedium usually makes people want to just quit using the linter altogether.

That's an opinion. Opinions can be configured in `.clj-kondo/config.edn`. Personally I don't find it tedious.

borkdude:
Like @yuhan0, my default clojure.test libspec looks like:

```
(ns something.test
  (:require [clojure.test :as t :refer [deftest is testing]]))
without are since I almost never use it.
```

> That kind of tedium usually makes people want to just quit using the linter altogether.

That's an opinion. Opinions can be configured in .clj-kondo/config.edn. Personally I don't find it tedious. Since the start of this linting rule you're the first one who wants this rule to be different for clojure.test. And now we have a config for it.

MnRac:
Personally I believe the linter really shouldn't have any opinion on which namespaces should be ignored and that should be left down to user preference. Especially considering there's people that would use it both ways.

borkdude:
Since I want to make a release one of these days, I'll merge this and make these changes:

-  Remove the specific setting for `clojure.test`
- Add a unit test for the new `:exclude` functionality

unless you will beat me to it.

technomancy:
I won't be able to do it in the next few days; go for it. Thanks.

Issue #719, opened 2020-01-28
Passing `--config` with file path causes java.lang.ClassCastException


ethpran:
**version**

clj-kondo v2019.12.14

**platform**

Native

**editor**

Intellij. Cursive

**problem**

passing a file path of my config file causes  java.lang.ClassCastException

**repro**

`clj-kondo --config some/path/config.edn --lint ./src`

```
Unexpected error. Please report an issue.
java.lang.ClassCastException: java.lang.String cannot be cast to java.io.File
	at clj_kondo.impl.core$read_edn_file$fn__6112.invoke(core.clj:51)
	at clj_kondo.impl.core$read_edn_file.invokeStatic(core.clj:51)
	at clj_kondo.impl.core$resolve_config.invokeStatic(core.clj:62)
	at clj_kondo.core$run_BANG_.invokeStatic(core.clj:84)
	at clj_kondo.main$main.invokeStatic(main.clj:85)
	at clj_kondo.main$main.doInvoke(main.clj:78)
	at clojure.lang.RestFn.applyTo(RestFn.java:137)
	at clojure.core$apply.invokeStatic(core.clj:665)
	at clj_kondo.main$_main$fn__6258.invoke(main.clj:104)
	at clj_kondo.main$_main.invokeStatic(main.clj:104)
	at clj_kondo.main$_main.doInvoke(main.clj:102)
	at clojure.lang.RestFn.applyTo(RestFn.java:137)
	at clj_kondo.main.main(Unknown Source)
```

**expected behavior**

Uses the config in my file as described [here](https://github.com/borkdude/clj-kondo/blob/master/doc/config.md)

I will submit a pull request. it just seems that `clj-kondo.impl.core/read-edn-file` function expects a java.io.File, but the cli arg gets passed as string

borkdude:
@ethpran Thanks for the fix. I'll leave this open so I can remind myself to write a unit test for it as well, so it will break if we change something.

ethpran:
what exactly do you want to test? if you give me some direction, i dont mind writing something. im just not 100% familiar with the project so i want to make sure i hit the right points :)

borkdude:
@ethpran In clj_kondo/main_test.clj the executable is tested using `(lint! "(foobar)" ...)` where on the dots you provide the command line arguments. There we could add a test for loading a config from a file. PR welcome.

ethpran:
[...]

PR #720, opened 2020-01-28
fixes parsing config from file path from cli. issue #719

ethpran:
fixes ClassCastException bc string is passed but `read-edn-file` has type hint for `java.io.File`


borkdude:
Thanks!

PR #724, opened 2020-01-31
adds a test which passes a config as a file path. relevant to issue #719

ethpran:
added a single simple test. without the config arg, there are 9 errors and 7 warnings, but with the config there are 0 for each. let me know if this works or if we should check some other cases


borkdude:
Thank you!

Issue #733, opened 2020-02-02
Infer return type from if and when expressions


borkdude:
```
$ clojure -A:clj-kondo --lint - <<< "(inc (if :foo :bar 'baz))"
<stdin>:1:6: warning: Expected: number, received: symbol or keyword.
```

TODO:

- [x] fix example:

```
(defn foo1 []
  :bar)

(defn foo2 []
  'baz)

(defn foo [x]
  (if (< x 5) (foo1) (foo2)))

(inc (foo 1))
```

PR #739, opened 2020-02-06
[#733] infer return tag from if and when expressions

borkdude:


borkdude:
Let's review https://github.com/borkdude/clj-kondo/pull/739/files, the .clj files a bit before merge.

PR #747, opened 2020-02-09
[#260] Add linting for single arity comparison

### LINKED ISSUE #260: Warn about constantly true expression `(= foo)`
1-arity use of `=` that is useful for apply, but most likely a mistake if used like `(= foo)` because it is constantly true.

abogoyavlensky:
Request adds linting for calls of logical operators and comparators with single arity that always produce the same value.

Linting diff:
```bash
130d129
< cljs/analyzer.cljc:2869:10: warning: 1-arity use of cljs.core/or always evaluates to the same value
332,335d330
< cljs/core.cljs:1872:11: warning: 1-arity use of cljs.core/and always evaluates to the same value
< cljs/core.cljs:1877:11: warning: 1-arity use of cljs.core/and always evaluates to the same value
< cljs/core.cljs:1905:11: warning: 1-arity use of cljs.core/and always evaluates to the same value
< cljs/core.cljs:1910:11: warning: 1-arity use of cljs.core/and always evaluates to the same value
1922,1923c1917
< src/clj_kondo/impl/linters.clj:29:19: warning: 1-arity use of clojure.core/and always evaluates to the same value
< linting took 10279ms, errors: 395, warnings: 1527
---
> linting took 10898ms, errors: 395, warnings: 1521
```

Resolves #260

### DIFF
--- a/test/clj_kondo/single_arity_comparison_test.clj
+++ b/test/clj_kondo/single_arity_comparison_test.clj
@@ -7,13 +7,13 @@
   (testing "test full linting error for single arity comparison in clojure"
     (assert-submaps
      '({:file "<stdin>", :row 1, :col 1, :level :warning,
-        :message "single arity use of clojure.core/= is constantly true"})
+        :message "Single arity use of clojure.core/= is always true"})
      (lint! "(= 1)")))
 
   (testing "test full linting error for single arity comparison in cljs"
     (assert-submaps
      '({:file "<stdin>", :row 1, :col 1, :level :warning,
-        :message "single arity use of cljs.core/not= is constantly false"})
+        :message "Single arity use of cljs.core/not= is always false"})
      (lint! "(not= 1)" "--lang" "cljs")))
 
   (testing "test linting comparison operators with single arity"
@@ -21,7 +21,7 @@
             op ["=" ">" "<" ">=" "<=" "=="]
             :let [errors (lint! (str "(" op " 1)") "--lang" lang)]]
       (is (= 1 (count errors)))
-      (is (= (format "single arity use of %s.core/%s is constantly true"
+      (is (= (format "Single arity use of %s.core/%s is always true"
                      (get {"clj" "clojure"} lang "cljs")
                      op)
              (-> errors
--- a/src/clj_kondo/impl/config.clj
+++ b/src/clj_kondo/impl/config.clj
@@ -76,7 +76,7 @@
                                  ;; different from str
                                  :aliases {#_clojure.string #_str}}
               :unused-import {:level :warning}
-              :single-arity-comparison {:level :warning}}
+              :single-operand-comparison {:level :warning}}
     :lint-as {cats.core/->= clojure.core/->
               cats.core/->>= clojure.core/->>
               rewrite-clj.custom-zipper.core/defn-switchable clojure.core/defn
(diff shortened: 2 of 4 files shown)


borkdude:
@abogoyavlensky First feedback:

Right now it seems you're checking for all calls of and and or with 1 arity and then warn about "always evaluates to the same value", but that's not entirely accurate.

E.g. in `(defn foo [x] (and x))` the expression `(and x)` does not always return "the same value". I think here the warning should be something like `"redundant call of clojure.core/and"` which is something else that what #260 is about: catching cases of `(< x)` which will always return true, no matter what the input is. I think the name could also be more clear, but I'm not sure which one... I'll think about it.

borkdude:
There are going to be other linters that are going to check for discarded constant values, e.g. `(when true 1 2 3)`, `1` and `2` are useless expressions here. So the name `:constant-value` is a bit too general. I think `:constant-comparison` is a better name for #260.

borkdude:
So concretely:

- remove `and` and `or` from the operators
- rename the linter to `:constant-comparison`

A small improvement could be to include the constant value `true` or `false` in the error message.

abogoyavlensky:
@borkdude Thank you! I will let you know a bit later as soon I will fix that.

test/clj_kondo/single_arity_comparison_test.clj
```diff
@@ -0,0 +1,42 @@
...
+    (assert-submaps
+     '({:file "<stdin>", :row 1, :col 1, :level :warning,
+        :message "single arity use of cljs.core/not= is constantly false"})
+     (lint! "(not= 1)" "--lang" "cljs")))
+
+  (testing "test linting comparison operators with single arity"
+    (doseq [lang ["clj" "cljs"]
+            op ["=" ">" "<" ">=" "<=" "=="]
+            :let [errors (lint! (str "(" op " 1)") "--lang" lang)]]
+      (is (= (count errors) 1))
```
vemv: `(count errors) 1` -> `1 (count errors)`

clojure.test expects `expected actual` order; else error reports are affected

abogoyavlensky: Thanks! Fixed it

abogoyavlensky:
@borkdude Just updated the request. I Removed `and` and `or` from linting. And chose second option for linter's name `:single-arity-comparison` because `(> 2 1)` is constant comparison too but we don't check it for now.

borkdude:
New insight: arity might not be the right word either, since single arity means a function with just one signature instead of multiple... @vemv suggested using the word `operand` instead, so `:single-operand-comparison`. What do you think?

abogoyavlensky:
@borkdude Yes, it would be more accurate name. I will change it.

borkdude:
@abogoyavlensky I made some small performance tweaks. The idea here is to do as little work as possible (creating sets, using contains?, comparisons, etc) and bail out as soon as we discover that the lint warning is not applicable. `utils/one-of` is a faster way to check if an element is contained in a set that can be expressed a compile time (it expands into a case statement).

abogoyavlensky:
@borkdude Thank you for pointing out and for improvements! I renamed `arity` to `operand` in code and comments as you and @vemv  suggested.

borkdude:
Thank you for your contribution!

PR #750, opened 2020-02-09
`impl.linters`: unwrap redundant `and`

vemv:
This is the only flaw Eastwood found in the codebase (other than a `clojure.core/type` shadowing in `defn reg-finding!` - not everyone likes those fixed, so I didn't out of caution)


borkdude:
Funny, I was going to suggest that name for a clj-kondo linter redundant and:

https://github.com/borkdude/clj-kondo/pull/747#issuecomment-583858700

PR #751, opened 2020-02-10
Implement :bang-suffix-consistency linter

vemv:
## What

Conceived as an optional linter to aid following `Use the bang! only for things not safe in an STM transaction` as per https://clojure.org/community/contrib_howto#_coding_guidelines (or at least, a variant of the convention taken to the extreme).

Otherwise, as humans it can be easy to use the "bang" in inconsistent ways, decreasing its overall usefulness.

## Notes

* This convention seems subjective so I'm not sure Kondo should default to it.
* I'm not sure about the performance of `postwalk`ing a vanilla Clojure list (as opposed to a rewrite-clj zipper). It wouldn't surprise me if it wasn't ideal.
  * Another reason for the proposed exclusion from the default linter config...
* Currently I don't use Kondo integrated with tooltips, so I don't know what this linter would look like for a tooltip-using user that decided to enable this linter.

Cheers - V


test/clj_kondo/main_test.clj
```diff
@@ -2483,6 +2483,32 @@
   (is (empty? (lint! "(ns main.core (:require [\"/vendors/daterangepicker\"]))"
                      "--lang" "cljc" "--cache" "true"))))
 
+(deftest lint-bangs!
+  (are [input expected] (testing input
```
vemv: I know this `are` looks kind of funny but everything is there for a reason (`testing`, `pr-str`, `true` at tail position...)

The result being: tests that fail very informatively (the day they fail).

Also I tried to write the test in such a way that unrelated bugs don't make it fail.

I can refactor into sth else but also would be happy to explain a given detail.

borkdude:
@vemv I think this linter is a good idea, but I have some improvements, that are easier for me to explain in code than in words. Can you open your PR branch for me to push?

test/clj_kondo/main_test.clj
borkdude: I broke the tests while trying to refactor and got this for output:

```
FAIL in (lint-bangs!) (main_test.clj:2487)
(defn a [] (((((((([[[1 1 1 1 (b!)]]])))))))))
({:file "<stdin>", :row 1, :col 32, :level :error, :message "unresolved symbol b!"})
expected: (= :invalid (get {0 :valid} (->> result (filter (comp #{"Functions using !-suffixed code should also be !-suffixed"} :message)) (count)) :invalid))
  actual: (not (= :invalid :valid))
```

I didn't find that very clear. I'd prefer if you refactored this into the same style of using `assert-submaps` like the other tests. That is also going to test the reported location, which is pretty important for in-editor usage of clj-kondo, which is a primary use case.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -404,6 +426,15 @@
                                         :macro? macro?))
                              %)
                            bodies)
+        _ (when (some-> config :linters :bang-suffix-consistency :level #{:error :warning})
```
borkdude: There is a function for that:

https://github.com/borkdude/clj-kondo/pull/747/files#diff-a0f7c1f7683ffc1de22015480c7ce38dR271

I didn't include it my PR for your PR, but I think it would be good to add it after you merge it.

borkdude: I now included it in my PR as well.

test/clj_kondo/main_test.clj
borkdude: I made a start with this in my PR.

vemv:
Thanks a lot for the response!

Will give the changes a spin this evening, obviously they look good and I'll be happy for the linter to be more performant.

vemv:
I have expanded the test suite with some intrincate cases. Pleasantly, clj-kondo isn't easy to fool :) but I could come up with a couple failing cases.

Can give them a shot if you'd like, although obviously they'll be easier for you.

src/clj_kondo/impl/linters.clj
```diff
@@ -121,6 +121,21 @@
       ;; (prn "tags" tags)
       (types/lint-arg-types ctx called-fn arg-types tags call))))
 
+(defn lint-bangs!
+  [ctx filename fn-name in-def]
+  (when in-def
+    (let [has-bang? (-> fn-name str (str/ends-with? "!"))]
```
vemv: Out of curiosity, was the lack of when-let a choice here?

test/clj_kondo/main_test.clj
vemv: > I didn't find that very clear. 

`(not (= :invalid :valid))` is telling you `not= <expected> <actual>` in that specific order (as per usual in clojure.test)

Since the expectation was `:invalid` but the test case is `:valid`, something is wrong. `({:file "<stdin>", :row 1, :col 32, :level :error, :message "unresolved symbol b!"})` tells you the full linting output, so you can see that the output , in fact, lacks the precise message we wanted.

---

The big selling point of `are` is that you can have large, aligned, DRY rows of assertions in the following style:

```clj
 "(defn a [] [[[[1 1 1 (apply b!)]]]])" :invalid
```

It reads very similarly to a clojure hashmap `{input expectation}`.

As a result one can be encouraged to add _many_ test cases - more than people usually do IME.

Anyway, I'd be OK if we refactor prior to merging.

borkdude:
@vemv My comment here https://github.com/borkdude/clj-kondo/pull/751/files/5bdeeddd4fe8877b91a6b7570aadd4bbde22fea8#diff-2db49ff5e46fe759c8d952ccfa076fa1 still stands. The errors now read:

```
FAIL in (lint-bangs!) (main_test.clj:2487)
(defn a [] [b!])
({:file "<stdin>", :row 1, :col 7, :level :warning, :message "Functions using !-suffixed code should also be !-suffixed"} {:file "<stdin>", :row 1, :col 13, :level :error, :message "unresolved symbol b!"})
expected: (= :valid (get {0 :valid} (->> result (filter (comp #{"Functions using !-suffixed code should also be !-suffixed"} :message)) (count)) :invalid))
  actual: (not (= :valid :invalid))

FAIL in (lint-bangs!) (main_test.clj:2487)
(defn a [] ^b! c)
({:file "<stdin>", :row 1, :col 7, :level :warning, :message "Functions using !-suffixed code should also be !-suffixed"} {:file "<stdin>", :row 1, :col 13, :level :error, :message "unresolved symbol b!"} {:file "<stdin>", :row 1, :col 16, :level :error, :message "unresolved symbol c"})
expected: (= :valid (get {0 :valid} (->> result (filter (comp #{"Functions using !-suffixed code should also be !-suffixed"} :message)) (count)) :invalid))
  actual: (not (= :valid :invalid))
```

which isn't very helpful. I'd appreciate it if you adapted the tests to use `assert-submaps` like all the other tests.

src/clj_kondo/impl/linters.clj
borkdude: No, just a glitch.

borkdude:
@vemv Ah, I missed the `testing` bit, so you can see what expression is failing:

```
FAIL in (lint-bangs!) (main_test.clj:2487)
(defn a [] [b!])
```

One thing I'm missing in these tests is testing the location of the error: it should imho point at the name of the function. And the error message should probably mention the name of the function.

So: `Functions using !-suffixed code should also be !-suffixed"` becomes `Function a uses !-suffixed functions, consider renaming it to a!`.

borkdude:
One other thing: since we're invoking this logic inside `lint-var-usages`, we can probably know if the used var is a function. If it's not a function, we can probably leave out the warning (which would fix the tests).

vemv:
Feel free to assign the PR to me, I could allocate the time during the week.

borkdude:
@vemv Here is the output of `script/diff` with the linter enabled:

https://pastebin.com/psUJVEMG

I think it would be good run through (some of) these to check for real life cases.

borkdude:
@vemv Fixed the failing test cases with this PR: https://github.com/reducecombine/clj-kondo/pull/2

borkdude:
Updated the script/diff output here: https://pastebin.com/mczDF9CU

borkdude:
Currently this linter is probably too strict with Reagent components:

Screenshot 2020-02-11 10 53 57

borkdude:
What about:

```
(defn foo [] (let [a (atom nil)] (reset! a nil)))
```

and

```
(defn foo [] (persistent! (transient {})))
```

vemv:
> Currently this linter is probably too strict with Reagent components:

The pattern (so far) boils down to: 

```clj
(defn pure []
  (fn [] ;; impure
    (swap! ...)))
```

which fortunately is a generic enough problem (detecting Reagent defns specificially might be fragile).

I consider this `defn pure` to be in fact pure (since invoking it mutates nothing). OTOH I can think of a couple more contrived cases:

```clj
(defn dubious [xs]
  (->> xs
       (map (fn []
              (swap! ...)))))

(defn impure [xs]
  (->> xs
       (run! (fn []
               (swap! ...)))))
```

Not sure of what to make about this. Same for the `local mutation` and `persistent!` cases.

Maybe I'd add  the observaiton that a false negative is better than a false positive. So maybe, if we can detect that a `defn` is too complex to analyze, we can abort the linting for that specific `defn`?

borkdude:
@vemv You're free to put more time in this, but my gut feeling is that the use of bangs in Clojure is too wishy-washy/subjective to make a good linter out of this. E.g. `run!` doesn't really do anything that makes you want to put a bang on your function name probably and `println` does do a side effect, but doesn't have a bang. Also the problems with local mutation, reagent components, etc. are probably too hard to solve as a linter. I'm inclined to see this PR as a nice experiment of which we both learned something and leave it at that.

vemv:
> the use of bangs in Clojure is too wishy-washy/subjective to make a good linter out of this.

Not sure if I agree in the conclusion, but also it's also not particularly important to pursue the absolute truth here ^^

> I'm inclined to see this PR as a nice experiment of which we both learned something and leave it at that.

I can say the same, I'm happy to have hacked on Kondo, learning some nuances for next time I'm inclined to propose an idea.

Thanks for your time!

Issue #741, opened 2020-02-07
Unsorted namespaces: investigate CLJC inconsistency with emacs


borkdude:
When sorting this namespace in emacs with `clojure-sort-ns`, the conditionals go on top:

``` clojure
(ns foo
  (:require
   #?(:clj [foo.bar])
   [bar.foo]))
```

but clj-kondo regards this as unsorted, since bar.foo goes before foo.bar.

PR #753, opened 2020-02-13
[#741] ignore unsorted namespaces in cljc branches

borkdude:


borkdude:
@Heliosmaster Added a test + fix that ignored unsorted namespaces if they occur in .cljc branches for now.

PR #823, opened 2020-03-27
Fixes #821 by adding analyzers for common JDBC wrappers

### LINKED ISSUE #821: support clojure.java.jdbc/with-transaction macro
From Slack:

Both clojure.java.jdbc and next.jdbc have a "with transaction" style macro that has the form (with-transaction [sym expr opts] & body) -- sym is introduced by the macro, for the body (like binding or several other core macros) but if I tell kondo to :lint-as any of those macros, it then complains about opts since a binding form should be an even number of forms. Can anyone think of a core macro that has an optional extra expression in a binding like that?

seancorfield:
Adds a JDBC analyzer. Mostly treats the `with-*` macros like `let` but adds checks for 2 or 3 forms and that the first form is a symbol (limited binding forms are valid).

### DIFF
--- a/src/clj_kondo/impl/analyzer/jdbc.clj
+++ b/src/clj_kondo/impl/analyzer/jdbc.clj
@@ -1,7 +1,7 @@
 (ns clj-kondo.impl.analyzer.jdbc
   {:no-doc true}
   (:require
-   [clj-kondo.impl.analyzer.common :refer [analyze-children
+   [clj-kondo.impl.analyzer.common :refer [analyze-expression**
                                            analyze-like-let]]
    [clj-kondo.impl.findings :as findings]
    [clj-kondo.impl.utils :as utils :refer [node->line]]))
@@ -18,8 +18,9 @@
   [{:keys [filename callstack] :as ctx} expr]
   (let [call (-> callstack first second)
         [f bindings & body] (:children expr)
-        [sym db-expr opts]  (:children bindings)]
-    (when-not (<= 2 (count (:children bindings)) 3)
+        binding-forms (:children bindings)
+        [sym db-expr opts] binding-forms]
+    (when-not (<= 2 (count binding-forms) 3)
       (findings/reg-finding!
        ctx
        (node->line filename bindings :error :syntax
@@ -29,12 +30,12 @@
        ctx
        (node->line filename sym :error :syntax
                    (format "%s binding form requires a symbol" call))))
-    (let [opts-analyzed (analyze-children ctx opts)
-          analyzed
-          (analyze-like-let ctx
-                            (assoc expr :children
-                                   (cons f
-                                         (cons (assoc bindings :children
-                                                      (list sym db-expr))
-                                               body))))]
+    (let [opts-analyzed (analyze-expression** ctx opts)
+          ;; a normal binding vector with just these two forms:
+          simple-binding (assoc bindings :children (list sym db-expr))
+          ;; reconstruct the expression with the normal binding instead:
+          expr-with-simple-binding
+          (assoc expr :children (concat [f simple-binding] body))
+          ;; now we can treat it like a regular let expression:
+          analyzed (analyze-like-let ctx expr-with-simple-binding)]
       (concat opts-analyzed (doall analyzed)))))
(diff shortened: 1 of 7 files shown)


test/clj_kondo/jdbc_test.clj
```diff
@@ -0,0 +1,14 @@
+(ns clj-kondo.jdbc-test
+  (:require
+   [clj-kondo.test-utils :refer [lint!]]
+   [clojure.java.io :as io]
+   [clojure.test :as t :refer [deftest is testing]]
+   [missing.test.assertions]))
+
+(deftest next-jdbc-test
```
borkdude: Instead of only testing the avoidance of false positive, it might also be nice to test if some of  the findings you emitted work? E.g. not a symbol as a binding.

src/clj_kondo/impl/analyzer/jdbc.clj
```diff
@@ -0,0 +1,40 @@
...
+      (findings/reg-finding!
+       ctx
+       (node->line filename bindings :error :syntax
+                   (format "%s binding form requires exactly 2 or 3 forms" call))))
+    (when-not (utils/symbol-token? sym)
+      (findings/reg-finding!
+       ctx
+       (node->line filename sym :error :syntax
+                   (format "%s binding form requires a symbol" call))))
+    (let [opts-analyzed (analyze-children ctx opts)
```
borkdude: opts is a single node here. single nodes get analyzed using `analyze-expression**`.

src/clj_kondo/impl/analyzer/jdbc.clj
```diff
@@ -0,0 +1,40 @@
...
+
+  We check there are 2 or 3 forms in the first argument (the binding vector).
+  We analyze the 3rd form if it is present (the options).
+  We analyze the whole expression as a let, after modifying the binding
+  expression to only have two children."
+  [{:keys [filename callstack] :as ctx} expr]
+  (let [call (-> callstack first second)
+        [f bindings & body] (:children expr)
+        [sym db-expr opts]  (:children bindings)]
+    (when-not (<= 2 (count (:children bindings)) 3)
```
borkdude: bind (:children bindings) to a variable to avoid the lookup twice

src/clj_kondo/impl/analyzer/jdbc.clj
```diff
@@ -0,0 +1,40 @@
...
+                   (format "%s binding form requires exactly 2 or 3 forms" call))))
+    (when-not (utils/symbol-token? sym)
+      (findings/reg-finding!
+       ctx
+       (node->line filename sym :error :syntax
+                   (format "%s binding form requires a symbol" call))))
+    (let [opts-analyzed (analyze-children ctx opts)
+          analyzed
+          (analyze-like-let ctx
+                            (assoc expr :children
```
borkdude: for clarity, maybe bind this expression to a name with `expr-without-opts`?

test/clj_kondo/jdbc_test.clj
borkdude: The way I usually test "findings" is to first call (prn (lint! "...")) and then paste that output back into the test file and test it with (assert-submaps the-output (lint! "..."))

seancorfield: I wasn't sure how you handled those tests in the suite. `corpus/jdbc/next.clj` has several of those tests that I used to manually verify the behaviors. Should I just add that into this test with verification of the four expected findings?

borkdude: Yeah. For example start looking here: https://github.com/borkdude/clj-kondo/blob/master/test/clj_kondo/main_test.clj#L14
and scan that file for corpus. It contains lots of examples of how this is tested.

seancorfield:
In the future, I'll try to remember to create an issue branch for each PR if I make any further updates!

borkdude:
Thanks a lot for this work.

ShlomitSibony:
Hi @borkdude ,

I'm currently working on a project that uses an old jdbc version in which `with-db-transaction` parameters can include up to 3 bindings (1-3).
`(with-db-transaction [t-con db-spec :isolation level :read-only? true]
    ... t-con ...)`
I'm using 2 bindings and I get the error `with-db-transaction binding form requires exactly 2 or 3 forms`

How can I exclude this type of error?

borkdude:
@ShlomitSibony I believe the options should be wrapped inside a map, as the third value in the binding.

PR #825, opened 2020-03-29
Add linter for conflicting aliases in ns form

### LINKED ISSUE #679: Warn about conflicting aliases
```
user=> (require '[clojure.set :as s] '[clojure.spec.alpha :as s])
Execution error (IllegalStateException) at user/eval1 (REPL:1).
Alias s already exists in namespace user, aliasing clojure.set
```

Heliosmaster:
Fixes #679

### DIFF
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -225,6 +225,7 @@
                  analyzed))
              kw+libspecs)
         _ (doseq [analyzed analyzed]
+            (namespace/lint-conflicting-aliases! ctx analyzed)
             (let [namespaces (map :ns analyzed)]
               (namespace/lint-unsorted-required-namespaces! ctx namespaces)
               (namespace/lint-duplicate-requires! ctx namespaces)))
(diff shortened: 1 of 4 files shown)


borkdude:
@Heliosmaster This looks great. One minor issue: I'd like the warning to appear on the alias node instead of the namespace node. Is it possible to fix this?

Heliosmaster:
@borkdude done :-) One minor caveat though the `:as` key of the namespace map points to the `:as` keyword rather than the alias itself. Is it enough?

In case not, we need to rework the implementation of `analyze-libspec` [here](https://github.com/borkdude/clj-kondo/blob/master/src/clj_kondo/impl/analyzer/namespace.clj#L150)

borkdude:
@Heliosmaster 
> the :as key of the namespace map points to the :as keyword rather than the alias itself. Is it enough?

I'm seeing this in my editor though?

Screenshot 2020-03-30 19 47 22

Heliosmaster:
> @Heliosmaster
> 
> > the :as key of the namespace map points to the :as keyword rather than the alias itself. Is it enough?
> 
> I'm seeing this in my editor though?
> 
> Screenshot 2020-03-30 19 47 22

Uhm, then I must have been mistaken. I am not using the editor integration and was manually checking from the CLI. Nevermind then 😄

borkdude:
I think the level can be bumped to `:error`: if it's a compile time error in Clojure, it can be an error in clj-kondo as well.

Issue #815, opened 2020-03-20
Warn on use of get-in with single key


oliyh:
**Describe the solution you'd like**
Code of the form `(get-in m [k])` should result in a warning saying prefer `(k m)` (when k is a keyword) or `(get m k)` when k is not a keyword.

**Additional context**
All three forms take a `not-found` final arg, it would be nice if the warning accounted for that too.

Heliosmaster:
Fixed by #826

PR #826, opened 2020-03-31
[#815] warn on get-in with single key

JDurstberger:
I added the warning when `get-in` is used with a single key. 

I didn't add the recommendation to use `(m k)` or `(get m k)` depending on if the single key is a keyword or not, because I wasn't sure how to determine if a node is a keyword. 

I am happy to add that if you can point me in the direction.


borkdude:
Nice job so far!

You can check if a node is a keyword by calling `:k` on it:

```
user=> (require '[clj-kondo.impl.rewrite-clj.parser :as p])
nil
user=> (:k (p/parse-string ":foo"))
:foo
```

Some thoughts:

- Should we generalize this to assoc-in and update-in as well? Then the name should also be more general.
- Should we enable this linter by default? It's not wrong to use get-in with a single element vector, so maybe it should be opt-in?

borkdude:
Output from the linting diff:

```
Run ./script/diff || true
Cloning into 'clj-kondo'...
82,83d81
< cljs/analyzer.cljc:817:23: warning: get-in with single key
< cljs/analyzer.cljc:823:13: warning: get-in with single key
86d83
< cljs/analyzer.cljc:975:29: warning: get-in with single key
397,399d393
< cljs/externs.clj:192:7: warning: get-in with single key
< cljs/externs.clj:199:9: warning: get-in with single key
< cljs/externs.clj:216:5: warning: get-in with single key
1923,1924c1917
< src/clj_kondo/impl/cache.clj:155:3: warning: get-in with single key
< linting took 18661ms, errors: 395, warnings: 1528
---
> linting took 19552ms, errors: 395, warnings: 1521
```
Might want to check if these reported places are legit.

borkdude:
Found a false positive here: https://github.com/clojure/clojurescript/blob/0c353f1947089cb8b1f010b4294b94ac109d4ef6/src/main/clojure/cljs/externs.clj#L192

I think it's wise to check of the second argument to `get-in` etc is a vector. If not, then don't report anything. The third element could also be a function call which returns a vector for example.

borkdude:
Another false positive:

```
src/clj_kondo/impl/cache.clj:155:3: warning: get-in with single key
```

The code for that is:

```
(get-in (from-cache-1 nil :clj 'clojure.core) ['defn])
```

borkdude:
From @seancorfield:

I would definitely turn this linter on, if it wasn't on by default (as long as it doesn't give false positives). It's fair to extend it to assoc-in but with update-in you need to bear in mind that we didn't have update until Clojure 1.7 so a lot of older code had to have update-in with [:k].
Perhaps :single-key-in as the general name for the linter?

src/clj_kondo/impl/linters.clj
```diff
@@ -5,6 +5,7 @@
    [clj-kondo.impl.config :as config]
    [clj-kondo.impl.findings :as findings]
    [clj-kondo.impl.namespace :as namespace]
+   [clj-kondo.impl.rewrite-clj.node.protocols :as node]
```
borkdude, line 8: we usually use clj-kondo.impl.utils/tag for this, since it's so commonly used.

src/clj_kondo/impl/linters.clj
```diff
@@ -102,12 +103,26 @@
                                (node->line (:filename ctx) expr :warning
                                            :constant-test-assertion "Test assertion with only constants.")))))
 
+(defn lint-single-key-in [ctx called-name call]
+  (let [keys (-> (:children call)
```
borkdude, line 107: It's often good to shortcut linters in case the configuration is set to :off. 
For this to work: 
- Add a default config in impl/config.clj for this linter and set the level to `:off`.
- Check at the start of the linter if the setting is not `identical?` to `:off` and then execute the linter. This saves performance when people do not use the linter.

src/clj_kondo/impl/linters.clj
```diff
@@ -102,12 +103,26 @@
...
+                 (nth 2))
+        keys-count (-> keys
+                       (:children)
+                       (count))]
+    (when (and (= :vector (node/tag keys)) (= 1 keys-count))
```
borkdude, line 112: I think it's better to shortcut the linter in case the keys element is not a vector. So introduce a another when with this check and then continue.

JDurstberger: I don't quite understand. 
The docs say that `and` does shortcut so the only benefit of lifting the vector check out is that we wouldn't evaluate the `keys-count`. 

Is this a slow expression?

JDurstberger:
Do you prefer conflicts to be resolved with rebases or merges?

borkdude:
@JDurstberger I prefer merge. The entire branch will be squashed once ready.

borkdude:
Thanks!

PR #831, opened 2020-04-01
Rename :if config key into :missing-else-branch and add doc

### LINKED ISSUE #830: Rename :if linter in :missing-else-branch

Heliosmaster:
See #830

### DIFF
--- a/src/clj_kondo/impl/config.clj
+++ b/src/clj_kondo/impl/config.clj
@@ -1,8 +1,9 @@
 (ns clj-kondo.impl.config
   {:no-doc true}
   (:require
-   [clj-kondo.impl.profiler :as profiler]
-   [clj-kondo.impl.utils :refer [vconj deep-merge map-vals]]))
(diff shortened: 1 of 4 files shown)


src/clj_kondo/impl/config.clj
```diff
@@ -103,11 +104,17 @@
...
 (defn merge-config! [cfg* cfg]
   (if (empty? cfg) cfg*
       (let [cfg (cond-> cfg
-                  (:skip-comments cfg)
-                  (-> (update :skip-args vconj 'clojure.core/comment 'cljs.core/comment)))]
+                        (:skip-comments cfg)
+                        (-> (update :skip-args vconj 'clojure.core/comment 'cljs.core/comment))
+
+                        :always
+                        (update-in [:linters] set/rename-keys renamed-config-keys))]
```
borkdude, line 117: I prefer explicitly handling it like:

```
(:if cfg) (assoc :missing-else-branch (:if cfg))
```
because that is more performant.

PR #841, opened 2020-04-06
Edit the vscode docs

PEZ:
I've simplified the vscode part of the docs. To me it doesn't make sense to use anything else but the `clj-kondo` extension. Also there was some stuff about Windows there that I don't think is relevant any more.


borkdude:
I like the changes, except the removal of the clojure-lint extension docs. In general it's good to give people the options if they exist. In rare cases people might not have Java installed and the clojure-lint extension will still work with VSCode.

PEZ:
I think it just confuses things. 😄 But I restored the info anyway, including the only reasonable use case I can see.

PR #849, opened 2020-04-09
Add :refer linter, default off

### LINKED ISSUE #847: Emit warning for all usages of `:refer`
Within my peers we have an in-house guidance which strongly discourages the use of `:refer` and `:refer-macros` within namespace `requires` to make the code more legible by clearly highlighting where the function/macro is defined. That is, instead of
```clojure
(ns my.cool.ns
  (:require [awesome.bits :refer :all]
            [important.stuff :refer [this where?]])

(-> where-did-this-come-from
    this
    belongs
    where?)
```
we enjoy this form more:
```clojure
(ns my.cool.ns
  (:require [awesome.bits :as awe]
            [important.stuff :as stuff])

(-> awe/where-did-this-come-from
    important/this
    awe/belongs
    important/where?)
```
As suggested by @borkdude himself on Slack, something similar to `:use {:level :warning}` for enforcing this would be great. Considering options I see the following possible configurations for this:
 1. for each use of `:refer`
   - ...emit warning for `:refer [<symbols>]`
   - ...emit warning for `:refer :all`
   - ...always emit a warning
 1. emit warning for each use of `:refer-macros` _(for CLJC/CLJS)_
 1. emit warning for each use of `:include-macros` _(for CLJC/CLJS)_

I did consider building something on top of [analysis data](https://github.com/borkdude/clj-kondo/tree/master/analysis) provided by clj-kondo but quickly decided against that in favor of having something more native, besides the personal general lack of time and unfamiliarity of kondo's internal workings.

Heliosmaster:
Fix for #847 

This linter will warn with **any** usage of `:refer` or `:refer-all` in `:require` clause.

We already have a `:refer-all` linter, which is similar but dedicated to refer-alls (I didn't touch it).

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -38,7 +38,7 @@ Some linters are not enabled by default. Right now these linters are:
 
 - `:missing-docstring`: warn when public var doesn't have a docstring.
 - `:unsorted-required-namespaces`: warn when namespaces in `:require` are not sorted.
-- `:refer`: warn when there is **any** usage of `:refer` or `:refer-all` in your namespace requirements.
+- `:refer`: warn when there is **any** usage of `:refer` in your namespace requirements.
 - `:single-key-in`: warn when using assoc-in, update-in or get-in with single key
 
 You can enable these linters by setting the `:level`:
--- a/doc/config.md
+++ b/doc/config.md
@@ -38,6 +38,7 @@ Some linters are not enabled by default. Right now these linters are:
 
 - `:missing-docstring`: warn when public var doesn't have a docstring.
 - `:unsorted-required-namespaces`: warn when namespaces in `:require` are not sorted.
+- `:refer`: warn when there is **any** usage of `:refer` or `:refer-all` in your namespace requirements.
 - `:single-key-in`: warn when using assoc-in, update-in or get-in with single key
 
 You can enable these linters by setting the `:level`:
--- a/doc/config.md
+++ b/doc/config.md
@@ -38,7 +38,7 @@ Some linters are not enabled by default. Right now these linters are:
 
 - `:missing-docstring`: warn when public var doesn't have a docstring.
 - `:unsorted-required-namespaces`: warn when namespaces in `:require` are not sorted.
-- `:refer`: warn when there is **any** usage of `:refer` in your namespace requirements.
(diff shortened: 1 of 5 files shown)


doc/config.md
```diff
@@ -38,6 +38,7 @@ Some linters are not enabled by default. Right now these linters are:
 
 - `:missing-docstring`: warn when public var doesn't have a docstring.
 - `:unsorted-required-namespaces`: warn when namespaces in `:require` are not sorted.
+- `:refer`: warn when there is **any** usage of `:refer` or `:refer-all` in your namespace requirements.
```
OscarMarshall, line 41:
```suggestion
- `:refer`: warn when there is **any** usage of `:refer` in your namespace requirements.
```

"Refer all" still uses the `:refer` keyword but uses `:all` as the value (e.g. `[clojure.test :refer :all]`).

Heliosmaster: Whoops, i got confused with the name of our linter :-D  Thanks!

borkdude:
@Heliosmaster and @OscarMarshall Thank you.
@esuomi Is this what you had in mind? 

Screenshot 2020-04-10 11 08 50

borkdude:
@Heliosmaster I think the message should remove the colon in `:require`:

Screenshot 2020-04-10 11 11 04

esuomi:
@borkdude Yep, this would be a fix for my immediate issue. I guess semantically it would make more sense to highlight the `:refer`, not `:require` though :)

Heliosmaster:
@borkdude I can use the :refer, i guess, but currently `(:require-kw namespace)` is just a keyword, rather than a node. Should i change it?

borkdude:
@Heliosmaster @esuomi The linter now reports at the location of `:refer`. It also reports in `:use` + `:only` and `:refer-macros` in CLJS.

Screenshot 2020-04-15 20 17 29

borkdude:
Hmm, maybe `:use` + `:only` doesn't make sense. Thoughts?

Heliosmaster:
> Hmm, maybe `:use` + `:only` doesn't make sense. Thoughts?

Ideally `:use` [should be discouraged](https://github.com/bbatsov/clojure-style-guide#prefer-require-over-use], but i confess i never saw it with `:only`

borkdude:
I've removed the consideration of `:use` for this linter.

PR #879, opened 2020-05-04
Feature : add js-arity info

### LINKED ISSUE #90: Built-in arity infomation of JavaScript functions
Using a document like https://github.com/google/closure-compiler/blob/081e13cca07a3bce2807f94406164bc37439cf91/externs/es3.js
we might be able to extract arities of common JS functions.

kolharsam:
This should resolve #90


borkdude:
@kolharsam I'll have a look soon.

.gitignore
```diff
@@ -75,3 +77,4 @@ clj-kondo*.snap
 .DS_Store
 !corpus/*jar
 lsp
+run();
```
borkdude, line 80: why is this needed?

borkdude:
@kolharsam I think this is a good start, but I prefer if possible to not check in huge files into source control that are only used for extracting information. Rather write a shell script which clones the project that has the relevant file es3.json and then runs jsdoc on that. And then invoke some Clojure to produce the necessary transit files, same as we do with the Java extraction. There we clone OpenJDK and run code over that to emit transit files.

.gitignore
kolharsam: My apologies. I think this was an unintended mistake.
Will revert.

kolharsam:
@borkdude I was just thinking about this and wanted your opinion on it.
What if we implement this arity information feature as a bb-pod? 
I was thinking if `clj-kondo` would be the pod-client and obtain information from a bb-script or a JS file?

borkdude:
What problem would hiding information in a pod solve?

kolharsam:
@borkdude I was able to work on the feedback that you gave from the last review. So let me know if the issues have been sorted in the latest commits. The script does work but I noticed that the JSON generated didn't the information of functions under the `Math` object. This is a potential thing to `FIXME`.

Also, the next step, as discussed involves the generation the `transit` files. Could I just leverage `jet` for this purpose?

borkdude:
Thanks! For generating the transit files from the JSON you can just write a regular Clojure script, similar to what we do for gathering Java arity information in script/extract-java.
 Follow the steps of what that script does and what it outputs, the code should be very similar for JS.

PR #921, opened 2020-06-23
Update hooks docs

lread:
Current docs are good and guided me when writing my first hook.
Hooks aren't scary hard to write, nor are they easy peasy. Here are some changes
to hooks docs that I think might help a bit:

- Let folks know hooks are interpreted with sci
- Introduce the concept of rewrite-clj nodes early so that subsequent references
  to nodes have context
- Give the reason why someone might want to write an analyze-call hook earlier
- Try to focus more on the code than its representation as a node
- Give a couple of examples of what Clojure code looks like when parsed into
  a rewrite-clj node
- Other minor changes and, hopefully welcome, tweaks.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -368,13 +368,20 @@ If you prefer not to lint the contents of `(comment ...)` forms, use this config
 
 ## Hooks
 
-Hooks are a way to enhance linting with user provided code.
+Hooks are a way to enhance linting via user provided code.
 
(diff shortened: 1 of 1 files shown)


lread:
Writing docs is hard! Lemme know what you think of my changes.

borkdude:
Excellent, thank you!

PR #922, opened 2020-06-24
Document `:from-var`

seanpoulter:
Hi @borkdude. Thanks for your help on Slack last week. As discussed, here's a PR to update the analysis docs.

Would you expect this much to change? It looks like the keys have changed order and the column positions have incremented by one. What's the chance you changed the column base index from 0 to 1?

**Steps to Generate Example Code**
Follow the docs:
* Create `/tmp/foo.clj`
* Run `clj-kondo` using `clj --main cli-kondo.main` from the repo:

  ```shell
  clj --main clj-kondo.main --lint /tmp/foo.clj --config '{:output {:analysis true :format :edn}}' \
  | jet --pretty --query ':analysis'
  ```

### DIFF
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -52,6 +52,7 @@ The analysis output consists of a map with:
   - `:name`: the name of the used var
   - `:from`: the namespace from which the var was used
   - `:to`: the namespace of the used var
+  - `:from-var`: the function name from which the var was used
 
   Optional:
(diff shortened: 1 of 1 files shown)


borkdude:
> What's the chance you changed the column base index from 0 to 1?

I don't think this has changed.

borkdude:
Thanks!

PR #926, opened 2020-06-26
docs: note that JDK 8 / 11 based graal is required

rschmukler:
Makes a note that currently only JDK 8 based graal VMs are supported.

Since many clojure devs aren't yet familiar with compiling native
images it's nice to help them stay on a safe path that works.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -5,6 +5,8 @@
 To build the `clj-kondo` binary from source:
 
 * Download [GraalVM](https://github.com/graalvm/graalvm-ce-builds/releases) java8 19 or newer.
+  Currently `clj-kondo` can only be built with java11 based graal VMs, so using JDK 8 builds will
+  not work.
 
 * Set the `GRAALVM_HOME` variable. E.g.:
 
@@ -20,7 +22,7 @@ Place the binary somewhere on your path.
 
 ### Optional steps
 
-These steps are only necessary to update resources that are shipped with clj-kondo. 
+These steps are only necessary to update resources that are shipped with clj-kondo.
 
 * To update the built-in cache for Clojure and ClojureScript:
 


borkdude:
@rschmukler Hi. Your PR prompted me to upgrade to GraalVM 11. That is now done. So the docs can be updated to mention that.

rschmukler:
Awesome work! I'll submit an updated version of the docs. Thanks 👍

rschmukler:
@borkdude Just confirmed that I was able to build using graal w/ JDK 11.

I tested with the JDK 8 based graal and it now fails on macro expansion. I presume this is an intentional switch. I've updated the documentation to now explicitly state that only JDK 11 based graal installs will work. In case it wasn't, below is the the error I got:

```
Execution error (UnsupportedClassVersionError) at java.lang.ClassLoader/defineClass1 (ClassLoader.java:-2).
sci/impl/Reflector has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

My graal version:

```
λ $GRAALVM_HOME/bin/java -version
openjdk version "1.8.0_252"
OpenJDK Runtime Environment (build 1.8.0_252-b09)
OpenJDK 64-Bit Server VM GraalVM CE 20.1.0 (build 25.252-b09-jvmci-20.1-b02, mixed mode)
```

I don't have any real opinions as to whether it's worth supporting two versions of the JDK in something as cutting edge as Graal, just wanted to make sure that the choice was explicit on your end.

Thanks again for the quick addition to the hooks API and your time upgrading!

borkdude:
Yes, the switch is intentional: clj-kondo is going to only support one version of GraalVM. However, it can still be run on the JVM with both 8 and 11.

Issue #938, opened 2020-07-22
Lint amount of arguments against format string in clojure.core/format


borkdude:
E.g. `(format "%s %s" 1)` should yield a lint error.

TODO:
- [x] When the first argument is not a string literal, linting should be skipped.
- [x] Support ns-local config
- [x] Handle format options with indexes like `(format "%1$s %1$s" 1)`
- [x] Tests

PR #941, opened 2020-07-26
[#938] enable format string checking for clojure.tools.logging

borkdude:


celwell:
Just double-checking, is that `#_` supposed to be here: https://github.com/borkdude/clj-kondo/pull/941/files#diff-e1aaa64b205752f191f004233f9bd9beR1162

We are experiencing some false positives on this sort of thing:
`(format "/blah/%s/blah?query=Luke%%20Skywalker" s0)`

Kondo is saying I need 3 args not 2.

borkdude:
Made an issue for it here which also describes the workaround of turning this linter off for now:

https://github.com/borkdude/clj-kondo/issues/944

Would appreciate it if you could list more false positives in the issue, so I can fix them.

borkdude:
@celwell New version with a fix is released.

celwell:
@borkdude Thanks for the quick fix, and for clj-kondo in general!

PR #946, opened 2020-07-29
Provide option to pin install to a specific version

cldwalker:
Hi borkdude.
 Thanks for the wonderful linter. We use it at work :) I'd like to propose a --release-version flag to the install script in order to pin the version. The use case for this is that we want to rely on a pinned version to avoid any accidental bugs in the latest version. In particular, we got bit by https://github.com/borkdude/clj-kondo/issues/944 which led to some CI churn.

Example usage: `script/install-clj-kondo --dir $PWD/foo --release-version 2020.06.12`

Thoughts? Happy to change the option name to whatever you prefer.

Cheers, Gabriel

### DIFF
--- a/doc/companies.md
+++ b/doc/companies.md
@@ -29,6 +29,7 @@ listed!
 - [Metosin](https://www.metosin.fi)
 - [NextJournal](https://nextjournal.com)
 - [Nubank](https://www.nubank.com.br)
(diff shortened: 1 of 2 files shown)


borkdude:
Yes please! README.md could also be updated. Sorry about the new false positives. I would appreciate the examples that triggered your failing CI build so I can fix them.

PR #947, opened 2020-07-29
Mention new install option from #946

cldwalker:
Heyo. Thanks for merging in the install option quickly. Here's the requested update to the relevant doc


borkdude:
@cldwalker FYI, I renamed `--release-version` to just `--version` since `release` seemed a bit redundant. The old name is still supported, but not mentioned in the docs.

cldwalker:
Thanks for the heads up. I've changed the CI job to --version so no need to support the old name on my behalf

PR #951, opened 2020-08-07
Provide impl for rum.core/defcs

FieryCod:


borkdude:
From Slack:

I think the following transformation will be slightly better:
```
(defsc foo [state x] body1 body2) => (defn foo [x] (let [state {}] body1 body2)
```
So you call it as `(f 1)` and clj-kondo will think this is correct, since the generated form only receives one arg (or n-1 args in general) and state is still recognized as a valid binding.  

And `(f 1 2)` should trigger an arity warning

PR #953, opened 2020-08-09
Add .pre-commit-hooks.yaml

robhanlon22:
A simple configuration file for [pre-commit](https://pre-commit.com/) that allows clj-kondo to be run as, well, a pre-commit hook!

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: clj-kondo
+  name: clj-kondo
+  description: A linter for Clojure code that sparks joy
+  entry: --entrypoint clj-kondo borkdude/clj-kondo:2020.07.26 --lint
+  language: docker_image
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+- id: clj-kondo
+  name: clj-kondo
+  description: A linter for Clojure code that sparks joy
+  entry: --entrypoint clj-kondo borkdude/clj-kondo:latest --lint
```
robhanlon22, line 4: Ideally this would point a specific version, but there is no Docker image that is tagged for the current release.

borkdude:
I generally don't check in personal project config into source control since people tend to make different choices. Feel free to add it to .gitignore and document it somewhere on this page, near the existing pre-commit hook docs: https://github.com/borkdude/clj-kondo/blob/master/doc/ci-integration.md#pre-commit-hook.

PR #971, opened 2020-08-26
Mention nested let* in hook expansion

### LINKED ISSUE #966: redundant-let false positives in macro hook expansions
**version**

v2020.07.29

**macro usage**

yes, example from docs produces false positives

**platform**

release from github

**editor**

Emacs 27

**problem**

I've noticed that clj-kondo lints it's hook expansion results, which sometimes produces false positive `redundant`-style warnings.
For example, with the code from the documentation:

```clojure
(ns hooks.with-bound
  (:require [clj-kondo.hooks-api :as api]))

(defn with-bound [{:keys [:node]}]
  (let [[binding-vec & body] (rest (:children node))
        [sym val opts] (:children binding-vec)]
    (when-not (and sym val)
      (throw (ex-info "No sym and val provided" {})))
    (let [new-node (api/list-node
                    (list*
                     (api/token-node 'let)
                     (api/vector-node [sym val])
                     opts
                     body))]
      {:node new-node})))
```

```clojure
{:hooks {:analyze-call {my-lib/with-bound hooks.with-bound/with-bound}}}
```

and given clojure file:

```clojure
1| (ns my-app.core
2|   (:require [my-lib]))
3|
4| (my-lib/with-bound [life 42 {:question-answered true}]
5|   (println life))
6|
7| (let [answer 41]
8|   (my-lib/with-bound [life answer {:question-answered true}]
9|     (println life)))
```

I get this warning:

```
./core.clj::: warning: Redundant let expression.
```

Which contains no line and column information. This is due to expansion of `my-lib/with-bound` to `let` thus producing:

```clojure
7 | (let [answer 41]
8 |   (let [life answer]
9 |     {:question-answered true}
10|     (println life)))
```

Which is indeed a redundant `let`, but not due to mistake.

Because we teach clj-kondo how to lint this macro, we also have further inspectations of expanded result, and if `life` is not used in the `with-bound` clj-kondo will report it.
However, since

> It is not important that the code is rewritten exactly according to the macroexpansion

Suche false positives should be controlled. However if we disable all `:redundant-let` for this macro we'll lose this lint for the body of the expanded macro, e.g.:

```clojure
(let [answer 41]
  (my-lib/with-bound [life answer {:question-answered true}]
    (let [a 10]
      (let [b 20] ;; <= redundant let
        (+ life a b)))))
```

flycheck-clj-kondo also produces this kind of warning:

```
Suspicious state from syntax checker clj-kondo-clj: Flycheck checker clj-kondo-clj returned 2, but its output contained no errors: <stdin>::: warning: Redundant let expression.
linting took 15ms, errors: 0, warnings: 1

Try installing a more recent version of clj-kondo-clj, and please open a bug report if the issue persists in the latest release.  Thanks!
```

Which seems as a minor incompatibility.

As a workaround I've wrapped the expanded result into `try-finally` block, e.g.:

```clojure
(let [answer 41]
  (try
    (finally nil)
    (let [life answer]
      {:question-answered true}
      (println life))))
```

This does not produce `redundant-let` warning, but is a bit hacky way to avoid such situation.

Sorry if this was already reported, I couldn't find this particular issue by searching woth `redundant` query

andreyorst:
fix #966

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -455,12 +455,17 @@ This is the code for the hook:
       (throw (ex-info "No sym and val provided" {})))
     (let [new-node (api/list-node
                     (list*
-                     (api/token-node 'let)
+                     (api/token-node 'let*)
                      (api/vector-node [sym val])
                      opts
                      body))]
       {:node new-node})))
 ```
+
+Note, in order to prevent false positive `redundant-let` warning, when using
+this macro within actual let block, instead of using `(api/token-node 'let)`
+we're using `(api/token-node 'let*)`. Clj-kondo will lint `let*` syntactically
+the same, but won't report it as a nested `let`.
 
 This code will be placed in a file `hooks/with_bound.clj` in your `.clj-kondo`
 directory.


borkdude:
Thanks!

PR #1002, opened 2020-09-09
Readme: Add --parallel to Project Setup

robert-stuttaford:
Also provide copy/paste-ready examples for project setup calls for Boot and Clojure CLJ


borkdude:
Thanks!

Issue #1010, opened 2020-09-14
False positive unresolved-symbol deftest test name due to strange interaction with :refer :all


tervorz:
**version**

clj-kondo v2020.09.09

**platform**

I'm using the native version on MacOS 10.15.6, however this is reproducible if I build master locally and run the jar against test files described in the problem section below.

**editor**

Not particularly applicable given this is reproducible outside it, but I'm using Emacs 26.3 (Doom) with flycheck.

**problem**

I get an unresolved-symbol error on the names of my tests when using `deftest`, but only when using `:refer :all` for `clojure.test` and also for any other second namespace. Very nuanced. So `(deftest my-test ...)` gives me an unresolved-symbol error on `my-test`.

**repro**

I wrote a unit test for this in my fork: https://github.com/tervorz/clj-kondo/blob/deftest-resolve-test-name-false-positive/test/clj_kondo/main_test.clj#L2134

Here is the test:
```
  (is (empty? (lint! (io/file "corpus" "deftest_resolve_test_name_fail.clj")
                     '{:linters {:unresolved-symbol {:level :error}
                                 :refer-all {:level :warning
                                             :exclude [clojure.test clojure.string]}}}))))
```

Contents of `deftest_resolve_test_name_fail.clj`:
```
(ns deftest-resolve-test-name-fail
  (:require [clojure.string :refer :all]
            [clojure.test :refer :all]))
(deftest my-test (is (blank? "")))
```

Linter output:
```
{:file "corpus/resolve_test_name_fail.clj",
 :row 4,
 :col 10,
:level :error,
:message "unresolved symbol my-test"}
```

This bug is pretty nuanced. It only surfaces if you have two `:refer :all` namespaces in your test file, and _only_ if `clojure.test :refer :all` comes after some other namespace's `:refer :all`. If `clojure.test` comes second, it works correctly, as can be seen in this unit test: https://github.com/tervorz/clj-kondo/blob/deftest-resolve-test-name-false-positive/test/clj_kondo/main_test.clj#L2130

**config**
```
{:linters {:unresolved-symbol {:level :error}
           :refer-all         {:level   :warning
                               :exclude [clojure.test clojure.string]}}}
```

**expected behavior**
No error.

I attempted to fix it on my own, but I couldn't pinpoint where the special handling was happening around deftest test names.

Thanks for your work on this! I'm new, but enjoying using it for my linting so far.

borkdude:
[...]

PR #1013, opened 2020-09-17
[#1010] fix for deftest false positive unresolved-symbol

tervorz:
Fix for deftest false positive unresolved-symbol with more than one :refer :all requires.

There's a little hardcoding that I think needs to happen here, due to the lack of info we get when :refer :all is used. I'd be happy to add this to the config options if you'd like, but as we discussed, it's so nuanced that that's probably overkill.

Lmk what changes might be needed and I'll be happy to fix.

### DIFF
--- a/corpus/deftest_resolve_test_name_fail.clj
+++ b/corpus/deftest_resolve_test_name_fail.clj
@@ -1,0 +1,6 @@
+(ns deftest-resolve-test-name-fail
+  (:require [clojure.string :refer :all]
+            [clojure.test :refer :all]))
+(deftest my-test
+  (is (blank? ""))
+  (is (thrown? (throw (Exception.)))))
--- a/corpus/deftest_resolve_test_name_pass.clj
+++ b/corpus/deftest_resolve_test_name_pass.clj
@@ -1,0 +1,6 @@
+(ns deftest-resolve-test-name-pass
+  (:require [clojure.test :refer :all]
+            [clojure.string :refer :all]))
+(deftest my-test
+  (is (blank? ""))
+  (is (thrown? (throw (Exception.)))))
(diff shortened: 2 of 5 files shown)


borkdude:
@tervorz Thanks. I think the unit tests are a good start. However, the approach taken in your solution hardcodes information that clj-kondo already has: it knows that deftest is a var in clojure.test via files in the built-in analysis cache:

https://github.com/borkdude/clj-kondo/blob/master/resources/clj_kondo/impl/cache/built_in/clj/clojure.test.transit.json

So I think it would be better to find out why clj-kondo couldn't figure it out in the case you reproduced, because it should be able to. The resolving of the var to a namespace is done here: https://github.com/borkdude/clj-kondo/blob/2fb62280e84c116342e52d062e6ed26ba99c1ccb/src/clj_kondo/impl/linters.clj#L214.

tervorz:
@borkdude Awesome, I appreciate the feedback and the hints. I'll try to circle back to this at some point soonish to see if I can utilize that built in cache and resolve it in the right place.

borkdude:
@tervorz The built-in cache is already utilized.

E.g. you will see an arity error when you do:

``` 
(require '[clojure.test :as t])

(t/testing)
``` 

The arity information comes from the (built-in) analysis cache. It's just probably the logic that tries to resolve unqualified vars to the right namespaces that might have a bug.

tervorz:
> @tervorz The built-in cache is already utilized.
> 
> E.g. you will see an arity error when you do:
> 
> ```
> (require '[clojure.test :as t])
> 
> (t/testing)
> ```
> 
> The arity information comes from the (built-in) analysis cache. It's just probably the logic that tries to resolve unqualified vars to the right namespaces that might have a bug.

@borkdude Yep I was following. Thanks for the clarification though.

PR #1027, opened 2020-10-02
Add claypoole to examples

vincentjames501:


borkdude:
Hey @vincentjames501, thanks! These examples now go to https://github.com/clj-kondo/config so they can be automatically installed in the user's config.

borkdude:
So can you point your PR towards that repo?

PR #1048, opened 2020-10-22
Assoc count of files in summary when analysis is activated

pepijn:
Attempt to resolve https://github.com/borkdude/clj-kondo/issues/1046

Considerations (I'm unfamiliar with the codebase, this is my first PR):
- I decided to only assoc the `:files` when analysis is activated, because I found the `analysis` atom the path of least resistance to obtaining filenames.
- I didn't want to deref `analysis` in every case because I assume there to be performance implications.


borkdude:
@pepijn I think it would be nicer to not depend on analysis.

Check the code in this area: https://github.com/borkdude/clj-kondo/blob/a4898324d9ef66ca06d3b4c8e6ce5d1f7ad98cd2/src/clj_kondo/impl/core.clj#L249 where all files (from stdin, directories or jars) are passing through. That might give you some ideas.

pepijn:
Makes sense, thanks for the pointer @borkdude.

I took a look and made some changes, what's your opinion on them?

I decided to add another atom to the mix because the most likely candidate `sources` seems to be tied with being a parallel execution or not. Didn't want to mess with that.

borkdude:
@pepijn Starting to look good. But since we're not interested in the filenames themselves, a simple `(atom 0)` + `swap!` + `inc` will do I think. Also a test would be nice.

pepijn:
@borkdude made the changes. The test is quite minimal. If you'd like more, could you give me a pointer on what part of the logic you'd like to see tested better? Both the schedule and summarize functions are largely untested now—I could refactor that.

src/clj_kondo/core.clj
```diff
@@ -151,6 +153,10 @@
 ;;;; Scratch
 
 (comment
+  (do (-> (run! {:lint   ["/Users/pepijn.looije/dev/nu/itaipu/tourniquet/src"]
+                 :config {:output {:analysis false}}})
+          :summary))
+
```
pepijn, line 159: oops

borkdude:
@pepijn Not sure what you have in mind, but this looks good enough to me.

pepijn:
If you're happy I'm happy. I'm ready to merge.

PR #1060, opened 2020-11-05
circleci: remove jvm job step "create uberjar" and store compiled -standalone.jar per platform as artifact

bennyandresen:
I was unable to native-image the previously generated standalone.jar with graalvm8-20.2 or with graalvm11-20.2.

I've not changed the circleci run steps to use script/compile because I didn't want to change too much around.

Both +clj-1.10.2 and +native-image have to be enabled for it to work, otherwise there are different errors.
(I could provide logs if this is of interest.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,16 +40,6 @@ jobs:
           paths:
             - ~/.m2
           key: v1-dependencies-{{ checksum "deps.edn" }}
-      - run:
-          name: Create uberjar
-          command: |
-            mkdir -p /tmp/release
-            lein do clean, uberjar
-            VERSION=$(cat resources/CLJ_KONDO_VERSION)
-            cp target/clj-kondo-$VERSION-standalone.jar /tmp/release
-      - store_artifacts:
-          path: /tmp/release
-          destination: release
 
   linux:
     docker:
--- a/.circleci/script/release
+++ b/.circleci/script/release
@@ -1,10 +1,13 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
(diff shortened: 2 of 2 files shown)


bennyandresen:
Removed the update to latest available clojure-1.10.2-alpha3 version, as that failed to compile because of a `sci' reflector error.

bennyandresen:
from the above ci/circleci:jvm test:
```
Syntax error macroexpanding at (interop.cljc:1:1).
Execution error (UnsupportedClassVersionError) at java.lang.ClassLoader/defineClass1 (ClassLoader.java:-2).
sci/impl/Reflector has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
````
I don't have that problem on my local machine, but I also started with a pretty fresh `.m2` cache.

borkdude:
@bennyandresen The problem is the Java version in the JVM step. It should be set to Java 11.

borkdude:
Although I would not recommend changing it for the current JVM step since we want to verify that clj-kondo works on java8 as well.

borkdude:
@bennyandresen So when compiling with Java 8 it's better to NOT include the native-image profile since that pulls in specific Java 11 fixes that do not work with Java 8:

https://github.com/borkdude/clj-kondo/blob/5a4e3a139012586ddb788910010cdb1fa97383b5/project.clj#L21

borkdude:
It might be better if the native-image step on CircleCI just posted the uberjar as an artifact as well. This way you are guaranteed to use the same one

PR #1098, opened 2020-12-10
Fix: Intellij settings path

audriu:
Added more detailed path to file watchers settings


borkdude:
@audriu Could it be that this is different on Windows / Linux and macOS?

audriu:
good point! I only tested on GNU/Linux...

PR #1108, opened 2021-01-01
Fixes #982 Adds locals to analysis with an extra config flag

### LINKED ISSUE #982: Add locals and local-usages to analysis
From https://github.com/snoe/clojure-lsp/pull/176

The issue is that more can be done with the analysis information if locals are also present. Think renaming, code-completion, symbol-lookup, etc...

> About locals: clj-kondo knows where a local is introduced and where the local is used. We could add that to the analysis info

Whatever is easiest, these are two that @borkdude suggested:
> `{:locals [{:name x :row 12 :col 16}], :local-usages [{:name x :row 17 :col 17 :source-row 12 :source-col 16}]}`
> `{:locals [{:name x :row 12 :col 16 :id 123}], :local-usages [{:name x :row 17 :col 17 :id 123}]}`

**Additional context**
A conversation on slack about the wider goals of analysis means that linting performance should not be impacted, and possibly behind a future flag.

snoe
> I imagine the difficulty comes in with macros, and the reticence is because linting may not need it? I think carve shows the utility of the analyzer itself, so I wonder if we think about other tools it could be a useful exercise. Renaming, I think, is right up there beside dead code for utility. A cli tool that could correct any inconsistent aliases, or rename a fully qualified var without resorting to sed and careful diff checking. Is it fair to want the analyzer to grow beyond kondo with some examples like this?
> My thinking is that kondo's analyzer feels like the closest thing we have to a tooling analyzer.

borkdude
> I think I'm fine with this, with the following caveats: 1) Some things are easier to lint while analyzing, this is why clj-kondo intertwines linting with analysis sometimes. So splitting an analyzer into a library is hard because of this. 2) I don't like clj-kondo linting getting slower because of extra work it has to do for other projects. The in-editor experience should be as fast as possible. So if we can prevent that, that's cool. I think we can accomplish that just using some configuration options. (edited)

snoe:
As described in #982, this is to add an extra level of analysis to kondo for other tools.

A new config output flag `:locals` to turn this on.

`:scope-end-col` and `:scope-end-row` are important keys on the `:locals` analysis for completion type tools that need to know at what positions a var is valid for (using the `:end-row` and `:end-col` of a symbol is sufficient for start of completion in most cases but an argument could be made for a more explicit `:scope-start-row/col`.

The `:str` is important to know how an expr appears in code as it is difficult (impossible?) to recreate the string from the symbol and namespace with namespaced keywords and destructuring differences.

Locally I have 5 failing tests on master, and this branch has the same, so it seems like an environment problem that I hope ci accounts for.

https://github.com/clojure-lsp/clojure-lsp/pull/176

### DIFF
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -11,6 +11,16 @@ data, use the following configuration:
 When using clj-kondo from the command line, the analysis data will be exported
 with `{:output {:format ...}}` set to `:json` or `:edn`.
 
+## Extra Analysis
+
+Further analysis can be returned by providing `:analysis` with a map of options:
+
+``` shellsession
+{:output {:analysis {... ...}}
+```
+
+- `:locals`: when truthy return `:locals` and `:local-usages` described below
+
 ## Data
 
 The analysis output consists of a map with:
@@ -60,6 +70,20 @@ The analysis output consists of a map with:
     was resolved: `:clj` or `:cljs`
   - several attributes of the used var: `:private`, `:macro`, `:fixed-arities`,
     `:varargs-min-arity`, `:deprecated`.
+
+- `:locals`, a list of maps with:
+  - `:filename`, `:row`, `:col`, `:end-row`, `:end-col`
+  - `:id`: an identification for this local, `:local-usages` will reference this
+  - `:name`: the name of the used local
+  - `:str`: the as written string of the local from the file and position
+  - `:scope-end-row`: the row in which this local will go out of scope
+  - `:scope-end-col`: the column in which this local will go out of scope
+
+- `:local-usages`, a list of maps with:
+  - `:filename`, `:row`, `:col`, `:end-row`, `:end-col`
+  - `:id`: an identification for this local, refers to the local declaration with the same `:id` in `:locals`
+  - `:name`: the name of the used local
+  - `:str`: the as written string of the local from the file and position
 
 Example output after linting this code:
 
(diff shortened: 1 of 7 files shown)


borkdude:
@snoe Looking very good overall. I have some minor remarks. I'll be writing down some comments in this message and let you know when I'm done.

- [ ] I would prefer the `:locals` option to be part of the `:analysis`: `{:output {:analysis {:locals true} ...}}`

borkdude:
@snoe I consider the above review done. Please do not use force-pushes so we don't invalidate the links in the review.

snoe:
> is `:tag `used in clojure-lsp for something? 
No, sorry. I see I should be more explicit about select-keys instead of just adding to the internal `binding` shape. Eventually it probably will be useful, when you rename a destructured keyword in cursive it asks if you want to rename the keyword globally. So knowing if a local came from keyword destructuring could be useful in the future.

> why have both :name and :str
`(let [{:keys [::x/b]} 1] b)` in lsp, I can rename `b` or rename the alias `x` and the destructuring would change appropriately. I might with some difficulty be able to do the `b` by looking back from `end-col`, not sure about the `x` though with all the ways to specify the namespace. Having the thing as written has been generally useful in lsp internals tho. Maybe it's another flag?

I should be able to fix most of the others, thank you.

snoe:
@borkdude I believe this is ready.

> this is not part of your change, so feel free to ignore, but I notice while reviewing we can improve delaying reading some options until they are actually needed
I think I know what you were referring to here, `keys-destructuring?` and `skip-reg-binding?` could both go within some of the conditionals but as I made more repetition I was inclined to collapse the symbol v keyword handling and I think it would be better to address it outside this pr.

Issue #1056, opened 2020-11-01
Detect duplicate key in '{[1 2] 3, (1 2) 4}


borkdude:

jysandy:
@borkdude I'm planning on picking this up. From the looks of it, I'll have to make a change in this `lint-map-keys` function https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/linters/keys.clj#L17. I'm a new contributor, so I just wanted to run this by you and make sure you didn't have something else in mind.

jysandy:
It doesn't look like this particular case is unit tested, so I was thinking of adding a test or two to cover it as well.

borkdude:
@jysandy That's the correct place to look. Thanks for looking into it!

PR #1114, opened 2021-01-07
[#1056] Detect duplicate key in '{[1 2] 3, (1 2) 4}

jysandy:
Implements https://github.com/clj-kondo/clj-kondo/issues/1056.

Output of `script/diff`: 

```
Cloning into 'clj-kondo'...
remote: Enumerating objects: 9672, done.
remote: Total 9672 (delta 0), reused 0 (delta 0), pack-reused 9672
Receiving objects: 100% (9672/9672), 11.04 MiB | 5.55 MiB/s, done.
Resolving deltas: 100% (5594/5594), done.
1c1
< "Elapsed time: 8.203851 msecs"
---
> "Elapsed time: 8.956134 msecs"
1966c1966
---
2253c2253
---
2693c2693
< linting took 10730ms, errors: 400, warnings: 2291
---
> linting took 11817ms, errors: 400, warnings: 2291
```

### DIFF
--- a/parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
@@ -43,7 +43,7 @@
 (defn ^:no-doc concat-strings
   "Convert nodes to strings and concatenate them."
   [nodes]
-  (string/join " " (map string nodes)))
+  (reduce str (map string nodes)))
 
 ;; ## Inner Node
 
--- a/parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
@@ -43,7 +43,7 @@
 (defn ^:no-doc concat-strings
   "Convert nodes to strings and concatenate them."
   [nodes]
-  (reduce str (map string nodes)))
+  (string/join " " (map string nodes)))
 
 ;; ## Inner Node
 
--- a/src/clj_kondo/impl/linters/keys.clj
+++ b/src/clj_kondo/impl/linters/keys.clj
@@ -5,7 +5,7 @@
    [clj-kondo.impl.utils :refer [node->line tag]]))
 
 (defn key-value
-  "We only support tokens as key values for now."
+  "We only support tokens, vectors and lists as key values for now."
   [node]
   (case (tag node)
     :token (or (when-let [v (:k node)]
--- a/src/clj_kondo/impl/linters/keys.clj
+++ b/src/clj_kondo/impl/linters/keys.clj
@@ -12,6 +12,8 @@
                  (if (:namespaced? node)
                    (str v) v))
                (str node))
+    :vector (map key-value (:children node))
+    :list (map key-value (:children node))
     nil))
 
 (defn lint-map-keys
--- a/src/clj_kondo/impl/linters/keys.clj
+++ b/src/clj_kondo/impl/linters/keys.clj
@@ -2,8 +2,7 @@
   {:no-doc true}
   (:require
    [clj-kondo.impl.findings :as findings]
-   [clj-kondo.impl.utils :refer [node->line tag]]
-   [clojure.string :as string]))
+   [clj-kondo.impl.utils :refer [node->line tag]]))
 
 (defn- map-without-nils
   "Like map, but returns nil if any mapped value is nil."
@@ -33,30 +32,6 @@
     :quote (recur (first (:children node)))
     nil))
 
-(defn- stringify-key-expr
-  [node]
-  (case (tag node)
-    :vector (->> (:children node)
-                 (map stringify-key-expr)
-                 (string/join " ")
-                 (format "[%s]"))
-    :list (->> (:children node)
-               (map stringify-key-expr)
-               (string/join " ")
-               (format "(%s)"))
-    :set (->> (:children node)
-              (map stringify-key-expr)
-              (string/join " ")
-              (format "#{%s}"))
-    :map (->> (:children node)
-              (map stringify-key-expr)
-              (string/join " ")
-              (format "{%s}"))
-    :quote (str (:prefix node)
-                (reduce str (map stringify-key-expr
-                                 (:children node))))
-    (str node)))
-
 (defn lint-map-keys
   ([ctx expr]
    (lint-map-keys ctx expr nil))
@@ -75,13 +50,13 @@
                ctx
                (node->line filename
                            key-expr :error :duplicate-map-key
-                           (str "duplicate key " (stringify-key-expr key-expr)))))
+                           (str "duplicate key " key-expr))))
             (when-not (known-key? k)
               (findings/reg-finding!
                ctx
                (node->line filename
                            key-expr :error :syntax
-                           (str "unknown option " (stringify-key-expr key-expr)))))
+                           (str "unknown option " key-expr))))
             (update acc :seen conj k))
           acc))
       {:seen #{}}
--- a/src/clj_kondo/impl/linters/keys.clj
+++ b/src/clj_kondo/impl/linters/keys.clj
@@ -5,6 +5,22 @@
    [clj-kondo.impl.utils :refer [node->line tag]]
    [clojure.string :as string]))
 
+(defn- ensure-no-nils
+  "Returns nil if the collection contains a nil, else returns coll.
+  Doesn't search for nested nils."
+  [coll]
+  (cond
+    (set? coll) (if (contains? coll nil)
+                  nil
+                  coll)
+    (sequential? coll) (if (contains? (set coll) nil)
+                         nil
+                         coll)
+    (map? coll) (if (contains? (set (flatten (seq coll))) nil)
+                  nil
+                  coll)
+    :else coll))
+
 (defn key-value
   "We only support the following cases for now."
   [node]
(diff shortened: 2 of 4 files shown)


jysandy:
Currently, the duplicate sequence keys are reported without spaces in them, for example: `{[1 2] \"bar\" (1 2) 12}` is reported as `duplicate key (12)`. Working on a fix for this.

test/clj_kondo/impl/analyzer_test.clj
```diff
@@ -109,7 +112,48 @@
              (do
                (reset! findings [])
                (analyze "1..1")
-               @findings))))))
+               @findings))))
+
+    (testing "duplicate map keys"
```
borkdude, line 117: This test should be added to clj_kondo/main_test.clj: `deftest map-duplicate-keys`. This file should not be touched for this feature.

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
```diff
@@ -43,7 +43,7 @@
 (defn ^:no-doc concat-strings
   "Convert nodes to strings and concatenate them."
   [nodes]
-  (reduce str (map string nodes)))
+  (string/join " " (map string nodes)))
```
jysandy, line 46: Although the tests passed, I was a little hesitant to change this function because of potential breakage. If that's an issue, I can just make this a separate function and call it only for `SeqNode`s.

jysandy:
The code currently does not handle `{[1 2] 3, '(1 2) 4}`. Let me know if I should try to handle that in this PR.

test/clj_kondo/impl/analyzer_test.clj
jysandy: Addressed this in the latest commits.

borkdude:
@jysandy I think that would be nice to handle as well!

src/clj_kondo/impl/linters/keys.clj
```diff
@@ -5,13 +5,16 @@
...
+  "We only support tokens, vectors, quoted forms and lists as key values for now."
   [node]
   (case (tag node)
     :token (or (when-let [v (:k node)]
                  (if (:namespaced? node)
                    (str v) v))
                (str node))
+    :vector (map key-value (:children node))
+    :list (map key-value (:children node))
+    :quote (recur (first (:children node)))
```
jysandy, line 17: Took the first of the children here — I'm not sure in what case a `quote` node can have multiple children.

jysandy:
`{[1 2] 3, '(1 2) 4}` should now be detected.

jysandy:
What do you think of handling `{{1 2} \"bar\" {1 2} 12}`, or handling set literals as keys?

src/clj_kondo/impl/linters/keys.clj
borkdude: I think that is the correct solution, thank you.

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
borkdude: Yeah, please don't change it here. Made a custom print function for the lint output.

borkdude: Although I don't think it should break anything actually. It might make sense to leave it here... I'll have to think about it. This might actually be an improvement ;).

borkdude:
@jysandy All examples for which the Clojure reader/compiler itself will throw, are welcome to add.

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
borkdude: @jysandy To be sure, let's not change the vendored version of rewrite-clj but make a custom printer that does what you did there.

jysandy: I took a cursory look at the callers of this function, and I think it might actually be an improvement. That being said, I've made a new printer instead in the latest commit like you asked.

src/clj_kondo/impl/linters/keys.clj
```diff
@@ -2,18 +2,64 @@
   {:no-doc true}
   (:require
    [clj-kondo.impl.findings :as findings]
-   [clj-kondo.impl.utils :refer [node->line tag]]))
+   [clj-kondo.impl.utils :refer [node->line tag]]
+   [clojure.string :as string]))
+
+(defn- ensure-no-nils
```
jysandy, line 8: This is in case a collection contains a node we don't support as a key. Let me know if this defensive programming isn't necessary.

jysandy:
Added support for map literals and set literals, and a test for a nested literal. That's about all I can think of at the moment.

src/clj_kondo/impl/linters/keys.clj
```diff
@@ -2,18 +2,64 @@
...
 (defn key-value
-  "We only support tokens as key values for now."
+  "We only support the following cases for now."
   [node]
   (case (tag node)
     :token (or (when-let [v (:k node)]
                  (if (:namespaced? node)
                    (str v) v))
                (str node))
+    :vector (ensure-no-nils (map key-value (:children node)))
```
borkdude, line 32: Can you try to write it in a short-cutting fashion, so we don't iterate more than needed?

E.g.: (pseudocode): `(reduce (fn [acc child] (if-let [v (key-value child)] ... (reduced nil)) (:children node))`

src/clj_kondo/impl/linters/keys.clj
```diff
@@ -2,18 +2,64 @@
...
+    :list (ensure-no-nils (map key-value (:children node)))
+    :set (ensure-no-nils (set (map key-value (:children node))))
+    :map (ensure-no-nils (apply hash-map (map key-value (:children node))))
```
borkdude, line 35: You can check for nils before making a hash-map out of this, so you don't have to iterate over the hash-map later.

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
borkdude: @jysandy What would be the consequences if we changed the rewrite-clj function? What are the callers and where does this improve things? I have something in mind, but I wonder if you have the same in mind.

src/clj_kondo/impl/linters/keys.clj
borkdude: Added a comment in the PR diff.

I think the defensive programming is correct here, but it can be made more efficient:

- By shortcutting on the first nil key-value
- By checking for nil before coercing into a set or map instead of walking that set or map later on

jysandy: Addressed in the latest commit.

src/clj_kondo/impl/linters/keys.clj
jysandy: Addressed in the latest commit.

src/clj_kondo/impl/linters/keys.clj
jysandy: Addressed in the latest commit.

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
jysandy: These are the callers according to Cursive:

Screenshot 2021-01-12 at 12 39 30 PM

For `fn`, `forms`, `meta` and `seq`, I think it makes sense to have the children separated by spaces when printed, so that they are visually distinct. For `quote`, `reader_macro` and `uneval`, I don't think these nodes should have more than one child anyway. But I could be wrong.

src/clj_kondo/impl/linters/keys.clj
```diff
@@ -5,21 +5,15 @@
...
-                         nil
-                         coll)
-    (map? coll) (if (contains? (set (flatten (seq coll))) nil)
-                  nil
-                  coll)
-    :else coll))
+(defn- map-without-nils
+  "Like map, but returns nil if any mapped value is nil."
+  [f coll]
+  (reduce (fn [acc v]
```
borkdude, line 11: Might want to use a transient here, like in `mapv`:

```
([f coll]
     (-> (reduce (fn [v o] (conj! v (f o))) (transient []) coll)
         persistent!))
```

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
borkdude: I don't think clj-kondo relies a specific form of the printed form but it may help debugging in hooks code when you call `println` or `str` on generated hook code. So all in all, I think this will be an improvement. Let's try it and see if the tests pass. You can temporarily revert your custom print function.

src/clj_kondo/impl/linters/keys.clj
jysandy: Addressed this, thanks for the tip.

src/clj_kondo/impl/linters/keys.clj
```diff
@@ -8,12 +8,13 @@
 (defn- map-without-nils
   "Like map, but returns nil if any mapped value is nil."
   [f coll]
-  (reduce (fn [acc v]
-            (if-let [new-v (f v)]
-              (conj acc new-v)
-              (reduced nil)))
-          []
-          coll))
+  (-> (reduce (fn [acc v]
```
borkdude, line 11: Use `some->` to short-cut on nil, because `(persistent! nil)` will cause trouble.

jysandy: Fixed this, good catch.

parser/clj_kondo/impl/rewrite_clj/node/protocols.clj
jysandy: @borkdude I've removed the custom printer in the latest commit.

borkdude:
@jysandy Thanks!

Issue #1116, opened 2021-01-11
lint-missing-else-branch and lint-single-key-in not called on cljs fns


tomdl89:
**version**

master

**platform**

native & JVM

**editor**

emacs

**problem**

The `lint-specific-calls!` case form is missing `cljs.core` versions of the fns it checks for `lint-missing-else-branch` and `single-key-in`. Probably just a simple oversight 

**repro**

Lint a cljs file where you have an if-let with a missing else branch, for example

**config**

You need to have `:single-key-in` set to `:warning` to repro `lint-single-key-in`

**expected behavior**

lint cljs too

PR to follow...

borkdude:
@tomdl89 I usually don't close issues until the next release. See https://github.com/clj-kondo/clj-kondo/projects/1

PR #1117, opened 2021-01-11
Add cljs.core cases in `lint-specific-calls!`

tomdl89:
Fix for [1116](https://github.com/clj-kondo/clj-kondo/issues/1116).

It would be cool to get the `lint!` fn from clj-kondo-test-utils to test against both clj and cljs (or at least have an option to allow it) but I couldn't figure that out. This fixes the issue, but I'd appreciate input on how we could prevent regression (espeically when adding new specific-call linters) by use of tests.

### DIFF
--- a/test/clj_kondo/main_test.clj
+++ b/test/clj_kondo/main_test.clj
@@ -740,10 +740,15 @@ foo/foo ;; this does use the private var
   (assert-submaps
    '({:file "<stdin>",
       :row 1,
-      :col 1,
+      :col 9,
+      :level :warning,
+      :message "Missing else branch."}
+     {:file "<stdin>",
+      :row 1,
+      :col 35,
       :level :warning,
       :message "Missing else branch."})
-   (lint! "(if-let [x 1] true)"))
+   (lint! "#?(:clj (if-let [x 1] true) :cljs (if-let [x 1] true))" "--lang" "cljc"))
   (is (empty? (lint! "(if-let [{:keys [row col]} {:row 1 :col 2}] row 1)"))))
 
 (deftest if-some-test
--- a/test/clj_kondo/main_test.clj
+++ b/test/clj_kondo/main_test.clj
@@ -2610,26 +2610,44 @@ foo/foo ;; this does use the private var
   (assert-submaps
    '({:file "<stdin>",
       :row 1,
-      :col 12,
+      :col 20,
+      :level :warning,
+      :message "get-in with single key"}
+     {:file "<stdin>",
+      :row 1,
+      :col 43,
       :level :warning,
       :message "get-in with single key"})
-   (lint! "(get-in {} [:k])" {:linters {:single-key-in {:level :warning}}}))
+   (lint! "#?(:clj (get-in {} [:k]) :cljs (get-in {} [:k]))" "--lang" "cljc"
+          "--config" {:linters {:single-key-in {:level :warning}}}))
 
   (assert-submaps
    '({:file "<stdin>",
       :row 1,
-      :col 14,
+      :col 22,
+      :level :warning,
+      :message "assoc-in with single key"}
+     {:file "<stdin>",
+      :row 1,
+      :col 50,
       :level :warning,
       :message "assoc-in with single key"})
-   (lint! "(assoc-in {} [:k] :v)" {:linters {:single-key-in {:level :warning}}}))
+   (lint! "#?(:clj (assoc-in {} [:k] :v) :cljs (assoc-in {} [:k] :v))" "--lang" "cljc"
+          "--config" {:linters {:single-key-in {:level :warning}}}))
 
   (assert-submaps
    '({:file "<stdin>",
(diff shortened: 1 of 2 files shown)


borkdude:
@tomdl89 The `lint!` function accepts cmd line args, like `(lint! "foo" "--lang" "cljs")`. A test to add this to the single-key-in tests would be great!

tomdl89:
I tried:
```
(lint! "(if-let [foo bar] (foo baz))" {:lang :cljs})
```
and I'm pretty sure it produced a lint error _before_ my change, making me think it doesn't replicate actually linting a cljs file. I'll try again, but if you think of anything I may be doing wrong, please let me know. Thanks!

borkdude:
@tomdl89 You need to use string opts like on the cmd line:

```
"--lang" "cljs"
```

tomdl89:
@borkdude ah yes that worked! adding tests now

tomdl89:
Is there a way to test both in clj and cljs other than duplicating the test with the above cmd line args? I hoped
```
"--lang" "cljc"
```
would achieve that, but it appears not 😞

borkdude:
@tomdl89 I think cljc should actually work:

```
$ clj-kondo --lint - --lang cljc <<< '#?(:clj future :cljs nil)'
linting took 15ms, errors: 0, warnings: 0
```

tomdl89:
@borkdude lemme know what you think

test/clj_kondo/main_test.clj
```diff
@@ -740,10 +740,15 @@ foo/foo ;; this does use the private var
...
+      :col 9,
+      :level :warning,
+      :message "Missing else branch."}
+     {:file "<stdin>",
+      :row 1,
+      :col 35,
       :level :warning,
       :message "Missing else branch."})
-   (lint! "(if-let [x 1] true)"))
+   (lint! "#?(:clj (if-let [x 1] true) :cljs (if-let [x 1] true))" "--lang" "cljc"))
```
borkdude, line 751: Actually I think the reader conditionals aren't needed here: code outside reader conditionals counts for all hosts, so if one of them would generate a an error, you would still see it.

borkdude: E.g.:

```
$ clj-kondo --lint - --lang cljc <<< 'future'
<stdin>:1:1: error: unresolved symbol future
```

We are getting this error because of CLJS, even though it was valid in CLJ.

tomdl89: I need not "one of them" but "both of them" to generate an error, otherwise this test won't spot regressions where the cljs case is omitted. Hopefully I have not misunderstood.

borkdude: This lints the code unconditionally as both CLJ and CLJS, because it is not in a conditional:

```
clj-kondo --lint - --lang cljc <<< 'future'
```

tomdl89: When I change the test to:
```
   (lint! "(if-let [x 1] true)" "--lang" "cljc")
```
and remove the second asserted submap, the test does pass, but if I comment out the `[cljs.core if-let]` from the cases in `lint-specific-calls!` the test _still_ passes, and it shouldn't, because I want to test against such a regression.

borkdude: @tomdl89 I understand that, but `(if-let [x 1] true)` is exactly the same as `#?(:clj (if-let [x 1] true) :cljs (if-let [x 1] true))`. It might make sense to pull the test apart for each lang and not use cljc then.

Just use a variable and doseq:

```
(doseq [lang ["clj" "cljs"]]
  (is ...))
```

tomdl89: Actually, it's not exactly the same, because the former outputs one result, and the latter outputs two results. Indeed, if I comment out the cljs case in `lint-specific-calls!` with the reader-conditional approach I get this error:
```
Expected 2 results, but got: 1
```
so it works well at capturing the regression I'm targeting. I'm happy to use the doseq approach if you feel it's more readable? Sorry to be contrary - just want to make sure I implement a solid test for this 🙂

borkdude: Let's do the `doseq` approach since it feels more robust.

test/clj_kondo/main_test.clj
```diff
@@ -771,13 +772,14 @@ foo/foo ;; this does use the private var
       :level :error,
       :message "if-some binding vector requires exactly 2 forms"})
    (lint! "(if-some [x 1 y])"))
-  (assert-submaps
-   '({:file "<stdin>",
-      :row 1,
-      :col 1,
-      :level :warning,
-      :message "Missing else branch."})
-   (lint! "(if-let [x 1] true)"))
```
tomdl89, line 780: interestingly, the test for if-some was actually testing if-let

tomdl89:
@borkdude thanks for the tip - doseq looks much neater

Issue #1078, opened 2020-11-23
Missing error for requiring lib under prefix


DeLaGuardo:
**version**

clj-kondo v2020.11.07

**problem**

require block in namespace declaration can have a prefixed form. Like this

```
(ns prefix.core
  (:require [prefix
             [lib :as lib]
             [core :as core]]))
```

in this form lib names should not have dots otherwise compiler will throw an exception. However clj-kondo is not generating an error or warning for namespace declaration like this:

```
(ns prefix.core
  (:require [prefix
             [lib :as lib]
             [lib.core :as lib.core]]))
```

**repro**

```
echo '(ns some.ns (:require [prefix [lib :as lib] [lib.core :as lib.core]])) (lib/init) (lib.core/init)' | clj-kondo --lint -
```

**expected behavior**

```
(ns prefix.core
  (:require [prefix
             [lib :as lib]
             [lib.core :as lib.core]]))
              ~~~~~~~~
              ^ 
              Found lib name 'lib.core' containing period with prefix 'prefix'.  lib names inside prefix lists must not contain periods  
```

jysandy:
@borkdude shall I pick up this as well? It looks like a change would be needed somewhere here https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/analyzer/namespace.clj#L246

What would be the correct approach here: throw an exception and stop further linting, or report the error without halting analysis?

In case you were wondering BTW — I have some time available this month, so I thought I might stop by and pick up some issues.

borkdude:
@jysandy Sure!

borkdude:
@jysandy The approach in clj-kondo is always try to continue linting despite errors. Try to make the best of it.

PR #1118, opened 2021-01-13
[#1078] Detect periods contained in lib names inside prefix lists

jysandy:
Fixes https://github.com/clj-kondo/clj-kondo/issues/1078.

Output of `script/diff`:
```
Cloning into 'clj-kondo'...
remote: Enumerating objects: 82, done.
remote: Counting objects: 100% (82/82), done.
remote: Compressing objects: 100% (65/65), done.
remote: Total 9754 (delta 29), reused 48 (delta 11), pack-reused 9672
Receiving objects: 100% (9754/9754), 11.10 MiB | 5.71 MiB/s, done.
Resolving deltas: 100% (5623/5623), done.
1c1
< "Elapsed time: 10.222993 msecs"
---
> "Elapsed time: 9.500667 msecs"
1966c1966
---
2253c2253
---
2693c2693
< linting took 11818ms, errors: 400, warnings: 2291
---
> linting took 12279ms, errors: 400, warnings: 2291
```

### DIFF
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -61,7 +61,8 @@
                                                  libspec-expr
                                                  :error
                                                  :syntax
-                                                 "Lib names inside prefix lists must not contain periods.")))
+                                                 (format "found lib name '%s' containing period with prefix '%s'. lib names inside prefix lists must not contain periods."
+                                                         form prefix))))
             [(with-meta (token-node full-form)
                (cond-> (assoc (meta libspec-expr)
                               :raw-name form)
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -46,7 +46,7 @@
                                                        libspec-expr
                                                        :error
                                                        :syntax
-                                                       "nested prefix lists are not allowed.")))
+                                                       "Prefix lists can only have two levels.")))
               (mapcat (fn [f]
                         (normalize-libspec ctx
                                            (symbol (str (when prefix (str prefix "."))
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -54,6 +54,14 @@
           (valid-ns-name? form)
           (let [full-form (symbol (str (when prefix (str prefix "."))
                                        form))]
+            (when (and prefix
+                       (str/includes? (str form) "."))
+              (findings/reg-finding! ctx
+                                     (node->line (:filename ctx)
+                                                 libspec-expr
+                                                 :error
+                                                 :syntax
+                                                 "Lib names inside prefix lists must not contain periods.")))
             [(with-meta (token-node full-form)
                (cond-> (assoc (meta libspec-expr)
                               :raw-name form)
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -41,12 +41,18 @@
         children (:children libspec-expr)
         form (sexpr libspec-expr)]
     (cond (prefix-spec? form)
-          (mapcat (fn [f]
-                    (normalize-libspec ctx
-                                       (symbol (str (when prefix (str prefix "."))
-                                                    (first form)))
-                                       f))
-                  (rest children))
+          (do (when prefix
+                (findings/reg-finding! ctx (node->line (:filename ctx)
+                                                       libspec-expr
+                                                       :error
+                                                       :syntax
+                                                       "nested prefix lists are not allowed.")))
+              (mapcat (fn [f]
+                        (normalize-libspec ctx
+                                           (symbol (str (when prefix (str prefix "."))
+                                                        (first form)))
+                                           f))
+                      (rest children)))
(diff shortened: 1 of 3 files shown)


test/clj_kondo/main_test.clj
```diff
@@ -2529,6 +2529,36 @@ foo/foo ;; this does use the private var
 
 " {:linters {:unsorted-required-namespaces {:level :warning}}})))))
 
+(deftest require-prefix-form-containing-periods-test
```
jysandy, line 2532: Should I add these tests into `corpus/prefixed_libspec.clj` instead?

jysandy: Then they'll be tested under https://github.com/clj-kondo/clj-kondo/blob/master/test/clj_kondo/main_test.clj#L487

jysandy: Not entirely sure that's necessary however, it might be simpler to keep them separate.

borkdude: Splitting it out either way is fine with me.

test/clj_kondo/main_test.clj
```diff
@@ -2529,6 +2529,36 @@ foo/foo ;; this does use the private var
...
+  (is (= '({:col     31
+            :file    "<stdin>"
+            :level   :error
+            :message "Lib names inside prefix lists must not contain periods."
+            :row     3})
+         (lint! "(ns baz
+                   (:require [clj-kondo.impl.analyzer
+                              foo.bar]))"
+                {:linters {:unused-namespace {:level :off}}}))))
+
```
borkdude, line 2561: Can you explain what the difference in test is between the first and second one? Maybe clarify using `testing`.

jysandy: The first one contains one valid lib name (`babashka`), whereas the second one does not. On second thought though, it doesn't seem like a valuable test case. I think I'll get rid of it, and replace it with a test case that contains multiple invalid lib names.

jysandy:
I also ran into a case where nested prefix forms are not accepted by the compiler, but clj-kondo doesn't detect them. 
Ex: 
```
(ns quux
  (:require [clj-kondo.impl 
             [analyzer
              [foo]
              [bar]]]))
```
I'll include a fix for this.

test/clj_kondo/main_test.clj
jysandy: This is changed.

test/clj_kondo/main_test.clj
```diff
@@ -486,7 +486,12 @@ foo/foo ;; this does use the private var
 
 (deftest prefix-libspec-test []
   (assert-submaps
-   '({:file "corpus/prefixed_libspec.clj",
+   '({:col 14
+      :file "corpus/prefixed_libspec.clj"
+      :level :error
+      :message "nested prefix lists are not allowed."
```
jysandy, line 492: Not sure if this is a helpful error message. Trying to think of a better one.

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -41,19 +41,34 @@
...
-                                       (symbol (str (when prefix (str prefix "."))
-                                                    (first form)))
-                                       f))
-                  (rest children))
+          (do (when prefix
+                (findings/reg-finding! ctx (node->line (:filename ctx)
+                                                       libspec-expr
+                                                       :error
+                                                       :syntax
+                                                       "nested prefix lists are not allowed.")))
```
jysandy, line 49: Not sure if this is a helpful error message. Trying to think of a better one.

borkdude: If you are using punctuation, can you also start the sentence with an uppercase? What is the case you are finding here?

borkdude: In general I try to use the same message as the Clojure compiler itself does.

jysandy: This is the case of having nested prefix forms, which doesn't seem to be allowed. Ex:
```
(ns quux
    (:require [clj-kondo.impl
                       [analyzer
                         [foo]
                         [bar]]]))
```
This results in a spec failure, the compiler doesn't seem to have a custom error message for it.
```
Syntax error macroexpanding clojure.core/ns at (REPL:1:1).
((:require [foo [bar [baz :as b]] [baz :as c]])) - failed: Extra input spec: :clojure.core.specs.alpha/ns-form
```

borkdude: So the issue is that nested libspecs actually only support one nesting?

jysandy: @borkdude yes, that's right.

borkdude: > Prefix Lists

> It's common for Clojure code to depend on several libs whose names have
the same prefix. When specifying libs, prefix lists can be used to reduce
repetition. A prefix list contains the shared prefix followed by libspecs
with the shared prefix removed from the lib names. After removing the
prefix, the names that remain must not contain any periods.

A nicer message may be "Prefix lists can only have two levels." perhaps?

jysandy: Sounds good to me, I've changed it.

Issue #1129, opened 2021-01-18
Keyword analysis


snoe:
**Is your feature request related to a problem? Please describe.**
Analyzing keywords is important for clojure-lsp.
It allows for completion, renaming, reference searching.

**Describe the solution you'd like**
The solution was to mirror `locals` somewhat. 
The attached PR has tests and README changes to describe the shape and usage.

**Describe alternatives you've considered**
It's possible that standalone keywords could be included in `var-usages` and destructured keyword information `assoc` onto `locals` when using `::keys` (somewhat similar to `alias` info in `namespace-usages`)

**Additional context**
Not addressed in this feature (but this helps) is also where keywords are used as `def`-like identifiers, such as in `spec/def`. 

One thing to note, is that destructured `keyword-usages` overlap with `locals`. I think the addition of the `:keys-destructuring` flag will allow tools to handle the overlap as needed but care will need to be taken as to which bucket the tool is really looking for. e.g. Are you renaming the keyword, the local, or both? Are you looking for references of the keyword, the local, or both?

I may have found a bug: `::clojure.core/inc` is valid according to kondo because `qualify-ns` has a map-entry `[clojure.core clojure.core]`, but clojure will blow up with this code since it is not using a registered alias. I believe kondo should be tracking these, but the TODO's in the tests reflect this inconsistency.

PR Incoming

borkdude:
@snoe Feel free to post an issue about `::clojure.core/inc`, we should give a warning if this isn't used with an alias.

borkdude:
@snoe In `clj-kondo.impl.analyzer.usages` there is a `analyze-keyword` function. Is this something you could re-use/copy instead of the regex approach you are not certain about? 

I haven't looked too deeply into the PR, but I will do so in the coming days, hopefully.

borkdude:
Maybe the analysis output can just be called `:keywords` as there is no usage vs definition.

snoe:
In clojure-lsp it is useful to call `(s/def ::spec)` a definition. There are a number of community libraries (re-frame in particular) that also treat keywords as defacto definitions. I would love to use sci on `(re-frame/reg-fx ::handler)` to be able to mark `::handler` as a definition somehow.

Even `clojure.core/derive` could look a lot like a definition.

[...]

PR #1130, opened 2021-01-18
Adding keyword analysis option to analyzer

snoe:
Addresses #1129

There are TODOs about :aliases I described in the issue. 

I tried very hard to use `namespace/resolve-name` instead of the regex, but the way aliases are handled by the clojure reader completely stumped me. And I wasn't entirely sure that's the right place anyhow.


src/clj_kondo/impl/analysis.clj
```diff
@@ -96,3 +96,10 @@
                        :filename filename
                        :lang (when (= :cljc (:base-lang ctx)) (:lang ctx))
                        :id (:id binding)))))
+
+(defn reg-keyword-usage! [{:keys [:analysis] :as ctx} filename usage]
```
borkdude, line 100: Minor nitpick, but when using only one key it's generally better for performance not to use destructuring. Just use `(when-let [analysis (:analysis ctx)] ...)`

src/clj_kondo/impl/analyzer/spec.clj
```diff
@@ -31,6 +31,11 @@
                                                      (str "Unresolved symbol: " sym)))))))
     (analyze-children ctx body)))
 
+(defn analyze-def [{:keys [analyze-children resolved] :as ctx} expr]
+  (let [[kw-expr & body] (next (:children expr))]
+    (let [fq-def (symbol (-> resolved first name) (-> resolved second name))]
+      (analyze-children ctx (cons (assoc kw-expr :def fq-def) body)))))
+
```
snoe, line 38: @borkdude I'm not sure if there's a better way to get the fully qualified name of the call?

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1532,6 +1532,11 @@
                         (test/analyze-cljs-test-async ctx expr)
                         ([clojure.test are] [cljs.test are] #_[clojure.template do-template])
                         (test/analyze-are ctx expr)
+                        ([clojure.spec.alpha def] [cljs.spec.alpha def])
```
snoe, line 1535: @borkdude Is this the right way to build in the s/def support? Or should I define a default  `:hooks` config for these somehow?  (I couldn't find anything else doing that, so I went this path).

borkdude: Yes, this is how built-in support for core macros is done.

src/clj_kondo/impl/analyzer/spec.clj
borkdude: If you are using clj-kondo you will notice you have a redundant let here ;) 

I'll check out the branch and will make some changes locally, to play with it. Would you mind giving me push rights to this PR branch, perhaps?

borkdude: @snoe Instead of passing `analyze-children` to this function, you can now use `clj-kondo.impl.analyzer.common/analyze-children`, this is set up in a way to avoid circular dependencies.

borkdude:
@snoe I think you can avoid the regex stuff using the following:

```
(prn (:namespaced? expr))
(prn (:k expr))
```

`:namespaced?` is only true when the keyword is used with a double colon. This is a pretty bad name, but I think it came from rewrite-clj originally.
`:k`: this has the raw keyword.

In case of `::foo`, `:namespaced?` is true, and `:k` is `:foo`. In absence of a namespace of the `:k` value, you can assume the current namespace was used.

In case of `::s/foo`, `:namespaced?` is true, and `:k` is `:s/foo`. Now `s` must be an alias since you cannot have `::clojure.core/foo` for example.

In case of `:s/foo`, `:namespaced?` is false or nil, and `:k` is `:s/foo`.  The namespace was a fully qualified one.

In case of `:foo`, `:namespaced?` is false, and `:k` is `:foo`.  An unqualified keyword.

src/clj_kondo/impl/analyzer/spec.clj
snoe: it's hooked up for both `clojure.spec.alpha/def` and `cljs.spec.alpha/def`, I suppose the clojure version might be fine?

I didn't realize it could take a symbol, I'll update, thanks.

snoe: I split the `case` to just pass in the literals, I believe this should be good now.

snoe:
@borkdude I think the last commit addressed all of the open comments here.

analysis/README.md
```diff
@@ -20,6 +20,7 @@ Further analysis can be returned by providing `:analysis` with a map of options:
 ```
 
 - `:locals`: when truthy return `:locals` and `:local-usages` described below
+- `:keywords`: when truthy return `:keywords` described below
```
borkdude, line 23: Can you also comment this out using HTML comments, also the docs below?
I think the docs are good, but let's not expose this feature to others until we are sure this is the right choice (everything in one list, no separation of defs, etc).

src/clj_kondo/impl/analysis.clj
```diff
@@ -97,3 +97,10 @@
                        :filename filename
                        :lang (when (= :cljc (:base-lang ctx)) (:lang ctx))
                        :id (:id binding)))))
+
+(defn reg-keyword-usage! [ctx filename usage]
+  (when-let [analysis (:analysis ctx)]
```
borkdude, line 102: Should we also check for `:analyze-keywords?` here?

src/clj_kondo/impl/hooks.clj
```diff
@@ -17,6 +17,14 @@
         filename (:filename ctx)]
     (findings/reg-finding! ctx (assoc m :filename filename))))
 
+(defn reg-keyword!
+  ([k]
+   (reg-keyword! k true))
+  ([k reg-by]
+   (assoc k :def (if (qualified-symbol? reg-by)
```
borkdude, line 24: why not just pass through reg-by here, regardless of what it is?

borkdude:
@snoe A few minor questions.

src/clj_kondo/impl/analysis.clj
snoe: done

snoe:
@borkdude good points, thanks. How's it look now?

analysis/README.md
borkdude: @snoe ^

PR #1132, opened 2021-01-18
Add :ns to unused-namespaces findings

ericdallo:
ATM we can't get the namespace/import/refer symbol from a `finding` so this PR:

- add `:ns` to the `unused-namespace` finding
- add `:refer` to the `unused-refer` finding
- add `:import` to the `unused-import` finding
- migrate unused-namespace tests from `main_test.clj` -> `unused_namespace_test.clj`


test/clj_kondo/unused_import_test.clj
```diff
@@ -81,4 +82,19 @@
...
+  (testing "return import info"
+    (assert-submaps
+      '({:type :unused-import,
+         :message "Unused import Date",
+         :level :warning,
+         :row 1,
+         :end-row 1,
+         :end-col 33,
+         :col 29,
+         :import java.util.Date
```
borkdude, line 94: I suggest renaming `:import` to `:class` and leave out the `:class` since this is redundant info.

ericdallo:
Thank you!

Issue #1131, opened 2021-01-18
Add `:alias` key to `var-usages`


snoe:
**Is your feature request related to a problem? Please describe.**
In order to rename an alias in a namespace-usage, clojure-lsp needs to know all the occurrences of that alias. Right now it does not seem possible to differentiate between a var that uses the fully qualified namespace or the alias.

**Describe the solution you'd like**
For the var-usages that refer to an external var using an alias, attach an `:alias` keyword to the analyzed data.

So in the following code, the usage for `bar/a` would not have an alias, and `b/b` would have a ns of `bar` and alias of `b`

```clojure
(ns foo (:require [bar :as b]))
bar/a 
b/b
```

I'm not sure if this should be behind a feature flag or not? 

**Describe alternatives you've considered**

There could be an `alias-usages` bucket for these things, but as far as I know aliases are only used by `keywords` (added in #1129) and `var-usages`.

PR incoming, hopefully in a few days

borkdude:
The proposed solution seems good to me.

borkdude:
Will review tomorrow! Getting late here :)

ericdallo:
I think we can close this, I tested and fix our clojure-lsp issue here https://github.com/clojure-lsp/clojure-lsp/pull/261/files :)

PR #1133, opened 2021-01-18
Add :aliases to ns ctx and :alias to var-usages

snoe:
Addresses #1131

### DIFF
--- a/src/clj_kondo/impl/namespace.clj
+++ b/src/clj_kondo/impl/namespace.clj
@@ -207,8 +207,11 @@
 
 (defn reg-alias!
   [{:keys [:base-lang :lang :namespaces]} ns-sym alias-sym aliased-ns-sym]
-  (swap! namespaces assoc-in [base-lang lang ns-sym :qualify-ns alias-sym]
-         aliased-ns-sym))
+  (swap! namespaces
+         (fn [n]
+           (-> n
+               (assoc-in [base-lang lang ns-sym :qualify-ns alias-sym] aliased-ns-sym)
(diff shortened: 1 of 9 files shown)


src/clj_kondo/impl/namespace.clj
```diff
@@ -207,8 +207,11 @@
 
 (defn reg-alias!
   [{:keys [:base-lang :lang :namespaces]} ns-sym alias-sym aliased-ns-sym]
-  (swap! namespaces assoc-in [base-lang lang ns-sym :qualify-ns alias-sym]
-         aliased-ns-sym))
+  (swap! namespaces
+         (fn [n]
+           (-> n
+               (assoc-in [base-lang lang ns-sym :qualify-ns alias-sym] aliased-ns-sym)
+               (assoc-in [base-lang lang ns-sym :aliases alias-sym] aliased-ns-sym)))))
```
borkdude, line 214: This can be made more efficient by using `(update-in [base-lang lang ns-sym] assoc :qualify-ns ... :aliases ...)`

snoe: @borkdude sorry, I'm having trouble seeing this, maybe I'm missing something obvious. It's tricky because `alias-sym` is a key on siblins. Are you suggesting something like this?

```
(update-in [base-lang lang ns-sym] 
  #(-> % 
    (update :qualify-ns assoc alias-sym aliased-ns-sym)
    (update :aliases assoc alias-sym aliased-ns-sym)))
```

borkdude: You're right, that gets convouted, let's leave it.

PR #1147, opened 2021-01-26
Add end-row and end-col to var-definitions bucket on analysis

ericdallo:
For clojure-lsp analysis, is useful to know what is the var-definition end range, with that we can return a range to LSP client to use for outline/headerlines.

Without that, we can't know the end range so when user move cursor inside the function, we can't know if it's inside the var-definition.

image

image (1)


ericdallo:
Thank you ;)

PR #1150, opened 2021-01-28
Redundant negation

### LINKED ISSUE #1080: filter/remove used with not/complement
We often see `filter` used with `not` when `remove` should be used:
`(filter (fn [x] (not (string? x))) ["string" 5 true []])`
which should be:
`(remove (fn [x] (string? x)) ["string" 5 true []])`
Similarly, this problem can happen with `complement`:
`(filter (complement string?) ["string" 5 true []])`
which should be:
`(remove string? ["string" 5 true []])`
People seem to forget `remove` more than they forget `filter` but we may as well fix it for both ways around:
`(remove (fn [x] (not (string? x))) ["string" 5 true []])`
should be:
`(filter (fn [x] (string? x)) ["string" 5 true []])`

I'd like clj-kondo to spot whenever a fn is supplied to `filter` or `remove` where the top level inside the fn body is `(not...)` or when `(complement foo)` is supplied.

As @SevereOverfl0w said recently, "Writing better code first time around" 😂

tomdl89:
This resolves #1080 and as discussed there, I've tried to cover all reasonable negation using clojure core I could envisage . Although it's quite big, I realised it could have been almost endless, as we could warn against `(not (not foo))` or `(not (not-any? foo? bars))` which both seem to me not worth doing. I think it's reasonable to say all these linters can be grouped under one name, but I'm open to discussion about splitting them out if that seems suitable.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -84,6 +84,7 @@ Some linters are not enabled by default. Right now these linters are:
 - `:unsorted-required-namespaces`: warn when namespaces in `:require` are not sorted.
 - `:refer`: warn when there is **any** usage of `:refer` in your namespace requires.
 - `:single-key-in`: warn when using assoc-in, update-in or get-in with single key.
+- `:redundant-negation`: warn on redundant use of `not` or `complement`.
 - `:shadowed-var`: warn when a binding shadows a var.
 
 You can enable these linters by setting the `:level`:
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -30,6 +30,7 @@ Table of contents:
   - [Redefined var](#redefined-var)
   - [Redundant do](#redundant-do)
   - [Redundant let](#redundant-let)
+  - [Redundant negation](#redundant-negation)
   - [Refer](#refer)
   - [Refer all](#refer-all)
   - [Single key in](#single-key-in)
@@ -491,6 +492,29 @@ because directly nested lets.
 *Example trigger:* `(let [x 1] (let [y 2] (+ x y)))`.
 
 *Example message:* `Redundant let expression.`
+
+### Redundant negation
+
+*Keyword:* `:redundant-negation`.
+
+*Description:* warn on usage of `not` or `complement` that is redundant. This includes
+`and` & `or` used with multiple `not`s, negated core fns where their complement is
+already available, and `filter` and `remove` used with negated fns. *N.b.* `:not-empty`
+is a separate (and default enabled) linter.
+
+*Default level:* `:off`.
+
+*Example trigger:* `(and (not (string? "foo")) (not (keyword? :bar)))`.
+
+*Example message:* `And & 2 nots used instead of 1 not with or`.
+
+*Example trigger:* `(not (nil? foo))`.
+
+*Example message:* `not and nil? used instead of some?`.
+
+*Example trigger:* `(filter #(not (string? %)) [:foo "bar" 42])`.
+
+*Example message:* `filter and not used instead of remove`.
 
 ### Refer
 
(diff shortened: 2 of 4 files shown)


tomdl89:
Added a commit which keeps linting `if-not` and `when-not` for redundant negation, but only that which results in double-negation. E.g. we warn against `(if-not (some? foo))` because it will expand to `(if-not (not (nil? foo)))` but we don't warn against `(if-not (nil? foo))`. This fixes the problem of raising a (seemingly false-) positive warning for something like `(if-not (= foo bar))` which is arguably a stylistic preference (over `(if (not= foo bar))`). Not sure if this is where we want to go with `if-not` and `when-not` or we'd rather just remove linting for redundant negation on these forms..?

borkdude:
So far that sounds reasonable.

borkdude:
@tomdl89 I haven't looked at the implementation so far, but now I did. I think you can implement this using the `:callstack` which is available in a `:call`. I think that will be more performant than walking the AST twice, which should always be avoided when possible. When you print, you get roughly this:

```
user=> (with-out-str (with-in-str "(not (not 1))" (clj-kondo/run! {:lint ["-"]})))
"([clojure.core not] [clojure.core not])\n([clojure.core not])\n"
user=> (with-out-str (with-in-str "(when-not (not 1))" (clj-kondo/run! {:lint ["-"]})))
"([clojure.core not] nil [clojure.core when-not])\n([clojure.core when-not])\n"
user=> (with-out-str (with-in-str "(if-not (not 1))" (clj-kondo/run! {:lint ["-"]})))
"([clojure.core not] [clojure.core if-not])\n([clojure.core if-not])\n"
```

tomdl89:
@borkdude sorry for the delay. Having looked at this, I'm not sure I follow. How are we "walking the AST twice" with my approach? I'm just adding on to the `lint-specific-calls!` fn, which is called on all used, valid vars anyway, no? I can see the `:callstack` is available to me, but so are the `:children` - what's the difference? Thanks

borkdude:
@tomdl89 I think it could simplify the code if this was implemented using `:callstack` which already establishes the relation between the current call and the parent call. This is also how `:redundant-do` and `:redundant-let` are implemented.

https://github.com/clj-kondo/clj-kondo/blob/156fdf5bfb4b6ebe037de527e478697bc9973960/src/clj_kondo/impl/analyzer.clj#L645-L648

borkdude:
About the name of this linter: I think `:redundant-negation` isn't the proper name perhaps, since it's more of a double negation than that one of the negations can be omitted. So maybe `:double-negation` is better? That's still not the right name. Maybe `:simplified-negation`? This needs more hammock time too =).

borkdude:
@tomdl89 I hope to take a look myself at the :callstack approach hopefully within a few days.

tomdl89:
Hi @borkdude that sounds great. Sorry I've stalled a bit on this. I'm really interested to see what the callstack implementation can look like for this.

borkdude:
@tomdl89 No worries :)

borkdude:
I pulled this branch, merged it with master and pushed it to a branch in this project. Perhaps you can point this PR against that branch and merge it with that branch.

I tried some cases which didn't work yet:

```
$ clj-kondo --lint - --config '{:linters {:redundant-negation {:level :warning}}}' <<< '(when-not (not 1) 1)'
linting took 11ms, errors: 0, warnings: 0
```

```
$ clj-kondo --lint - --config '{:linters {:redundant-negation {:level :warning}}}' <<< '(not (not 1))'
linting took 10ms, errors: 0, warnings: 0
```

I moved the tests to a separate namespace.

I didn't work on the `:callstack` approach yet.

PR #1155, opened 2021-01-31
Fixes #1153 

### LINKED ISSUE #1153: Keyword analysis namespace not reported correctly when using alias
``` clojure
(require '[clojure.string :as s])

:clojure.string/foo
::s/foo
```

The second keyword should be reported with namespace `clojure.string` and alias `s`.

Right now it gives:

``` clojure
[{:col 1,
  :end-col 20,
  :end-row 3,
  :filename "/tmp/keywords.clj",
  :name foo,
  :ns clojure.string,
  :row 3}
 {:col 1,
  :end-col 8,
  :end-row 4,
  :filename "/tmp/keywords.clj",
  :name foo,
  :ns :clj-kondo/unknown-namespace,
  :row 4}]
```

cc @snoe

snoe:
Had to use `get-namespace` because `(:ns ctx)` doesn't get updated on `analyze-require` (and probably the other namespace-altering fns).

### DIFF
--- a/src/clj_kondo/impl/analyzer/usages.clj
+++ b/src/clj_kondo/impl/analyzer/usages.clj
@@ -21,7 +21,8 @@
         alias-or-ns (some-> token namespace symbol)
         ns-sym (cond
                  (and aliased? alias-or-ns)
-                 (get-in ctx [:ns :aliases alias-or-ns] :clj-kondo/unknown-namespace)
+                 (-> (namespace/get-namespace ctx (:base-lang ctx) (:lang ctx) current-ns)
+                     (get-in [:aliases alias-or-ns] :clj-kondo/unknown-namespace))
 
                  aliased?
                  current-ns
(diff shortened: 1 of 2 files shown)


borkdude:
@snoe Thanks for the fix!

What do you think about the following:

```
$ clojure -M:clj-kondo --config '{:output {:analysis {:keywords true} :format :edn}}' --lint - <<< '::foo' | jet --query ':analysis :keywords' | puget
[{:col 1,
  :end-col 6,
  :end-row 1,
  :filename "<stdin>",
  :name foo,
  :ns user,
  :row 1}]
```

In the analysis, you cannot see that the keyword was used with `::foo`. Should we indicate this with `:auto-resolved true` perhaps? `:auto-resolved <alias>` can then mean: this keyword was autoresolved using the alias and `true` means auto-resolved in the current namespace.

Issue #1174, opened 2021-02-20
schema.core/defmethod linting error when dispatch-val is a vector


leoiacovini:
**version**

v2021.02.13

**platform**

JVM and via Brew

**editor**

VSCode with Calva

**problem**

`schema.core/defmethod` linting is not working properly when the `dispatch-val` is a `vector` (maps and lists work great).

**repro**

```clojure
(defmulti my-fn (fn [m] [(:a m) (:b m)]))

(sc/defmethod my-fn [:a :b] :- s/Keyword
  [m :- (s/pred map?)]
  :hello)
```

clj-kondo log:

```
error: unsupported binding form :-
error: unsupported binding form (s/pred map?)
```

**expected behavior**

To work regularly just like when using other types as `dispatch-val`s

Proposal PR: https://github.com/clj-kondo/clj-kondo/pull/1175

PR #1175, opened 2021-02-20
Fix schema.core/defmethod linting for vectors dispatch-values

leoiacovini:
Closes: https://github.com/clj-kondo/clj-kondo/issues/1174

--

I preferred to go with the approach to track the `fn-sym` and `index` instead of trying to infer when the vector was supposed to be the arguments or the dispatch value (for any schema.core macro) because the inferring logic was going to be more complex (in the code or even computationally) and would probably still have some bugs in corner cases when using the `:-` keyword inside function bodies

### DIFF
--- a/corpus/schema/defmethod.clj
+++ b/corpus/schema/defmethod.clj
@@ -1,7 +1,7 @@
 (ns schema.defmethod
   (:require
-   [schema.core :as sc]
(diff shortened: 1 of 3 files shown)


borkdude:
Thanks!

PR #1194, opened 2021-03-05
Update unresolved-symbol link in config.md

eccentric-j:
Original unresolved-symbol link in config.md went to a non-existent hash, updated it to the intended doc


borkdude:
Thanks!

PR #1214, opened 2021-03-12
chore(editor-integration): improve documentation

Valdermeyder:
`$FileDir$` si a default value for the Working directory field, so we can omit it.
It is more useful to use the root project directory `$ProjectFileDir$` for example to access a custom config file location

### DIFF
--- a/doc/editor-integration.md
+++ b/doc/editor-integration.md
@@ -228,8 +228,8 @@ and CLJC (`.cljc`)<sup>1</sup>.
 4. Scope: `Current file`
 5. In the Program field, type `clj-kondo`
 6. In the Arguments field, type `--lint $FilePath$`<br>
-You may use a custom config E.g `--lint $FilePath$ --config "{:lint-as {manifold.deferred/let-flow clojure.core/let}}"`
-7. In the Working directory field, type `$FileDir$`
+You may use a custom config E.g `--lint $FilePath$ --config "{:lint-as {manifold.deferred/let-flow clojure.core/let}}"` or `--lint $FilePath$ --config your-custom-path/clj-kondo.conf.js"`
+7. In the Working directory field, type `$ProjectFileDir$`
 8. Enable `Create output file from stdout`
 9. Show console: `Never`
 10. In output filters put `$FILE_PATH$:$LINE$:$COLUMN$: $MESSAGE$`


doc/editor-integration.md
```diff
@@ -228,8 +228,8 @@ and CLJC (`.cljc`)<sup>1</sup>.
 4. Scope: `Current file`
 5. In the Program field, type `clj-kondo`
 6. In the Arguments field, type `--lint $FilePath$`<br>
-You may use a custom config E.g `--lint $FilePath$ --config "{:lint-as {manifold.deferred/let-flow clojure.core/let}}"`
-7. In the Working directory field, type `$FileDir$`
+You may use a custom config E.g `--lint $FilePath$ --config "{:lint-as {manifold.deferred/let-flow clojure.core/let}}"` or `--lint $FilePath$ --config your-custom-path/clj-kondo.conf.js"`
```
borkdude, line 231: The .js extension seems wrong. In hindsight, I think it's better to remove the config description, as this should be apparent from the general usage of clj-kondo.

imrekoszo:
LGTM. Updated my settings to ProjectFileDir and it appears to work okay. I don't use a custom config path so can't review that.
Screenshot 2021-03-12 at 15 14 52

doc/editor-integration.md
```diff
@@ -228,8 +228,8 @@ and CLJC (`.cljc`)<sup>1</sup>.
...
+7. In the Working directory field, type `$ProjectFileDir$`
```
borkdude, line 232: In case of a mono-repo what is the `$ProjectFileDir` in Cursive? Using `FileDir` clj-kondo will see configs at a higher level than the root directory of the mono-repo.

borkdude:
@imrekoszo Thanks. Do you know the answer to:

https://github.com/clj-kondo/clj-kondo/pull/1214#discussion_r593204398 ?

doc/editor-integration.md
Valdermeyder: you are right about the extension, it's a typo

doc/editor-integration.md
Valdermeyder: Hard to say about monorepo, I don't have any examples. If I understand your doubts correctly each project inside monorepo could have its own clj-kondo config and IntelliJ IDEA default macro `$FileDir$` will find the closest one and you would like to keep this behavior.
If it is the main concerns we can just leave an example with a custom file config and remove step 7 as it is the default IntelliJ IDEA behavior
If this PR has no sense at all I'm fine to close it as well

imrekoszo: In the only monorepo project I work on it points to the root of the monorepo. I have a deps.edn there and have deps.edns in child folders (subprojects).

borkdude: Then it is better to leave it as is.

Valdermeyder:
This PR is not needed anymore as discussed above

PR #1219, opened 2021-03-17
[analysis] Add defined-by on missing var definitions

### LINKED ISSUE #1216: [analysis] missing defined-by on most var-definition bucket analysis
I'd like to know which `defined-by` a var-defintion has. ATM clj-kondo only send that for `clojure.test/deftest`

Maybe always include the `defined-by` with the function/macro symbol that defines it?

**Additional context**
[Discussion](https://clojurians.slack.com/archives/CHY97NXE2/p1615997011038300)

ericdallo:
Fixes #1216 

This should add the `:defined-by` with the full-qualified-symbol for all `var-definitions`.

### DIFF
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -51,6 +51,7 @@ The analysis output consists of a map with:
   - `:filename`, `:row`, `:col`, `end-row`, `end-col`
   - `:ns`: the namespace of the var
   - `:name`: the name of the var
+  - `:defined-by`: the namespaced symbol which defined this var
 
   Optional:
   - `:fixed-arities`: a set of fixed arities
@@ -148,12 +149,14 @@ $ clj-kondo --lint /tmp/foo.clj --config '{:output {:analysis true :format :edn}
                     :end-col 11,
                     :ns foo,
                     :name f,
+                    :defined-by 'clojure.core/defn-
                     :private true,
                     :fixed-arities #{1}}
                    {:added "1.2",
                     :ns foo,
                     :name g,
                     :filename "/tmp/foo.clj",
+                    :defined-by 'clojure.core/defmacro
                     :macro true,
                     :row 11,
                     :col 1,
--- a/corpus/hooks/state_flow.clj
+++ b/corpus/hooks/state_flow.clj
@@ -1,5 +1,0 @@
-(ns baz
-  {:clj-kondo/config '{:hooks {:analyze-call {state-flow.cljtest/defflow hooks.state-flow/defflow}}}}
-  (:require [state-flow.cljtest :refer [defflow]]))
-
-(defflow my-flow)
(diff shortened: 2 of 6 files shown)


ericdallo:
@borkdude I don't know if that is going to work with hook var-defintions like `state-flow/defflow`, could you point me to the function that should `reg-var!` that?

borkdude:
@ericdallo So you wrote a hook for state-flow/defflow but after expansion it's linted as normal clojure.core/defn right? Is that the problem? Btw, some tests are failing.

ericdallo:
Fixed tests, sorry for that!

@borkdude yes, and I just realized that the [hook is using a `def` node](https://github.com/nubank/state-flow/blob/master/resources/clj-kondo.exports/nubank/state-flow/nubank/state_flow.clj#L31), so I imagine that its `defined-by` will be `def`, right?

how can we make that work? is there anything we can add to the hook to tell clj-kondo to use a different `defined-by`?

ericdallo:
BTW, I think this PR should solve https://github.com/clojure-lsp/clojure-lsp/issues/340 as well :)

borkdude:
@ericdallo We could return an extra value in the hook, `{:node ... :defined-by ...}` but a more general name is maybe better...

ericdallo:
@borkdude it makes sense, any tips how to implement that?
I checked that [here](https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/analyzer.clj#L1385) we could get the `defined-by` informed by user via the `transformed` local and I suppose we should pass to [analyze-expression**](https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/analyzer.clj#L1420) and pass to some reg-var somewhere?

borkdude:
@ericdallo We can maybe put it on the `ctx` argument here: https://github.com/clj-kondo/clj-kondo/blob/3f63419f29915018fff4a1611458a6699ca53cad/src/clj_kondo/impl/analyzer.clj#L1420

ericdallo:
That [analyze-expression**](https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/analyzer.clj#L1420) is used by a lot of places and the places that `reg-var!`, call it before calling that function, so I think maybe we should `reg-var!` before calling the `analyze-expression**`?
I'm kind of lost on that code :p

borkdude:
@ericdallo My suggestion was to pass the `:defined-by ...` value as part of the `ctx`, so `(assoc ctx :defined-by defined-by)` and then pick up on it again in `analyze-call` if it's available. Not sure if that is the right approach, but worth a try. Since `defs` are usually not nested, this may work.

ericdallo:
@borkdude just tested your suggestion with the hook and it worked like a charm :)
I changed the hook to:
```clojure
...
{:node (with-meta new-node (meta node))
 :defined-by 'state-flow.cljtest/defflow}
```
And clojure-lsp got it correctly!

borkdude:
@ericdallo Cool! Maybe add a test for it and then we can merge?

ericdallo:
Sure, shoul I create a test for the stateflow macro?

borkdude:
Just capturing the behavior that hooks can return that `:defined-by` key.

ericdallo:
Done @borkdude, but the tests are passing local but not on the CI, it seems related with the config dir

ericdallo:
Finally fixed @borkdude! IMO we are good to go

PR #1221, opened 2021-03-18
Change no-warnings to copy-hooks check when copying hooks

ericdallo:
This will allow clj-kondo to copy hooks while returning findings/analysis, used by clojure-lsp for example.

This can be considered a breaking-change, not sure we should keep checking for no-warnings along with the new flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+### New
+
+- `--copy-hooks` flag to indicate linting is used to copy existing hooks to clj-kondo config dir.
+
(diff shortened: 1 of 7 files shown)


sogaiu:
@ericdallo Is it correct that the old name `no-warnings` will no longer be usable?

If so, I guess that means if I've written instructions in various places it might be good to track these down to update them...

Is it necessary to remove the old name?

borkdude:
@sogaiu I have deprecated the name `--no-warnings` in favor of `--dependencies` but the old one still works.

borkdude:
Additionally there is now a flag `--copy-configs` to control of config should be copied.

sogaiu:
Thanks for the clarification!

PR #1233, opened 2021-03-25
Fixes #1232 introduce report-duplicates linter config

### LINKED ISSUE #1232: Optionally report duplicates
In clojure-lsp it's possible to edit a file to produce many unresolved-var or unresolved-symbol lints, but the current behaviour is to only report the first occurence and swallow to subsequent ones. So a user has to find errors, fix the first one then find errors again, then fix the next etc... Many editors provide good support to navigate to all your diagnostics and this behavior neuters this functionality.

Spoke with @borkdude, we'd like to add a config flag on each affected linter `report-duplicates` that when nil (default) or otherwise non-truthy would maintain the existing behaviour and when truthy would return all of the lint occurrences.

Talked about top level configuration for the desired behaviour but since most linters already report duplicates this seems like a nice, specific solution that can be extended anywhere ignoring duplicates is introduced later.

PR incoming

snoe:
Addresses #1232

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -801,6 +801,12 @@ and helps preventing false positive unresolved symbols in this code:
   ,,,)
 ```
 
+You can report-duplicate warnings using:
+
+``` clojure
+{:linters {:unresolved-var {:report-duplicates true}}}
+```
+
 ### Unresolved var
 
 *Keyword:* `:unresolved-var`.
@@ -836,6 +842,12 @@ or exclude a selection of unresolved vars using qualified symbols:
 
 ``` clojure
 {:linters {:unresolved-var {:exclude [foo/x]}}}
+```
+
+You can report-duplicate warnings using:
+
+``` clojure
+{:linters {:unresolved-var {:report-duplicates true}}}
 ```
 
 ### Unsorted required namespace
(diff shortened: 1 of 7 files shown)


borkdude:
Looks good.

Maybe pull `(fnil conj [])` to a top-level function (I usually call it `(def vconj (fnil conj []))`) to avoid creating it on every call? I'm not sure if that helps, but I usually do that anyway.

Other thing: I think `unresolved-namespace` also doesn't report duplicates.

borkdude:
@snoe Just to make sure this isn't lingering: I'm waiting for you to update the `:unresolved-namespace` linter with the same option. No hurry, just making sure you didn't miss it.

snoe:
Oh I think we got crossed wires, @borkdude https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/namespace.clj#L326 This code isn't using a map to dedupe like the other linters and where the doseq happens to pull them out it seems like it would report all of them.

borkdude:
@snoe I'm seeing only one warning here:

Screenshot 2021-03-31 at 17 23 11

That is because we are saving unresolved namespaces in a set in analyzer/namespace.clj, line 348:

```
:unresolved-namespaces #{}
```

snoe:
thanks for the pointer @borkdude - that kept going over my head.

PR #1255, opened 2021-04-17
Report reader errors at the start of token

yuhan0:
This allows editor plugins to place the squiggly lines below the entire token instead of right at the end.

### DIFF
--- a/parser/clj_kondo/impl/rewrite_clj/parser/token.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/parser/token.clj
@@ -39,23 +39,24 @@
 (defn parse-token
   "Parse a single token."
   [reader]
-  (try
-    (let [first-char (r/next reader)
-          s (->> (if (= first-char \\)
-                   (read-to-char-boundary reader)
-                   (read-to-boundary reader))
-                 (str first-char))
-          v (r/string->edn s)]
-      (if (symbol? v)
-        (symbol-node reader v s)
-        (node/token-node v s)))
-    (catch Exception e
-      (let [{:keys [:type :ex-kind]} (ex-data e)]
-        (when (and *invalid-token-exceptions*
-                   (= :reader-exception type)
-                   (= :reader-error ex-kind))
-          (let [f {:row (rt/get-line-number reader)
-                   :col (rt/get-column-number reader)
-                   :message (.getMessage e)}]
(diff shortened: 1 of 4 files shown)


yuhan0:
Note: this causes a 1% slowdown in performance, probably because `parse-token` is now extracting the row and column on every token instead of on exception only.
I tried expanding out the `juxt`into separate bindings, but that made no difference when testing locally.

borkdude:
I think adding the row and col in a separate binding makes sense as it saves the overhead of juxt + creating a vector. 
Have you tested performance on various big Clojure files?

yuhan0:
>  Have you tested performance on various big Clojure files?
I'm not sure how to do that, I've only run the `script/diff` which I think compares the performance on several big repos with the master branch.
The results (before and after refactoring juxt)
```
< linting took 17813ms, errors: 264, warnings: 2292
---
> linting took 17718ms, errors: 264, warnings: 2292
```
```
< linting took 17678ms, errors: 264, warnings: 2292
---
> linting took 16590ms, errors: 264, warnings: 2292
```

borkdude:
@yuhan0 You could try to download a binary from the below builds (CircleCI under artifacts) and then run the binary over clj-kondo itself. This takes around 1100ms on my machine with the most recent release. I will also test with your branch version.

borkdude:
@yuhan0 Oh, we skip the macOS build on branches. So you would have to build yourself locally.

borkdude:
I don't see a significant difference:

```
borkdude@MBP2019 ~/Dropbox/dev/clojure/clj-kondo (yuhan0-fix-reader-error-loc) $ /usr/local/bin/clj-kondo --lint src:test
linting took 1122ms, errors: 0, warnings: 0
borkdude@MBP2019 ~/Dropbox/dev/clojure/clj-kondo (yuhan0-fix-reader-error-loc) $ ./clj-kondo --lint src:test
linting took 1127ms, errors: 0, warnings: 0
borkdude@MBP2019 ~/Dropbox/dev/clojure/clj-kondo (yuhan0-fix-reader-error-loc) $ /usr/local/bin/clj-kondo --lint src:test
linting took 1135ms, errors: 0, warnings: 0
borkdude@MBP2019 ~/Dropbox/dev/clojure/clj-kondo (yuhan0-fix-reader-error-loc) $ ./clj-kondo --lint src:test
linting took 1150ms, errors: 0, warnings: 0
```

PR #1260, opened 2021-04-21
Improve keywords analysis

### LINKED ISSUE #1156: Keyword analysis: auto-resolved field
@snoe 

What do you think about the following:

```
$ clojure -M:clj-kondo --config '{:output {:analysis {:keywords true} :format :edn}}' --lint - <<< '::foo' | jet --query ':analysis :keywords' | puget
[{:col 1,
  :end-col 6,
  :end-row 1,
  :filename "<stdin>",
  :name foo,
  :ns user,
  :row 1}]
```

In the analysis, you cannot see that the keyword was used with `::foo` instead of `:user/foo`. Should we indicate this with `:auto-resolved true` perhaps? `:auto-resolved <alias>` can then mean: this keyword was autoresolved using the alias and true means auto-resolved in the current namespace. The `:alias` can then disappear probably.

ericdallo:
Fix #1156

- Add `auto-resolved` field to keyword analysis
- Add `namespace-from-prefix` field to keyword analysis
- Fix ns overide for namespace maps like `#:a{:b/c 1}` the correct `:ns` should be `:b`  not `:a`

### DIFF
--- a/test/clj_kondo/analysis_test.clj
+++ b/test/clj_kondo/analysis_test.clj
@@ -137,16 +137,17 @@
             {:row 1, :col 11, :end-row 1, :end-col 13, :name "b", :filename "<stdin>"}]
           (:keywords a))))
     (testing "auto-resolved"
-      (let [a (analyze ":a ::b :user/c #:d{:e 1 :_/f 2 :g/h 3 ::i 4}"
+      (let [a (analyze "(ns foo)
+                        :a ::b :bar/c #:d{:e 1 :_/f 2 :g/h 3 ::i 4}"
                        {:config {:output {:analysis {:keywords true}}}})]
         (assert-submaps
               '[{:name "a" :auto-resolved false}
-                {:name "b" :auto-resolved true}
-                {:name "c" :auto-resolved false}
-                {:name "e" :auto-resolved false}
+                {:name "b" :auto-resolved true :ns foo}
+                {:name "c" :auto-resolved false :ns bar}
+                {:name "e" :auto-resolved false :ns d}
                 {:name "f" :auto-resolved false}
-                {:name "h" :auto-resolved false}
-                {:name "i" :auto-resolved true}]
+                {:name "h" :auto-resolved false :ns g}
+                {:name "i" :auto-resolved true :ns foo}]
               (:keywords a))))))
 
 (deftest locals-analysis-test
--- a/test/clj_kondo/analysis_test.clj
+++ b/test/clj_kondo/analysis_test.clj
@@ -127,33 +127,32 @@
     (testing "no namespace for key :a"
       (let [a (analyze "#:xml{:_/a 1}"
                        {:config {:output {:analysis {:keywords true}}}})]
-        (is (= '[{:row 1, :col 7, :end-row 1, :end-col 11, :name "a", :filename "<stdin>" :auto-resolved false :namespace-from-prefix false}]
+        (is (= '[{:row 1, :col 7, :end-row 1, :end-col 11, :name "a", :filename "<stdin>"}]
                (:keywords a)))))
     ;; Don't use assertmap here to make sure ns is absent
     (testing "no namespace for key :b"
       (let [a (analyze "#:xml{:a {:b 1}}"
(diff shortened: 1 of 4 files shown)


ericdallo:
@borkdude I could not add the `namespace-applied` from [here](https://github.com/clj-kondo/clj-kondo/issues/1156#issuecomment-823421125) as I found no way to know if the keyword is a namespaced keyword, any tips?

borkdude:
@ericdallo The code that adds the namespace for namespace-applied keywords is here:

https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/analyzer/usages.clj#L36

ericdallo:
You are right, I changed that node and didn't notice the `prefix`, done!

borkdude:
@ericdallo In case of an `_` I think it should not say `:namespace-applied true`

ericdallo:
Yeah, it make sense, what about `#:a{:b/c 1}` as the keyword will be `{:b/c 1}`?

borkdude:
Yeah, also not I think.

borkdude:
Maybe `:namespace-from-prefix true` is clearer

ericdallo:
Done!

test/clj_kondo/analysis_test.clj
```diff
@@ -126,14 +126,34 @@
     (testing "no namespace for key :a"
       (let [a (analyze "#:xml{:_/a 1}"
                        {:config {:output {:analysis {:keywords true}}}})]
-        (is (= '{:row 1, :col 7, :end-row 1, :end-col 11, :name "a", :filename "<stdin>"}
-               (first (:keywords a))))))
+        (assert-submaps
```
borkdude, line 129: Can you please revert this assert-submaps change? I consciously did not use it here since the namespace should be absent and not be an underscore.

ericdallo: Hum, got it, will change

borkdude: Maybe also add a comment about it why we didn't use assert-submaps there.

test/clj_kondo/analysis_test.clj
```diff
@@ -126,14 +126,34 @@
...
+          '[{:row 1, :col 7, :end-row 1, :end-col 11, :name "a", :filename "<stdin>"}]
+          (:keywords a))))
     (testing "no namespace for key :b"
       (let [a (analyze "#:xml{:a {:b 1}}"
                        {:config {:output {:analysis {:keywords true}}}})]
-        (is (= '[{:row 1, :col 7, :end-row 1, :end-col 9, :ns xml, :name "a", :filename "<stdin>"}
-                 {:row 1, :col 11, :end-row 1, :end-col 13, :name "b", :filename "<stdin>"}]
-               (:keywords a)))))))
+        (assert-submaps
```
borkdude, line 135: Same here.

test/clj_kondo/analysis_test.clj
```diff
@@ -123,17 +123,37 @@
...
+                 {:row 1, :col 11, :end-row 1, :end-col 13, :name "b", :filename "<stdin>" :auto-resolved false :namespace-from-prefix false}]
+          (:keywords a)))))
+    (testing "auto-resolved"
+      (let [a (analyze "(ns foo)
+                        :a ::b :bar/c
+                        #:d{:e 1 :_/f 2 :g/h 3 ::i 4}
+                        {:j/k 5 :l 6 ::m 7}"
+                       {:config {:output {:analysis {:keywords true}}}})]
+        (assert-submaps
+              '[{:name "a" :auto-resolved false :namespace-from-prefix false}
```
borkdude, line 146: I think it would be less noisy if we didn't add the keys if the value is `false`.

ericdallo: I thought about that, it works for me, but it will make test larger as well as assert-submap can't test keys absents

borkdude: Small price to pay compared to having larger payloads in the analysis output.

Issue #1259, opened 2021-04-20
Support returning non-zero exit codes on errors only


RickMoynihan:
### Problem 

It'd be nice to run `git rebase --interactive origin/master --exec "clj-kondo --lint ./src"` and have it fail on configured linters that issue an `:error`.

Unfortunately because of the way the shell interacts with git and the various pieces of the environment it's not easy to find a one-liner that can massage the return code and correctly abort the rebase at the failing commit.

### Suggested Solution 

Hence it would be nice to have a `--fail-on error` flag, which would only return a non-zero status code if there are any lint failures at `:error` level.  The proposed (explicitly stated) default for this would be `--fail-on warning`.

Then the above could be written like this:

`git rebase --interactive origin/master --exec "clj-kondo --lint ./src --fail-on error"`

Without having to create an intermediate script like this:

```shell
#!/bin/bash

clj-kondo --lint ./src

if [ $? -le 2 ]; then
        exit 0
else
        exit 1
fi
```

borkdude:
Rubocop:

bozhidar:cider: We have something like this, let me look it up.
It's fail-level - https://docs.rubocop.org/rubocop/usage/basic_usage.html#command-line-flags
4:16 PM
I think that's pretty common across linters.

Let's go for this then: `--fail-level error`.

PR #1261, opened 2021-04-21
Support new --fail-level command line flag

RickMoynihan:
Adds support for #1259


RickMoynihan:
I compiled myself a native image of this PR and have been using it this morning and it works really well.  With `git rebase -i origin/branch --exec "clj-kondo --lint ./src --fail-level error"`.

However I think it would also be worth adding a new issue to support `--display-only-fail-level-offenses` (from rubocop), as my repo currently reports a lot of clj-kondo warnings and eyeballing the errors that failed the lint is hard.  The obvious solution of `git rebase -i origin/branch --exec "clj-kondo --lint ./src --fail-level error | grep error"` doesn't work because it changes the failing error code back to `0`.

The workaround is to just rerun `clj-kondo` at the failed commit with `... | grep error` but `--display-only-fail-level-offenses` would I think be better, particular in CI etc, where you may end up having to redirect output to /dev/null on the first run, then re-run later on the error etc; or suffer lots more output to wade through.

README.md
```diff
@@ -188,10 +188,19 @@ directory, use `--cache false`.
...
 ## Exit codes
 
+Exit codes can be controlled by the `--fail-level <level>` option. The
+default fail level is `warning` which will return the following exit codes:
+
 - `0`: no errors or warnings were found
 - `2`: one or more warnings were found
 - `3`: one or more errors were found
 
+If `--fail-level error` is supplied a 0 exit will occur for any level:
```
borkdude, line 198: for any level?

PR #1266, opened 2021-04-25
Improve keyword reg

### LINKED ISSUE #1159: [Analysis] Support custom :def on keywords
ATM clojure-lsp check if the keywords analysis bucket has a `:def` and mark as a definition for find definition, we don't even need to check what is the value. it'd be nice to have this functionality for some macros that define vars with keywords, like `re-frame.core/reg-event-fx :foo/bar`.

clj-kondo do that automatically or provide a way to inform which macros register defs like that.

Slack discussion about that https://clojurians.slack.com/archives/CPABC1H61/p1612530939344500

ericdallo:
Fixes #1159 

- Rename `:def` -> `:reg`
- Add support for `re-frame.core` functions that register a keyword as definition.

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -13,6 +13,7 @@
    [clj-kondo.impl.analyzer.namespace :as namespace-analyzer
     :refer [analyze-ns-decl]]
    [clj-kondo.impl.analyzer.potemkin :as potemkin]
+   [clj-kondo.impl.analyzer.re-frame :as re-frame]
    [clj-kondo.impl.analyzer.spec :as spec]
    [clj-kondo.impl.analyzer.test :as test]
    [clj-kondo.impl.analyzer.usages :as usages :refer [analyze-usages2]]
@@ -1642,6 +1643,14 @@
                         (analyze-like-let ctx expr)
                         [babashka.process $]
                         (babashka/analyze-$ ctx expr)
+                        ([re-frame.core reg-event-db]
+                         [re-frame.core reg-event-fx]
+                         [re-frame.core reg-event-ctx]
+                         [re-frame.core reg-sub]
+                         [re-frame.core reg-sub-raw]
+                         [re-frame.core reg-fx]
+                         [re-frame.core reg-cofx])
+                        (re-frame/analyze-reg ctx expr (symbol (str resolved-as-namespace) (str resolved-as-name)))
                         ;; catch-all
                         (let [next-ctx (cond-> ctx
                                          (one-of [resolved-namespace resolved-name]
--- a/src/clj_kondo/impl/hooks.clj
+++ b/src/clj_kondo/impl/hooks.clj
@@ -1,7 +1,7 @@
 (ns clj-kondo.impl.hooks
   {:no-doc true}
   (:require [clj-kondo.impl.findings :as findings]
-            [clj-kondo.impl.utils :as utils :refer [vector-node list-node
+            [clj-kondo.impl.utils :as utils :refer [assoc-some vector-node list-node
                                                     sexpr token-node keyword-node
                                                     string-node #_map-node]]
             [clojure.java.io :as io]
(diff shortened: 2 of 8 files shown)


ericdallo:
@borkdude not sure we need to add the `:ref` as this seems enough to make clojure-lsp works, even so we could add in a next PR maybe?

src/clj_kondo/impl/hooks.clj
```diff
@@ -19,9 +19,9 @@
 
 (defn reg-keyword!
   ([k]
-   (reg-keyword! k true))
+   (reg-keyword! k nil))
```
ericdallo, line 22: it didn't seems right to register the keyword as a `:reg true` if user just `reg-keyword! :kw` without the intention of register that keyword as a definition, for those cases the 2 arity seems the correct approach.

ericdallo:
Tested with clojure-lsp and with this we have `find-definition` for re-frame keywords :)

src/clj_kondo/impl/hooks.clj
borkdude: Yes, so this arity can be removed? Would it be useful for LSP to also have `ref-keyword` so you can find if `re-frame.core/subscribe` or `re-frame.core/dispatch` (specifically one of those) uses the keyword? Or is this covered by `find-references` well enough?

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1642,6 +1643,14 @@
...
                         [babashka.process $]
                         (babashka/analyze-$ ctx expr)
+                        ([re-frame.core reg-event-db]
+                         [re-frame.core reg-event-fx]
+                         [re-frame.core reg-event-ctx]
+                         [re-frame.core reg-sub]
+                         [re-frame.core reg-sub-raw]
+                         [re-frame.core reg-fx]
+                         [re-frame.core reg-cofx])
+                        (re-frame/analyze-reg ctx expr (symbol (str resolved-as-namespace) (str resolved-as-name)))
```
borkdude, line 1653: Would it be useful if people can use `{:lint-as {foo.bar/baz re-frame.core/reg-event-fx}}`? In that case I think it's better to pass `resolved-namespace` instead of `resolved-as-namespace` (and also the `-name`)?

src/clj_kondo/impl/hooks.clj
ericdallo: Yeah, it's not necessary, the find references seems enough

borkdude: ok, let's remove it then

src/clj_kondo/impl/analyzer.clj
ericdallo: @borkdude you mean change the arity to something like:
```suggestion
                        (re-frame/analyze-reg ctx expr resolved-namespace resolved-name)
```

ericdallo: Oh, got it, done!

borkdude: I meant: use `resolved-namespace` instead of `resolved-as-namespace`. But this also seems fine.

src/clj_kondo/impl/hooks.clj
ericdallo: @borkdude not sure we should remove the 1 arity, imagine user wants to reg a keyword but not necessarily it's a keyword that is a definition, so just `(reg-keyword :my-key)` looks valid, otherwise user would always register the keyword with the `:def`

borkdude: Why would a user ever want to "register" a keyword without any context? And didn't you rename that key to `:reg`?

ericdallo: ops, yeah, `:reg` 😅.
You are right, if user don't register it, clj-kondo will reg as a normal keyword anyway.
I'll remove it so

borkdude:
@ericdallo Looks awesome! Could you add one more test for the `:lint-as` case? In case people want to use this mechanism for re-frame-like macros/fns

ericdallo:
Done @borkdude!

PR #1267, opened 2021-04-27
[analysis] Add output-schema-str to analysis output for schema.core

ericdallo:
Fixes https://github.com/clj-kondo/clj-kondo/issues/1268

Related to https://github.com/clojure-lsp/clojure-lsp/issues/181

- Add `output-schema-str` to `var-definitions` for `schema.core/def` and `schema.core/defn` (the only two possible output schemas var definitions).

I thought about `output-schemas-str` but it seems an unnecessary optimization for now as `schema.core` only allows one output schema per var.


borkdude:
This seems a bit of an ad hoc solution which only applies to the schema library. I would like to see an issue and some discussion first before we go ahead with this.

ericdallo:
@borkdude do you see any other better solution for this?

borkdude:
My last comment still stands.

borkdude:
Closing this due to staleness.

ericdallo:
I think this is still a good idea, but I have other priorities and didn't have the time to get this one, maybe leave it open still?

ericdallo:
oh, the issue is relevant, the PR can be closed indeed

PR #1290, opened 2021-06-02
Add :exclude config for the :use linter

IGJoshua:
This PR adds a new configuration option to the `:use` linter to permit excluding certain nses from the lint.

As it stands, on my machine there are two tests that fail, but they fail on my machine even without my changes, so this PR adds no failing tests.

The implementation I have here is very simplistic, just a one-liner, and I'd like some input on if this matches about what you'd like.

Before I mark this PR ready for merge I'll want to finish all the following tasks.
- [x] Add documentation in linters.md
- [ ] Update the changelog
- [ ]

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -959,3 +959,7 @@ it. That can be done as follows:
 
 *Example message:* `use :require with alias or :refer`.
 
+*Config:*
+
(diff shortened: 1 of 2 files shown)


IGJoshua:
Turns out that the `:use` linter follows the config for the `:refer-all` linter. This PR has been updated to only update the documentation to reflect this.

PR #1302, opened 2021-06-20
Setup Eastwood as a linter

vemv:
Updates Eastwood, proposes setting it up as a linter, and fixes the related issues in code.

---

Having Eastwood run in more projects would help me in its continued development. Would also go full circle relative to https://github.com/jonase/eastwood/commit/650f1d43c153f313b7f2998e10ab8f81b4fdbed3 !

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -12,7 +12,7 @@
            :extract
            {:extra-paths ["extract" "classes"]
             :extra-deps {clj-commons/conch {:mvn/version "0.9.2"}
-                         jonase/eastwood {:mvn/version "0.5.0"}}}
+                         jonase/eastwood {:mvn/version "0.5.2"}}}
            :test
            {:extra-deps
             {org.clojure/clojure {:mvn/version "1.9.0"}
@@ -21,7 +21,7 @@
              {:git/url "https://github.com/cognitect-labs/test-runner"
               :sha "cb96e80f6f3d3b307c59cbeb49bb0dcb3a2a780b"}
              clj-commons/conch {:mvn/version "0.9.2"}
-             jonase/eastwood {:mvn/version "0.5.0"}
+             jonase/eastwood {:mvn/version "0.5.2"}
              borkdude/missing.test.assertions {:mvn/version "0.0.2"}
              babashka/fs {:mvn/version "0.0.3"}
              babashka/process {:mvn/version "0.0.2"}
@@ -44,4 +44,7 @@
             :extra-deps
             {babashka/pods {#_#_:local/root "/Users/borkdude/Dropbox/dev/clojure/babashka/babashka.pods"
                             :git/url "https://github.com/babashka/babashka.pods"
-                            :sha "57cc95a2e2193a4bfb8505d44acaca900279369a"}}}}}
+                            :sha "57cc95a2e2193a4bfb8505d44acaca900279369a"}}}
+           :eastwood
+           {:main-opts ["-m" "eastwood.lint" {:source-paths ["src" "test"]
+                                              :exclude-linters [:unused-ret-vals]}]}}}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,9 +25,9 @@ jobs:
       - run:
           name: Install Clojure
           command: |
-            wget -nc https://download.clojure.org/install/linux-install-1.10.2.796.sh
-            chmod +x linux-install-1.10.2.796.sh
-            sudo ./linux-install-1.10.2.796.sh
+            wget -nc https://download.clojure.org/install/linux-install-1.10.3.855.sh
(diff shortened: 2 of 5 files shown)


src/clj_kondo/impl/cache.clj
```diff
@@ -136,9 +136,9 @@
                 idacs
                 [:clj :cljs :cljc])]
     (reduce (fn [idacs lang]
-              (update-in idacs [lang :defs]
-                         (fn [defs]
-                           (update-defs idacs cache-dir lang defs)))
+              #_(update-in idacs [lang :defs]
```
vemv, line 139: Unused, was hesitant about deleting it

deps.edn
```diff
@@ -44,4 +44,7 @@
             :extra-deps
             {babashka/pods {#_#_:local/root "/Users/borkdude/Dropbox/dev/clojure/babashka/babashka.pods"
                             :git/url "https://github.com/babashka/babashka.pods"
-                            :sha "57cc95a2e2193a4bfb8505d44acaca900279369a"}}}}}
+                            :sha "57cc95a2e2193a4bfb8505d44acaca900279369a"}}}
+           :eastwood
+           {:main-opts ["-m" "eastwood.lint" {:source-paths ["src" "test"]
+                                              :exclude-linters [:unused-ret-vals]}]}}}
```
vemv, line 50: `:unused-ret-vals` almost could be enabled, but there's a false positive.

It would be nice to be able to mark arbitrary defns in Eastwood as side-effectful to avoid this. As you know that is done in var-info.edn, but that is an entirely closed set.

borkdude:
@vemv Something is wrong (by default) on CircleCI with git deps. I pushed a "fix" for this.

As for adopting eastwood in CI: I think it's useful to use eastwood to detect "bad" things in clj-kondo's code, but it depends on how much extra work it would be to maintain a configuration for eastwood (sorry, I'm not familiar with using it at all) and how much extra time this would take in CI (I guess it would be ok). Maybe not fail the build if it detects stuff that's wrong, but just let it print warnings. Let's create issues out of the things it found as well so clj-kondo could maybe also try to support it eventually.

borkdude:
@vemv Btw, in eastwood you can update clj-kondo to the newest, the false positive should have been fixed.

vemv:
> Something is wrong (by default) on CircleCI with git deps. I pushed a "fix" for this.

Rebased, thanks! Hmm, now it fails for different reasons. I never got these failures locally.

> Maybe not fail the build if it detects stuff that's wrong, but just let it print warnings.

As you wish, but perhaps the cost would be low. Eastwood is generally stable (and improving). Runtime is ~1m in this project, perhaps not too much of a nuisance?

Worst-case scenario, e.g. there's an opaque issue, I'd take no offense whatsoever for a temporary disabling :)

> Let's create issues out of the things it found as well so clj-kondo could maybe also try to support it eventually.

Likewise as you wish. Personally I have it pretty clear that both linters have different approaches so they can complement each other. Otherwise one can end up pursuing arduous goals (like achieving snappy perf in Eastwood, or implementing a full JVM / clojure compiler in kondo for doing `eval`-powered analysis)

> Btw, in eastwood you can update clj-kondo to the newest, the false positive should have been fixed.

Yes fixed this morning in its `dev` branch thanks!

src/clj_kondo/impl/cache.clj
borkdude: I would be pretty surprised if this code is really unused.

vemv: Indeed, the lack of `!` (while there are other `!`s in code) confused me

Amending

borkdude:
@vemv Btw, are you also using clj-kondo in CI for eastwood?

vemv:
Yes! https://github.com/jonase/eastwood/blob/2e58aba73c9e39fa98ab46a2ead50d9ec31961e7/.circleci/lint.sh#L7

borkdude:
@vemv Cool. I just noticed you added eastwood to the JVM tests only, I think that makes a lot of sense. Let's do it.

borkdude:
@vemv You could try updating the Clojure CLI in the CircleCI build, I think it's a bit behind and this may fix the issue currently happening...

vemv:
> Checking out: https://github.com/borkdude/clj-kondo at 061b0ab11c86311d697302e9d0eede05669878e2

(If it persists, the issue might be that it's trying to fetch the wrong remote)

borkdude:
Seems all is good now?

vemv:
Yes :)

vemv:
Thanks much for the willingness. I created https://github.com/jonase/eastwood/issues/398 out of what I learned today

borkdude:
Thank you too!

PR #1343, opened 2021-08-01
[analysis] Fix `declare` name positions

### LINKED ISSUE #1342: [analysis] wrong `:name-col` and `:name-end-col` for `declare` var-definitions
**version**

2021-07-28

**platform**

JVM

**editor**

tested with clojure-lsp and clj-kondo from CLI

**problem**

clj-kondo return the analysis of var-definition for `declare` with wrong `:name-col`, it returns the `:name-col` as the same of `:col` making impossible to know where the symbol name starts.
This cause multiple issues on clojure-lsp like document highlighting and rename not renaming correctly: https://github.com/clojure-lsp/clojure-lsp/issues/497

**repro**

```bash
echo '(declare foo)' | clj -Sdeps '{:deps {clj-kondo/clj-kondo {:mvn/version "2021.07.28"}}}' -m clj-kondo.main  --lint "-" --config '{:cache true :copy-configs true :output {:analysis {:keywords true} :format :json :canonical-paths true}}' |jq .analysis
```

```json
...
{
      "end-row": 1,
      "name-end-col": 14,
      "name-end-row": 1,
      "name-row": 1,
      "ns": "user",
      "name": "foo",
      "defined-by": "clojure.core/declare",
      "filename": "<stdin>",
      "col": 1,
      "name-col": 1,
      "end-col": 14,
      "row": 1
    }
...
```

**expected behavior**

Follow other analysis returning the correct `:name-col` and `:name-end-col` of the other symbols.

```json
...
{
      "end-row": 1,
      "name-end-col": 13,
      "name-end-row": 1,
      "name-row": 1,
      "ns": "user",
      "name": "foo",
      "defined-by": "clojure.core/declare",
      "filename": "<stdin>",
      "col": 1,
      "name-col": 10,
      "end-col": 14,
      "row": 1
    }
...
```

ericdallo:
Fixes #1342

Tested with clojure-lsp as well

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -979,16 +979,17 @@
                             (current-namespace-var-name ctx var-name-node var-sym)))
                         var-name-nodes)]
     (doseq [var-name var-names]
-      (namespace/reg-var! ctx ns-name
-                          var-name
-                          expr
-                          (assoc (meta expr)
-                                 :name-row (:row (meta var-name))
-                                 :name-col (:col (meta var-name))
-                                 :name-end-row (:end-row (meta var-name))
-                                 :name-end-col (:end-col (meta var-name))
-                                 :declared true
-                                 :defined-by defined-by)))))
+      (let [var-name-meta (meta var-name)]
+        (namespace/reg-var! ctx ns-name
+                            var-name
+                            expr
+                            (assoc (meta expr)
+                                   :name-row (:row var-name-meta)
+                                   :name-col (:col var-name-meta)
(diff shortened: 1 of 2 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -982,6 +983,10 @@
                           var-name
                           expr
                           (assoc (meta expr)
+                                 :name-row (:row (meta var-name))
```
borkdude, line 986: It looks good, but I think you could only call `meta` once and store that in a local.

ericdallo: Done!

PR #1349, opened 2021-08-03
Expose ns-analysis fn in hooks api

hugoduncan:
Allow hooks to obtain ns analysis, when cached.

This is useful, for example, to rewrite macros that make implicit use of some symbols from other namespaces.

The implementation provides `hooks-api/ns-analysis`, which takes a  namespace symbol and returns any cached analysis for that namespace.  The returned map is keyed by language keyword (`:cl` and `:cljs`) and each value is a map keyed by a defined symbol.  The value for each  symbol is a subset of the information provided in the [analysis var-definitions](https://github.com/clj-kondo/clj-kondo/blob/master/analysis/README.md#data)

The function can also take an option map, with `:lang` specifying the source file language (`:clj`, `:cljs`, or `:cljc`)for which analysis should be returned. The default is to return analysis for all available source files for the given namespace.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -24,6 +24,7 @@
            me.raynes.conch/let-programs clojure.core/let
            datalog.parser.type/deftrecord clojure.core/defrecord
            clj-kondo.impl.rewrite-clj.potemkin/defprotocol+ clojure.core/defprotocol
-           clj-kondo.impl.rewrite-clj.potemkin/import-vars potemkin/import-vars}
+           clj-kondo.impl.rewrite-clj.potemkin/import-vars potemkin/import-vars
+           clj-kondo.test-utils/with-temp-dir clojure.core/let}
  :output {:exclude-files ["src/clj_kondo/impl/rewrite_clj_patch.clj"]}
  :hooks {:analyze-call {clj-kondo.impl.utils/one-of macroexpand.one-of/one-of}}}
--- a/src/clj_kondo/impl/hooks.clj
+++ b/src/clj_kondo/impl/hooks.clj
@@ -76,6 +76,9 @@
          (map-vals #(select-keys % selected-keys)))))
 
 (defn- ns-analysis*
+  "Adapt from-cache-1 to provide a uniform return format.
+  Unifies the format of cached information provided for each source
+  language."
   [lang ns-sym]
   (if (= lang :cljc)
     (->> (dissoc
@@ -89,8 +92,9 @@
 
 (defn ns-analysis
   "Return any cached analysis for the namespace identified by ns-sym.
-  Returns a map keyed by language keyword with values being var definitions
-  as output by the top level :analysis option."
+  Returns a map keyed by language keyword with values being maps of var
+  definitions keyed by defined symbol. The value for each symbol is a
+  subset of the values provide by the top level :analysis option."
   ([ns-sym] (ns-analysis ns-sym {}))
   ([ns-sym {:keys [lang]}]
    (if lang
--- a/src/clj_kondo/impl/hooks.clj
+++ b/src/clj_kondo/impl/hooks.clj
@@ -73,8 +73,7 @@
                        :fixed-arities :varargs-min-arity
                        :private :macro]]
     (->> (dissoc analysis :filename :source)
-         vals
-         (mapv #(select-keys % selected-keys)))))
+         (map-vals #(select-keys % selected-keys)))))
 
 (defn- ns-analysis*
   [lang ns-sym]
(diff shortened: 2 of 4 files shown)


borkdude:
@hugoduncan I think this is a good start. I don't think a multi-method is needed as the set of possibilities is fairly closed, but this is a detail. The most important issue at this point is: I don't want to expose the cache contents to the user directly, we'll have to go through an indirection to "curate" the API, so I can make breaking changes to the cache while keeping this API function stable.
I may bikeshed over the name of the function, but let's do that at the end.

borkdude:
Addendum. Preferably I'd only expose data that is available in analysis export right now: https://github.com/clj-kondo/clj-kondo/blob/master/analysis/README.md.

borkdude:
@hugoduncan To get started, perhaps just return a map with var -> info where info as map which contains the minimal information you need for your feature. And then we can extend that on a case by case basis.

hugoduncan:
I've pushed a change (before seeing your last comment), which returns var-definitions as a sequence as per the analysis readme.  I tried to get an exact match, but the cache doesn't have some of the keys that are currently being selected for (e.g. end-col).

borkdude:
>  but the cache doesn't have some of the keys that are currently being selected for

Yeah, this is because I don't cache more than needed, since writing/reading less is faster. Let's start very conservatively.

borkdude:
@hugoduncan Starting to look good. 

1. We need some tests for this. 
2. Have you verified that this covers your own use case?

hugoduncan:
I can confirm that this is sufficient to write the hook we need.  Let me know about the name, and I'll have a go add updating the changelog and docs.

borkdude:
@hugoduncan The thing I'm not sure about yet is the format. I think the data is good, but the shape of the date. We are transforming map vals into a list. We could just expose these as sym -> var info, so then it might be easier to write `(contains? ... 'bar)` if you are checking for the existence of a var. Also this allows extending the map with more info like `:meta ...` if we want to include metadata from the ns or so. What do you think?

hugoduncan:
That certainly makes sense, and easy to do.

borkdude:
OK, I think this is pretty much ready now, or perhaps we should think about the `{:clj ...}`, `{:cljs ...}` stuff more, but I think it's on a workable state for at least your use case and potentially more.

hugoduncan:
What are the concerns around language handling?  I don't have any experience of using knodo other than on plain clojure.

borkdude:
@hugoduncan Just thinking if the return value shape is nice and easy to work with. What we could do is include the `:filename` and `:lang` (either `:clj` or `:cljs` and `:cljc true | false` and not have this `{:clj ... :cljs ...}` nesting. In the most typical case people (like you) would probably use this for Clojure as in CLJS these kinds ns importing macros are quite painful work with. So then we would have one map with all the info necessary, perhaps.

hugoduncan:
So if iiuc, `:filename`, `:lang` and `:cljc` would be per var.  Can that work if the returned value is a map keyed on symbol -- could there not be differences between the clj and cljs versions of a var from a cljc file?

borkdude:
@hugoduncan You are correct, thanks for thinking along. I think this a reason to let it stay the way it's now.

borkdude:
@hugoduncan The cache is structured like `{:clj ... :cljs ... {:cljc {:clj ... :cljs ...}}}` for this reason. We're able to distinguish between vars from .clj, cljs or cljc files. We might want to do the same here? What if a user has defined a similar named var in a .clj file and in a .cljc file? You might even want to build a linter for this that warns about it, or so.

hugoduncan:
I was thinking that it would not be valid to have both a cljc file and either a clj or cljs file.  I take your point about providing a linter for this, but maybe that is an error outside of the scope of linting.  I could be persuaded either way.

hugoduncan:
fwiw, the var-definitions in the analysis data presumably returns a sequence rather than a map to allow for multiple definitions.

borkdude:
The shape is the analysis is mostly driven by "write fast" so just append things instead of indexing them for a certain use case. The shape of the cache is mostly driven by "read fast" so in a shape in which clj-kondo can get the information it typically needs to perform linting.

borkdude:
OK, based on the current format, I think we could in the future add `:cljc true` if a var came from a .cljc var and is both `:clj` and `:cljs`.  So I think this should be sufficient then...?

borkdude:
@hugoduncan Ready for merge? I'm double checking because of 

> PROOF OF CONCEPT - DO NOT MERGE

in the original message.

hugoduncan:
I'v updated the PR description to match the final implementation.

Issue #426, opened 2019-08-21
warning for a go-loop without a recur?


sogaiu:
**Is your feature request related to a problem? Please describe.**
When using a go-loop, it seems like most often one would like to have at least one recur within the go-loop's body.  Recently, I forgot to do this a few times and in each case it was a bug :)

**Describe the solution you'd like**
clj-kondo giving a warning for each go-loop that has no recurs.

**Describe alternatives you've considered**
Manually checking my code.

borkdude:
Probably we can generalize this issue to `loop`

drewverlee:
The implementation will need to match loops with the next recurs and highlight any that don't have a match. Consider this knarly code:

```
  (loop [x 1]
    (let [f (fn [] (loop (recur)))]
      (if true
        (loop (recur))
        (recur x))))
```

At a high level we should be able to just use a stack where we just pop whenever we see a recur. 

[]
[loop]
[loop loop]
see recur so pop
[loop]
[loop loop]
see recur so pop
[loop]
see recur so pop
[]

The analysis will need to happen after the body of the loop as discussed. @borkdude how are post walks handled? that is, how would i know when the body of a loop has been walked so i can look at the ctx and get out the information?

borkdude:
@drewverlee I'll repeat what I suggested on slack:

> For that specific linter, what I would do is look at analyzer.clj if we already have a special case for loop and recur. Then when going into the analysis for loop, add a key with a volatile or so to the ctx which tells us that we are inside of the loop. Then when encountering recur, you put a boolean true in the volatile (given that the volatile is there). Then after analyzing the body of loop, you will know that there was a recur. If not, emit a warning. 

In this way you can avoid walking the AST twice, which is more efficient.

To put in other words:

In the analysis of loop, you add a mutable object (atom or volatile) to the context. Upon encountering recur, you put a value in the mutable object. Then at the end (after) the analysis of loop, you look into the mutable object. If nothing has been put in the object, then there was no recur. 

If you also inject a mutable object in other forms where recur is expected, this works like the push/pop mechanism.

drewverlee:
Thanks @borkdude 

I understand that strategy and was able to create a solution with it that works in some cases. The cases I'm not sure how to handle are when in a given tree, a root node is a loop looking for a child recur, but gets a false positive because it contains a subtree with a lambda fn that has a recur target. E.g

```clojure
(loop  [x 1] <-- in loop 
[...]

PR #1356, opened 2021-08-16
Issue 426 warning for a loop without a recur

drewverlee:
This request should solve issue #426 , though there might be some minor clean-up (I store more in the atom than necessary, names, code organization) that should be considered before accepting it. But it's probably worth getting some feedback before I try doing any of that. 

The solution must include a stack or at least a pseudo stack so that nested loops are handled correctly. 

My primary worry is the logic is very disjoint, but given the nature of the problem that might be unavoidable. 

Here is the output from running the diff script. I'm not entirely sure what it's telling me:

```
➜  clj-kondo git:(issue-426-warning-for-a-loop-without-a-recur) ✗ ./script/diff
Cloning into 'clj-kondo'...
remote: Enumerating objects: 12375, done.
remote: Counting objects: 100% (1399/1399), done.
remote: Compressing objects: 100% (598/598), done.
remote: Total 12375 (delta 764), reused 1173 (delta 696), pack-reused 10976
Receiving objects: 100% (12375/12375), 11.82 MiB | 44.18 MiB/s, done.
Resolving deltas: 100% (7145/7145), done.
DEPRECATED: Libs must be qualified, change refactor-nrepl => refactor-nrepl/refactor-nrepl (/home/drewverlee/.clojure/deps.edn)
1923c1923
---
2173c2173
---
2558c2558
< linting took 8622ms, errors: 264, warnings: 2293
---
> linting took 8732ms, errors: 264, warnings: 2293
```

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -794,11 +794,23 @@
     (assoc-in ns [:qualify-ns alias-sym] ns-sym)))
 
 (defn analyze-loop [ctx expr]
-  (let [bv (-> expr :children second)]
+  (let [bv       (-> expr :children second)
+        filename (:filename ctx)]
     (when (and bv (= :vector (tag bv)))
-      (let [arg-count (let [c (count (:children bv))]
-                        (when (even? c)
-                          (/ c 2)))]
+      (let [arg-count   (let [c (count (:children bv))]
+                          (when (even? c)
+                            (/ c 2)))
+            recur-count (count (macroexpand/find-children  #(= 'recur (:value %)) (:children expr)))]
+
+        (when (zero? recur-count)
+          (findings/reg-finding!
+            ctx
+            (node->line
+              filename
+              expr
+              :loop-missing-recur
+              (format "loop expects at least 1 recur call in the body but got none"))))
+
         (analyze-like-let (assoc ctx
                                  :recur-arity {:fixed-arity arg-count}) expr)))))
 
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -795,28 +795,40 @@
 
 (defn analyze-loop [ctx expr]
   (let [bv       (-> expr :children second)
-        filename (:filename ctx)]
+        filename (:filename ctx)
+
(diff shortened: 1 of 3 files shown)


borkdude:
> The solution must include a stack or at least a pseudo stack so that nested loops are handled correctly.

That mechanism works automatically when you just insert an atom into the context and then analyze. The next time you encounter a fn or loop another atom is inserted so by the time you will encounter a recur, it automatically corresponds to the closest fn/loop: no need to build more stack-like logic.

drewverlee:
@borkdude 

> it automatically corresponds to the closest fn/loop: 

yes, but if it inserts a new atom on each loop then it forgets the past/parents. What if they still need a matching recur? Maybe there is some control flow I'm not following,  but what your suggesting was my first intuition as well which is what this commit looks like: 

https://github.com/clj-kondo/clj-kondo/pull/1356/commits/cbe7e1561717f4526d4bfcca037ce8ee7da970aa. 
(though you need the fix from the next commit to run it)

It fails on this case which I hope highlights the issue, ill try to explain what I think you suggesting and why it didn't work for me.  

```clojure
(def foo ;; pre-walk: insert atom
    [s]
    (loop [a 1] ;; pre-walk: insert atom. post-walk:  atom should be in what state?
      (if true
        (recur 1) ;; set atom to true
        (loop [a 1] ;; Pre-walk: insert atom (overwrite our true value). post walk: atom is true, which is good, BUT what do we set the atom to now, we have lost track of if we have a loop and recur above?
          (if false
            (recur (a)) ;; set atom to true
            a)))))
```

But given your confidence that it isn't necessary, I must be missing something. Apologies for the long discussion.

borkdude:
@drewverlee About the length of the discussion: no problem at all. I will try it out locally and try to see if I can detect any problems.

borkdude:
@drewverlee I took a look at your code. The thing I would do differently is this: do not re-use the atom previously inserted when you encounter a new loop or fn. Always insert a new one. So when recur is encountered that matches that loop or fn, it will pick up on that one. Unless there was an inner fn or loop, then the recur would pick up on the inner fn / loop atom and as a result the outer loop / fn atom would not have been populated. That is how you detect that there was a missing recur for the outer loop / fn. The only thing you need is an atom with a boolean in it.

PR #1387, opened 2021-09-20
[analysis] Add scope-end-row and scope-end-col to var-usages

ericdallo:
- Add `:end-row` and `:end-col` for all var-usages when using as a function call
- rollback https://github.com/clj-kondo/clj-kondo/commit/4f74945fb050210e82f76d9a9e5293a633a23b22 as we already have `:name-*` for var-usage name location

Related to https://github.com/clojure-lsp/clojure-lsp/issues/569

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1816,8 +1816,6 @@
                                     :end-row (:end-row expr-meta)
                                     :col col
                                     :end-col (:end-col expr-meta)
-                                    :scope-end-row (:end-row expr-meta)
-                                    :scope-end-col (:end-col expr-meta)
                                     :base-lang base-lang
                                     :lang lang
                                     :filename (:filename ctx)
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1523,7 +1523,6 @@
                                                    :config config
                                                    :ns ns-name})
                                          (catch Exception e
-                                           ;;TODOAQUI
                                            (findings/reg-finding!
                                             ctx
                                             (merge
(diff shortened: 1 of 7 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -1817,6 +1816,8 @@
                                     :end-row (:end-row expr-meta)
                                     :col col
                                     :end-col (:end-col expr-meta)
+                                    :scope-end-row (:end-row expr-meta)
```
borkdude, line 1819: @ericdallo Why are you adding twice the same data, once as `:end-row` and once as `:scope-end-row` (same for col)?

ericdallo: because only this place we want to add the scope, If I use the same end-row [here](https://github.com/clj-kondo/clj-kondo/pull/1387/files#diff-19afea30102cec0da2c4d0516d5e834219f6ccaadc2019a9b86e5af604050425R232) other usages will have scope and [this test](https://github.com/clj-kondo/clj-kondo/pull/1387/files#diff-f73159ab0c7e8a8b4d7d99a2b007cdc86046bb5dd44951e0f129bc43c1505a66R370-R383) will fail

borkdude:
@ericdallo Perhaps add a breaking change note to CHANGELOGS.md as well?

borkdude:
For the rest, this seems good to me.

borkdude:
@ericdallo In LSP, do you need to distinguish between function call usages and other usages or can you just always use the end-* positions now?

ericdallo:
@borkdude sometimes I need to know if I want the name location or the usage location, so there are places that use one and others that now need only the end-*

borkdude:
@ericdallo I keep a separate list of breaking changes at the end of the changelogs as well. Can you also add it there?

ericdallo:
Done @borkdude

PR #1389, opened 2021-09-23
Expose config functions

ericdallo:
Discussed [here](https://github.com/clojure-lsp/clojure-lsp/issues/331#issuecomment-925785508)


borkdude:
Can you clarify in the docstring what the second optional config argument does?

ericdallo:
Done @borkdude

PR #1418, opened 2021-10-14
Allow user to request all or specific metadata be returned with analysis data for vars and nses

### LINKED ISSUE #1280: Provide all metadata from vars analysis
Hi @borkdude. Once again thanks for your awesome tool!

I'd like to be able to mark vars with metadata and the filter vars with that metadata using analysis' `:var-definitions` in a bb script. With this feature, I could write bb scripts that generate configuration for tools that have var configuration. For example, I could generate a carve ignore list by marking fns with `^:repl`

It'd be helpful for a var  in `:var-definitions` to include all the metadata it contains. This metadata could be at the top level or under a `:meta` key if that seems cleaner. It seems analysis [obtains a var's metadata](https://github.com/clj-kondo/clj-kondo/blob/fb3393cf3be30f6fb517040d2bd7dbb4c40e3925/src/clj_kondo/impl/namespace.clj#L99) and then only provides some of that metadata as analysis data.

I think `:var-definitions` is the most appropriate place to contains this analysis data. I don't know of other tools that could provide this data as effectively in a bb script

lread:
Notes to reviewer:
1. See updates to docs for usage description
1. This change considers metadata to be explicitly user-coded metadata. 
For example `(ns my.ns.here "docs")` won't return `:doc` under `:meta` but `(ns ^{:doc "docs"} my.ns.here)` will.
Reader generated row and col metadata are excluded.
This makes sense to me but thought it worth mentioning.
1. I did rework `reg-namespace!` a bit to have it more match `reg-var!`. 
This was not necessary to solve the problem, but makes the code more consistent.
Will reverse this if you don't love it.
1. Happy learn what you uncover from your review and also happy to make changes.

Closes #1280

### DIFF
--- a/src/clj_kondo/impl/metadata.clj
+++ b/src/clj_kondo/impl/metadata.clj
@@ -52,12 +52,14 @@
            meta-map (if analyze-meta?
                       (assoc meta-map :user-meta [meta-map])
                       meta-map)
+           ;; clear user-coded metadata that can conflict with clj-kondo
+           ;; clj-kondo only sometimes sets these but later always checks them
+           meta-map (if (seq meta-map)
+                      (dissoc meta-map :name-row :name-col :name-end-row :name-end-col)
+                      meta-map)
            node (-> node
                     (dissoc :meta)
-                    (with-meta (-> meta-map
-                                   (merge (meta node))
-                                   ;; clear out positional root keys we rely on, but only sometimes set
-                                   (dissoc :name-row :name-col :name-end-row :name-end-col))))]
+                    (with-meta (merge meta-map (meta node))))]
        node)
      node)))
 


borkdude:
@lread Thank you! Will take a look tomorrow as I'm almost going to sleep here :)

borkdude:
I think it would be better if `lift-meta-content2` pushed the metadata under a different key, so we get:

```
{:row 1, :col 30, :end-row 1, :end-col 31, :meta {:cool :yes, :row true}}
```

Have to refactor quite a but of stuff but I think it's worth it.

borkdude:
@lread This is quite the adjustment with 38 failing tests. Are you sure? :)

borkdude:
I'd be happy to let you work on this. We can also do this as a follow up issue perhaps.

lread:
Ok @borkdude, when you have some time, looking forward to your feedback!

borkdude:
``` clojure
user=> (-> (with-in-str "(def ^{:row true :cool :yes} x)" (clj-kondo/run! {:lint ["-"] :config {:output {:format :edn :analysis {:var-definitions {:meta true}}}}})) :analysis :var-definitions first )
{:end-row 1, :meta {:cool :yes, :row true}, :name-end-col 31, :name-end-row 1, :name-row true, :ns user, :name x, :defined-by clojure.core/def, :filename "<stdin>", :col 1, :name-col 30, :end-col 32, :row 1}
```

Check the value of `:name-row` This should be `1` I think. This bug was already there but we might want to write a test for this and fix it now maybe.

It has to do with the order of metadata merging, we should always pick the rewrite-clj data as the last one since we always want that row internally

PR #1421, opened 2021-10-16
Add :namespace to duplicate-require linter

ericdallo:
This will help implement https://github.com/clojure-lsp/clojure-lsp/issues/527 and not require parsing the `:message` to get the namespace


borkdude:
yeah or `:duplicate-ns`, one character shorter :)

PR #1431, opened 2021-10-19
Analysis updates: support for defn 2nd attr-map, :doc derivation fixes

lread:
Please see commits for details.

Notes
- I tried to follow the spirit of existing code.
- I also tried avoid introducing too much extra work.
- I opted for brute for testing of :doc derivation instead of generative testing.

I look forward to review feedback/tweaks.


borkdude:
Thank you

Issue #1211, opened 2021-03-10
Support :context map where users can add abitrary data to node which shows up in analysis


borkdude:
The `:def` naming in keyword analysis output is too restrictive for the following use case:

1) Lint all keywords used from re-frame's reg-event-fx
2) Lint all keywords used from re-frame's dispatch

The keywords will have `:context re-frame/reg-event-fx` or `:context re-frame/dispatch` so you can discover events that were never used, or events that were dispatched without being defined.

The linter can be implemented using hooks for the respective re-frame functions and the `api/reg-keyword!` function.

This will be a breaking change for clojure-lsp but I think it makes sense to rename it.

See https://github.com/clj-kondo/clj-kondo/blob/master/analysis/README.md

/cc @snoe @ericdallo @mike-thompson-day8

borkdude:
People already started using this key elsewhere: https://gist.github.com/roman01la/c6a2e4db8d74f89789292002794a7142

New proposal: add `:context` key that can be used in things like `dispatch` (which doesn't really define anything, but uses something).

Or maybe just call the extra key `:used-in`?

borkdude:
We're just sticking with `:def` and possibly we can add `:ref` for things _using_ the keyword rather than "defining" it. /cc @ericdallo .

PR #1446, opened 2021-11-03
re-frame events and subscriptions like vars in keyword analysis

benedekfazekas:
- var usages in re-frame body register reg as from-var
- keyword usages record from-ns, from-var, from-reg (if they occur in a dispatch or subscription)
- from-var and in-def records keywords as string instead of keyword
  and test to prove the order of keywords is indifferent to detect usages
- analyze re-frame subscriptions, dispatches
 only add `from-var`, `from-reg` if we are in a subscription or a dispatch
 analyze subscriptions
 -- direct `subscribe` calls
 --  signal fns in reg-subs
 -- `:<-` constructs in reg-subs
 analyze dispatches
 -- direct `dispatch` and `dispatch-sync` calls
 -- analyze `reg-event-fx` to add from-var, from-reg for only those
  keywords that show up in the vector/map of `:dispatch` kw and its variations

### DIFF
--- a/src/clj_kondo/impl/analysis.clj
+++ b/src/clj_kondo/impl/analysis.clj
@@ -29,7 +29,8 @@
                                :name-end-row
                                :name-end-col
                                :end-row
-                               :end-col]))
+                               :end-col
+                               :reg]))
                 :arity arity
                 :lang lang
                 :from-var in-def))))))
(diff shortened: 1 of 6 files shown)


borkdude:
@benedekfazekas 

I've tried to review this PR a couple of times, but the problem I keep running into is that there are multiple things added in one go and I don't have a clear grasp on what was added for which (sub)problem.

Is it possible to split this PR up into a few smaller ones?

1) Post an issue to explain what the problem is that you are going to solve. Keep the problem as small as possible.
2) Post a corresponding PR with the smallest change possible.
3) Then we discuss the PR and move on to the next one.

This way it's easier for me to reason about what is going on and why things are being added. I want to be a bit cautious with what is added to the analysis data, how things are called, have a good reason for why its there, etc, because once its in there, we cannot change it and it's hard to find back in the commit history why we did things.

benedekfazekas:
fair enough. pity I squashed the commits. happy to walk you through it or answer specific questions but ofc will try to break it up into smaller chunks.

borkdude:
@benedekfazekas I'm sure we'll get there but let's do it in smaller steps. Then it's also easier for me to see where #1211 will fit in, which is related to this.

borkdude:
@benedekfazekas Can you create the first minimal issue to kick things off?

benedekfazekas:
bit crazy week, been to my dayjob's office first since the pandemic started at the beginning of the week. so will have only time to kick this off on the weekend. so bear with me a bit pls

borkdude:
@benedekfazekas Of course! Enjoy your office! :)

benedekfazekas:
keeping this in sync for reference

benedekfazekas:
no real point keeping this open I suppose

PR #1462, opened 2021-11-17
[Fixes 1461] Add reg as from-var for vars in a re-frame macro body

benedekfazekas:

### DIFF
--- a/test/clj_kondo/analysis_test.clj
+++ b/test/clj_kondo/analysis_test.clj
@@ -114,6 +114,15 @@
       (assert-submaps
        '[{:name "kw" :reg user/mydef}]
        (:keywords a))))
+  (testing "var usage in re-frame.core/reg-event-db body makes reg available in :context under :in-reg key"
+    (let [a (analyze "(require '[re-frame.core :as rf])
+                      (rf/reg-event-db ::a (constantly {}))"
+                     {:config {:output {:analysis {:keywords true}}}})]
+      (assert-submaps
+       '[{:name require}
+         {:name constantly :context {:in-reg {:k :a :reg re-frame.core/reg-event-db}}}
+         {:name reg-event-db}]
+       (:var-usages a))))
   (testing "hooks can add :reg"
     (let [a (analyze "(user/mydef ::kw (inc))"
                      {:config {:output {:analysis {:keywords true}}
@@ -1197,4 +1206,4 @@
             inc-usage (some #(when (= 'inc (:name %)) %) usages)
             context (:context inc-usage)]
         ;; THIS IS JUST AN EXAMPLE, subject to change per feedback from Benedek.
-        (is (submap? context '{re-frame.core/reg-sub {:row 2, :col 56, :end-row 2, :end-col 61}}))))))
+        (is (submap? '{re-frame.core/reg-sub {:row 2, :col 56, :end-row 2, :end-col 61}} context))))))
--- a/test/clj_kondo/analysis_test.clj
+++ b/test/clj_kondo/analysis_test.clj
@@ -114,15 +114,6 @@
       (assert-submaps
        '[{:name "kw" :reg user/mydef}]
        (:keywords a))))
-  (testing "var usage in re-frame.core/reg-event-db body makes reg available in :context under :in-reg key"
-    (let [a (analyze "(require '[re-frame.core :as rf])
-                      (rf/reg-event-db ::a (constantly {}))"
-                     {:config {:output {:analysis {:keywords true}}}})]
-      (assert-submaps
-       '[{:name require}
-         {:name constantly :context {:in-reg {:k :a :reg re-frame.core/reg-event-db}}}
-         {:name reg-event-db}]
-       (:var-usages a))))
   (testing "hooks can add :reg"
     (let [a (analyze "(user/mydef ::kw (inc))"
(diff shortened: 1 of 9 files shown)


benedekfazekas:
this is essentially part of #1446 more PRs will follow if this is discussed/agreed on cc @ericdallo

benedekfazekas:
followed up the `context` branch in this. so far looks good and simple. will do the follow up on the other PR as well (`reg-sub` handling) then locally merge the two so I can test them against morpheus; eg. how this is played out when used by a tool. hope this makes sense

test/clj_kondo/analysis_test.clj
```diff
@@ -114,6 +114,15 @@
       (assert-submaps
        '[{:name "kw" :reg user/mydef}]
        (:keywords a))))
+  (testing "var usage in re-frame.core/reg-event-db body makes reg available in :context under :in-reg key"
+    (let [a (analyze "(require '[re-frame.core :as rf])
+                      (rf/reg-event-db ::a (constantly {}))"
+                     {:config {:output {:analysis {:keywords true}}}})]
+      (assert-submaps
+       '[{:name require}
+         {:name constantly :context {:in-reg {:k :a :reg re-frame.core/reg-event-db}}}
```
borkdude, line 123: - `:k`: can we use a more descriptive name here? e.g. `:keyword`?
- don't we need to save more information than `:a`? E.g. we cannot distinguish between auto-resolved keywords or simple keywords here?
- why don't we just use as the key into the context?

```
{:context {re-frame.core/reg-event-db {:keyword {:name "a" :ns ... :auto-resolved true}
```

benedekfazekas: - re `:k` sure, sounds better
- sure, maybe ns as well
- hm.. not sure it would be easy to work with a context where the key is `re-frame.core/reg-event-db`. In the client code I would rather work with some constant key name I can easily check if there or not. maybe I don't fully understand your intentions here?!

borkdude: @benedekfazekas Another idea:

- We could maybe refer to the keyword using an identifier. This identifier points to the keyword in the `:keywords` analysis. This way we avoid problems of pre-analyzing the keyword for aliases etc, to get the namespace which is done in a later stage.

- Adding to the context should preferably happen using a qualified keyword, so each kind of hook doesn't overwrite keys of other than its own.  I'm find with adding a generic key, but let's use something with `re-frame` in the name. E.g. `{:context {:re-frame/reg {:keyword {:id "123"}}}}`

benedekfazekas: i deffo like the qualified keyword idea

borkdude: @benedekfazekas OK, so what about the value for the key? Just adding `:keyword :a` there isn't sufficient, I think.

benedekfazekas: worked on the shape of the data, hopefully along he lines of this discussion ;) let me know what you think

ericdallo: +1 about adding an id to the keyword, it's similar to how `:locals` works today, the issue is that we'd need to add ids for all keywords but won't have issues about something that was not analyzed yet

borkdude:
@benedekfazekas @ericdallo 

I updated the `context` branch with the reference ids. Instead of repeating data we can refer to the keyword analysis under `:keywords`.


This code:

``` clojure
(require '[re-frame.core :as re-frame]) 
(re-frame/reg-sub ::foo (fn [_]))
```

now gives this keyword analysis:

``` clojure
[{:row 1,
  :col 59,
  :end-row 1,
  :end-col 64,
  :ns user,
  :auto-resolved true,
  :context #:re-frame.core{:reg-sub-id "21"},
  :name "foo",
  :filename "<stdin>"}]
```

and this in var usages:

``` clojure
 {:fixed-arities #{1},
  :end-row 1,
  :name-end-col 77,
  :name-end-row 1,
  :name-row 1,
  :name inc,
  :filename "<stdin>",
  :from user,
  :col 73,
  :name-col 74,
  :context #:re-frame.core{:in-reg-sub-id "22"},
  :end-col 80,
  :arity 1,
  :row 1,
  :to cljs.core}
```

I think this contains all the data necessary that went into this PR. If you can live with this then I'll merge this to master.

benedekfazekas:
this could work. my initial thoughts that having this 'dynamic' ids in the context reads really nicely but would be a bit more difficult to work with in tooling code. e.g. you need to do some kind of pattern matching to find all the diff. re-frame ids. maybe it would worth to have a submap for all the regs in the context so a tool can expect only reg related keys in that submap and don't need to filter out other possible context keys.. 

also the idea here that we go fully normalised right? so if i want to show any info about the keyword/reg referenced by an id in the usage i supposed to go back keywords, find the right one and access the data there. that bit should be fine i suppose.

borkdude:
Having a common key can work. 

For the keyword analysis:

- We could just use `:re-frame.core` as the common key into the context:

`:context {:re-frame.core {:id "6" :var reg-event-db}

For the var analysis context:

`:context {:re-frame.core {:in-id "6"}}`

borkdude:
And yes, the idea is to normalize the data.

benedekfazekas:
sounds good.

benedekfazekas:
suppose this run its course. closing

PR #1469, opened 2021-11-25
[#1211] Analysis context

borkdude:


borkdude:
@benedekfazekas @ericdallo Here is the updated PR. The common key works out nicely also with respect to selecting data from the analysis output. I'll merge when you both give the thumbs up.

Added to the docs:

### Context data

Built-in and custom hook can add arbitrary data to the analysis using a
`:context` map. This context map will currently appear in `:var-usages` and
`:keywords`. You can opt-in to the entire context map using `:context true` or
select certain keys using `:context [:re-frame.core]`.

benedekfazekas:
this is coming along really nicely. :) let me play with it a bit this evening just to make sure how it fits into the further steps for the re-frame story

benedekfazekas:
👍 from me

PR #1487, opened 2021-12-07
Add testing-str context

ericdallo:

https://github.com/clojure-lsp/clojure-lsp/issues/653

### DIFF
--- a/src/clj_kondo/impl/analyzer/test.clj
+++ b/src/clj_kondo/impl/analyzer/test.clj
@@ -23,6 +23,13 @@
            body))))
    defined-by))
 
+(defn analyze-testing [ctx expr resolved-as-namespace]
+  (let [[testing testing-str & rest-children] (:children expr)
+        kns (keyword resolved-as-namespace)]
+    (common/analyze-children (assoc-in ctx [:context kns :testing-str] (utils/sexpr testing-str))
+                             testing)
+    (common/analyze-children ctx rest-children)))
+
 (defn analyze-cljs-test-async [ctx expr]
   (let [[binding-expr & rest-children] (rest (:children expr))
         binding-name (:value binding-expr)
--- a/src/clj_kondo/impl/analyzer/test.clj
+++ b/src/clj_kondo/impl/analyzer/test.clj
@@ -24,11 +24,10 @@
    defined-by))
 
 (defn analyze-testing [ctx expr resolved-as-namespace]
-  (let [[testing testing-str & rest-children] (:children expr)
+  (let [[_ testing-str & _rest-children] (:children expr)
         kns (keyword resolved-as-namespace)]
     (common/analyze-children (assoc-in ctx [:context kns :testing-str] (utils/sexpr testing-str))
-                             testing)
-    (common/analyze-children ctx rest-children)))
+                             expr)))
 
 (defn analyze-cljs-test-async [ctx expr]
   (let [[binding-expr & rest-children] (rest (:children expr))
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1839,6 +1839,9 @@
                          [cljs.test deftest])
                         (test/analyze-deftest ctx expr defined-by
                                               resolved-as-namespace resolved-as-name)
+                        ([clojure.test testing]
+                         [cljs.test testing])
+                        (test/analyze-testing ctx expr resolved-as-namespace)
                         ([clojure.core.match match] [cljs.core.match match])
                         (match/analyze-match ctx expr)
                         [clojure.string replace]
--- a/test/clj_kondo/analysis_test.clj
+++ b/test/clj_kondo/analysis_test.clj
@@ -1089,7 +1089,6 @@
                                    {:analysis true}}}))
                :analysis :namespace-definitions first)))))
 
-
 (deftest derived-doc
   (testing "namespace"
     (let [enable? [false true]]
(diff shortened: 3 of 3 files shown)


ericdallo:
@borkdude the test is failing because I don't know why the :context is coming empty, I probably missed something from this new :context logic, any thoughts?

src/clj_kondo/impl/analyzer/test.clj
```diff
@@ -23,6 +23,13 @@
            body))))
    defined-by))
 
+(defn analyze-testing [ctx expr resolved-as-namespace]
+  (let [[testing testing-str & rest-children] (:children expr)
+        kns (keyword resolved-as-namespace)]
+    (common/analyze-children (assoc-in ctx [:context kns :testing-str] (utils/sexpr testing-str))
```
borkdude, line 29: I think you need to use the ctx with the context while analyzing the testing call.

ericdallo: @borkdude you mean this?
```suggestion
    (common/analyze-children (assoc-in ctx [:context kns :testing-str] (utils/sexpr testing))
```
I tried that as well but it doesn't work

borkdude: No, that's not what I meant. You have to pass the ctx with the while analyzing the whole testing expression, not on the child.

ericdallo: hum, so I should analyze the whole `expr` with that ctx, right?
I'll give a try

ericdallo: Hum, I changed it but I still get the same empty context, I don't know what I'm missing here

borkdude: You have to include `:context true` or `:context [:keyword]` to receive the context in the analysis.

borkdude: OK, you already did that.

borkdude: I'll take a look later this week.

borkdude: @ericdallo Figured it out already. You didn't lint the children of testing.

```
(defn analyze-testing [ctx expr resolved-as-namespace]
  (let [[_ testing-str :as children] (:children expr)
        kns (keyword resolved-as-namespace)]
    (common/analyze-children (assoc-in ctx [:context kns :testing-str] (utils/sexpr testing-str))
                             children)))
```

ericdallo: even with that, the context comes empty to me 🤔

borkdude:
```
$ clojure -M:clj-kondo/dev --lint - --config '{:output {:analysis {:context true} :format :edn}}' <<< '(require (quote (clojure.test))) (clojure.test/testing "foo" (inc 1))' | jet --pretty
```

```
                         {:fixed-arities #{1},
                          :end-row 1,
                          :name-end-col 66,
                          :name-end-row 1,
                          :name-row 1,
                          :name inc,
                          :filename "<stdin>",
                          :from user,
                          :col 62,
                          :name-col 63,
                          :context {:clojure.test {:testing-str "foo"}},
                          :end-col 69,
                          :arity 1,
                          :row 1,
                          :to clojure.core}
```

ericdallo: Yeah, you are right, so probably my test is wrong somewhere 🤔

ericdallo: Oh, there is something weird happening indeed:
```clojure
{{:end-row 1,
                          :name-end-col 55,
                          :name-end-row 1,
                          :name-row 1,
                          :name testing,
                          :filename "<stdin>",
                          :from user,
                          :macro true,
                          :col 35,
                          :name-col 35,
                          :end-col 55,
                          :varargs-min-arity 1,
                          :row 1,
                          :to clojure.test}
                         {:fixed-arities #{1},
                          :end-row 1,
                          :name-end-col 66,
                          :name-end-row 1,
                          :name-row 1,
                          :name inc,
                          :filename "<stdin>",
                          :from user,
                          :col 62,
                          :name-col 63,
                          :context {:clojure.test {:testing-str "foo"}},
                          :end-col 69,
                          :arity 1,
                          :row 1,
                          :to clojure.core}
                         {:end-row 1,
                          :name-end-col 55,
                          :name-end-row 1,
                          :name-row 1,
                          :name testing,
                          :filename "<stdin>",
                          :from user,
                          :macro true,
                          :col 34,
                          :name-col 35,
                          :end-col 70,
                          :arity 2,
                          :varargs-min-arity 1,
                          :row 1,
                          :to clojure.test}]}}

```

borkdude: The issue is that only children of testing get this context, but not the testing call itself. I need to re-organize some stuff in `analyze-call` to support that. It's already supported for custom hooks.

borkdude: Also: I think you might want to store the stack of testing-strs, not only one maybe?

ericdallo: I see.
I thought about leaving that on LSP side, since we will already need to build the tree there, not sure adding there would help that much

ericdallo: I'll probably know more about it when start to implement on clojure-lsp

borkdude: That's ok, adding less data is better for performance, we will see.

ericdallo:
@borkdude I managed to create the test hierarchy on lsp side, now it's just missing the `testing-str` (which I replaced with `"mock"`) 🚀 :
```clojure
{:uri "file:///home/greg/dev/clojure-lsp/src/clojure_lsp/bla.clj",
 :tests
 [{:name foo,
   :range
   {:start {:line 4, :character 0}, :end {:line 11, :character 18}},
   :name-range
   {:start {:line 4, :character 9}, :end {:line 4, :character 12}},
   :kind :deftest,
   :children
   [{:name "mock",
     :range
     {:start {:line 5, :character 2}, :end {:line 6, :character 14}},
     :name-range
     {:start {:line 5, :character 3}, :end {:line 5, :character 10}},
     :kind :testing}
    {:name "mock",
     :range
     {:start {:line 7, :character 2}, :end {:line 11, :character 17}},
     :name-range
     {:start {:line 7, :character 3}, :end {:line 7, :character 10}},
     :kind :testing,
     :children
     [{:name "mock",
       :range
       {:start {:line 8, :character 4}, :end {:line 9, :character 16}},
       :name-range
       {:start {:line 8, :character 5}, :end {:line 8, :character 12}},
       :kind :testing}
      {:name "mock",
       :range
       {:start {:line 10, :character 4},
        :end {:line 11, :character 16}},
       :name-range
       {:start {:line 10, :character 5},
        :end {:line 10, :character 12}},
       :kind :testing}]}]}]}
```

borkdude:
@ericdallo If I understand correctly, that isn't yet using anything from this PR right? Just the testing var calls with their locations? So the only thing we need to add to the call is the context + string to show it. I hope to take a look at this soon.

ericdallo:
Exactly @borkdude :)

borkdude:
Awesome!

ericdallo:
@borkdude I was wondering if we could evolve this feature to other test frameworks like `midje` and `state-flow`:
like having the same tree for:
midje
```clojure
(ns a (:require [midje.sweet :refer [facts fact]]))
(facts foo
  (fact "something"))
```
state-flow
```clojure
(ns a (:require [state-flow.core :refer [defflow flow]]))
(defflow foo
  (flow "something"))
```
WDYT?

Issue #805, opened 2020-03-15
Feature request: Docstring linter similar to Emacs's checkdoc


bsless:
Emacs comes with a linter which helps writing good docstrings, called `checkdoc`, can be found [here](https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/checkdoc.el).
From the file's commentary:
```
;;   The Emacs Lisp manual has a nice chapter on how to write
;; documentation strings.  Many stylistic suggestions are fairly
;; deterministic and easy to check for syntactically, but also easy
;; to forget.  The main checkdoc engine will perform the stylistic
;; checks needed to make sure these styles are remembered.
```
It would be nice to add a similar feature as an optional linter, and it should be easy to do something similar based on [Clojure's style guide documentation section](https://github.com/bbatsov/clojure-style-guide#documentation)

I wouldn't mind working on it myself

borkdude:
Could you give a few examples of checks for Clojure docstrings? Input + output?

bsless:
Sure.
For the simplistic case:

```clojure
(defn foo [a b] (+ a b))
```
For no docstring there's already a linter.
- "adds two numbers":
  - First sentence should end with punctuation
  - First line should be capitalized
  - Probably "adds" should be imperative "add"
  - Argument 'a' should appear (as \`a\`) in the doc string.

That catches most errors already, a correct output would be:
```clojure
"Add `a` and `b`."
```

Another optional feature could be detecting reference to other symbols in the project. Assuming for example there are two functions, `foo` and `bar`, and `foo`'s docstring says:

"Does the opposite of bar"

A fix could be:
- Possible reference to var 'bar' should appear as [[bar]] in the doc string.

Does this seem sensible?

joodie:
Hi, I've just spotted this ticket.

I've been thinking about this problem recently, also inspired by `checkdoc`, and [the docstring rules in the clojure style guide](https://guide.clojure.style/#documentation). I've got some really minimal implementation working as hooks already.

Is there any interest in a PR that would add at least some of these rules?

  - First line of the docstring should be a complete, capitalized sentence ending with punctuation.
  - Docstring should not have trailing or leading whitespace
  - Arguments should be mentioned in docstrings.
  - Arguments in docstrings should be quoted
 
I would be happy to work on the feature.

borkdude:
@joodie Yes, a PR is welcome.

borkdude:
Already implemented?

[...]

PR #1489, opened 2021-12-09
Add simple docstring formatting linters.

joodie:
This PR is in response to #805 

It adds basic formatting checks on docstrings; leading and trailing whitespace, blank docstrings and summaries (first line of docstring should be a complete sentence).

Some notes on the implementation

 - These only lint existing docstrings; use the existing `:missing-docstring` linter to check that docstrings exist at all.
 - The naming of the linters is taken from the [clojure styleguide](https://guide.clojure.style/#documentation).
 - All linters except "blank docstring" are off by default - there is a blank docstring in `clojure/core_deftype.clj` which triggers the `script/diff` check.
 - This patch also modifies the analyser to find docstrings in `defprotocol` and set `:doc` in the resulting vars metadata.
 - `:doc` entries in metadata on symbols are linted but the resulting findings point to the wrapping expression (def* form) because of technical difficulties.

For future improvements, I'm thinking of checking that function/method arguments are mentioned and correctly quoted in the docstring. The current PR is already large enough though.

---

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -18,6 +18,7 @@
    [clj-kondo.impl.analyzer.test :as test]
    [clj-kondo.impl.analyzer.usages :as usages :refer [analyze-usages2]]
    [clj-kondo.impl.config :as config]
+   [clj-kondo.impl.docstring :as docstring]
    [clj-kondo.impl.findings :as findings]
    [clj-kondo.impl.hooks :as hooks]
    [clj-kondo.impl.linters :as linters]
@@ -457,6 +458,7 @@
         call (name (symbol-call expr))
         var-leading-meta (meta name-node)
         docstring (string-from-token (first children))
+        doc-node (when docstring (first children))
         children (if docstring (next children) children)
         meta-node (when-let [fc (first children)]
                     (let [t (tag fc)]
@@ -472,7 +474,9 @@
         ;; use dorun to force evaluation, we don't use the result!
         _ (when meta-node (dorun (analyze-expression** ctx meta-node)))
         _ (when meta-node2 (dorun (analyze-expression** ctx meta-node2)))
+        [doc-node docstring] (or (docstring/docs-from-meta meta-node) [doc-node docstring])
         meta-node-meta (when meta-node (sexpr meta-node))
+        [doc-node docstring] (or (docstring/docs-from-meta meta-node2) [doc-node docstring])
         meta-node2-meta (when meta-node2 (sexpr meta-node2))
         var-meta (if meta-node-meta
                    (merge var-leading-meta meta-node-meta)
@@ -489,10 +493,11 @@
               ctx)
         private? (or (= "defn-" call)
                      (:private var-meta))
-        docstring (or (some-> meta-node2-meta :doc str)
-                      (some-> meta-node-meta :doc str)
-                      docstring
-                      (some-> var-leading-meta :doc str))
+        [doc-node docstring] (if docstring
+                               [doc-node docstring]
+                               ;; TODO: too late to get raw var-leading-meta node
+                               (when-let [doc (some-> var-leading-meta :doc str)]
+                                 [expr doc]))
         bodies (fn-bodies ctx children expr)
         _ (when (empty? bodies)
             (findings/reg-finding! ctx
@@ -542,6 +547,8 @@
                    :varargs-min-arity varargs-min-arity
                    :doc docstring
                    :added (:added var-meta))))
+    (when docstring
+      (docstring/lint-docstring! ctx doc-node docstring))
     (mapcat :parsed parsed-bodies)))
 
 (defn analyze-case [ctx expr]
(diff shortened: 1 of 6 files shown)


joodie:
I'm not sure why the `circleci: jvm` test fails.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -486,10 +490,11 @@
               ctx)
         private? (or (= "defn-" call)
                      (:private var-meta))
-        docstring (or (some-> meta-node2-meta :doc str)
-                      (some-> meta-node-meta :doc str)
-                      docstring
-                      (some-> var-leading-meta :doc str))
+        [doc-node docstring] (if docstring
+                               [doc-node docstring]
+                               ;; TODO: too late to get raw var-leading-meta node
```
borkdude, line 495: The doc-node locals are shadowing each other and that probably relates to this TODO.

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -382,9 +385,13 @@
...
                   (merge metadata meta-node-meta)
                   metadata)
-        docstring (or (some-> meta-node-meta :doc str)
-                       docstring
-                       (some-> metadata :doc str))
+        [doc-node docstring] (or (docstring/docs-from-meta meta-node)
+                                 [doc-node docstring])
+        [doc-node docstring] (if docstring
+                               [doc-node docstring]
+                               ;; TODO: too late to get raw metadata-node
```
borkdude, line 392: Another TODO here.

src/clj_kondo/impl/analyzer.clj
joodie: The shadowing is intentional here and in the other spots - in general docstrings can be specified as either 

 - string in the correct position of the ‘def*’ expression. 
 - :doc entry in attr map literal
 - :doc entry in metadata on the _symbol_ naming the var (using ‘^{:doc …} var-name’ reader syntax. 

since there can be only one docstring for a var, the code has to determine what to do if multiple ways of defining the docstring are used.  

This logic was already present  before this patch but I extended the code because the linter needs the ‘doc-node’  _node_ (not just the ‘docstring’ _content_) in order to report on the position of the docstring in the sourcecode. 

The above is compete for all cases except the “metadata on the symbol” case - the analyzer code handles that through  calls to ‘lift-meta-content2’ which drops the raw node information and I could not quickly find a good way to pass that through and I need to play with it a bit more. It’s a pretty dense bit of code and I don’t really get all the nuances of what it’s doing.

My guess is that the right solution here is actually slightly different from the current implementation - instead of calling ‘lint-docstring!’ Immediately from the analyzer, push docstring positon information into the generated metadata map and call lint-docstring when reg-var! Is called. That way you can also do docstring checks later on - for instance to check that vars mentioned in docstrings exist in the namespace.

joodie:
I’ll take a stab at resolving the metadata  TODOs explained above and you can evaluate if it’s an improvement, ok @borkdude ?  Also: some documentation on the lift-metadata2 function would help I think

borkdude:
I'll get back on this later this week.

borkdude:
I solved a merge conflict. If you allow me to push to your PR branch, then I'll push it there too.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

```
$ git push ssh://git@github.com/jomco/clj-kondo master
ERROR: Permission to jomco/clj-kondo.git denied to borkdude.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

joodie:
I just invited you to the repo with write access.

Joost.

On Wed, Dec 15, 2021, at 10:51, Michiel Borkent wrote:
> 
> 
> I solved a merge conflict. If you allow me to push to your PR branch, then I'll push it there too.
> 
> https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork
> 
> `$ git push ***@***.***/jomco/clj-kondo master
> ERROR: Permission to jomco/clj-kondo.git denied to borkdude.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.
`

borkdude:
@joodie Cool. Note that you can in general allow upstream forks to push to your PR branches with a setting, without making them a collaborator, although this works too :).

borkdude:
Eeks, something went terribly wrong there... I'll fix it.

src/clj_kondo/impl/analyzer.clj
borkdude: Right now we already had logic to detect which is _the_ docstring and this logic is now kind of duplicated.

> instead of calling ‘lint-docstring!’ Immediately from the analyzer, push docstring positon information into the generated metadata map and call lint-docstring when reg-var! Is called. That way you can also do docstring checks later on - for instance to check that vars mentioned in docstrings exist in the namespace.

Yes, agreed, that would be a much better approach.

borkdude:
So the main remaining thing is this: https://github.com/clj-kondo/clj-kondo/pull/1489#discussion_r769497631

We already have logic to detect which is _the_ docstring. Let's re-use that logic and add information to it, instead of doing this work twice (like walking the metadata to detect a docstring in the docstring namespace).

borkdude:
I'll take a stab at this if you want me to.

borkdude:
Solved remaining TODO items. FYI, lift-meta-content lifts metadata nodes as metadata s-expressions on the metadata of the node, so we can directly access it instead of traversing the nodes. These nodes are stored under the `:meta` keys on nodes and are replaced by their (merged) s-expr representation.

src/clj_kondo/impl/docstring.clj
```diff
@@ -0,0 +1,53 @@
...
+
+(defn lint-docstring!
+  "Lint `docstring` for styling issues.
+
+  `node` is the node reported when docstring has findings, so ideally
+  it should be the text node for `docstring`."
+  [{:keys [filename config] :as ctx} node docstring]
+  (when (some? docstring)
+    (doseq [{:keys [type message]} (docstring-messages docstring)]
+      (when-not (identical? :off (-> config :linters type :level))
```
borkdude, line 31: The purpose of doing these checks is to prevent unnecessary work if people aren't using these linters, but the work is done here anyway. This should change.

borkdude:
@joodie As far as I'm concerned, this PR is good to go now. I'll leave it up for a few hours (going to the gym). If I didn't hear any complaints, I'll merge it when I return.

joodie:
Yes great!

Issue #587, opened 2019-11-05
Detect duplicate case test constants


borkdude:
```
user=> (case 1 1 1 1 1)
Syntax error macroexpanding case at (REPL:1:1).
Duplicate case test constant: 1
```

viesti:
This reminds me of another `case` thingy, where symbols as test constants might lead to a bit unexpected result:

```clojure
user=> (def the-answer 42)
#'user/the-answer
user=> (let [result 42]
  #_=>   (case result
  #_=>     the-answer :yay
  #_=>     :lolnay))
:lolnay
```

Symbols as test constants might be rare enough so that such use might be worthy of a warning.

jrasanen:
@borkdude: should @viesti's case be a separate issue?

viesti:
Yeah, most probably, I'm known to fail at communication :D

borkdude:
Thanks. I added the clj-together label. If this project gets funded by Clojurists Together I can do it during the funding period.

borkdude:
@jrasanen @viesti Yes, that should be a separate issue, but I expect that issue to be hard to support, as you cannot really tell whether the symbols used in case are intentional or not.

PR #1491, opened 2021-12-10
Detect duplicate case test constants

mknoszlig:
- [x] This PR correponds to #587 
- [x] This PR contains a test

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -551,8 +551,19 @@
         (if (identical? :cljs (:lang ctx))
           [(ctx-with-linters-disabled ctx [:unresolved-symbol :private-call])
            nil]
-          [ctx {:quote? true}])]
+          [ctx {:quote? true}])
+        test-constants (take-nth 2 (next children))]
     (analyze-expression** ctx matched-val)
+    (doseq [[_ occs] (->> (if (odd? (count test-constants))
+                                 (drop-last test-constants)
+                                 test-constants)
+                          (group-by identity))
+            :when (> (count occs) 1)
+            :let  [dupe (last occs)]]
+      (findings/reg-finding!
+       ctx
+       (node->line (:filename ctx) dupe :duplicate-case-test-constant
+                   (format "Duplicate case test constant: %s" dupe))))
     (loop [[constant expr & exprs] (rest children)]
       (when constant
         (if-not expr
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -551,37 +551,30 @@
         (if (identical? :cljs (:lang ctx))
           [(ctx-with-linters-disabled ctx [:unresolved-symbol :private-call])
            nil]
-          [ctx {:quote? true}])
-        clauses        (rest children)
-        test-constants (reduce (fn [acc c]
-                                 (if (= :list (:tag c))
-                                   (into acc (:children c))
-                                   (conj acc c)))
-                               []
-                               (take-nth 2 (if (even? (count clauses))
-                                             clauses
-                                             (butlast clauses))))]
+          [ctx {:quote? true}])]
     (analyze-expression** ctx matched-val)
-    (doseq [[_ occs] (group-by identity test-constants)
-            dupe     (rest occs)
-            :when    (> (count occs) 1)]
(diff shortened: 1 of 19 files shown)


borkdude:
@mknoszlig Any idea why CI isn't running? Have you added this project to your own CircleCI account perhaps?

mknoszlig:
@borkdude i don't have a CircleCi account myself so that shouldn't be the problem. it may be because i added the [skip ci] tag to the doc commit (bad idea on my part). is there a way for you to trigger ci manually?  alternatively i'm happy to fix the commit msg on my side and try again.

borkdude:
Please do another "dummy" commit. I'll squash the commits anyway so that shouldn't hurt.

borkdude:
Hmm, somehow this error only happens on pull requests...

```
ERROR in (run!-test) (core.clj:4789)
file arguments jar file as string or file
expected: (= findings (:findings (clj-kondo/run! {:lint [(io/file (System/getProperty "user.home") ".m2" "repository" "org" "clojure" "spec.alpha" "0.2.194" "spec.alpha-0.2.194.jar")]})))
  actual: java.lang.ClassCastException: class java.io.File cannot be cast to class java.lang.CharSequence 
  ```

**Comment by mknoszlig** [comment]
yeah, had the same problem yesterday, iirc

**Comment by borkdude** [comment]
That seems to have helped.

**Comment by borkdude** [comment]
Now to the feedback of your PR:

We already have `analyze-case` in `analyzer.clj`. I would prefer if we did the linting there, so we don't have to dissect case node twice.

**Comment by mknoszlig** [comment]
excellent point, not sure how i missed that. i moved it into `analyze-case`

**Comment by mknoszlig** [comment]
i think i introduced a bug while moving it misbehaves for eg `(case :a 1 :a)`. I'll add the test case and a fix later tonight.

**Comment by mknoszlig** [comment]
@borkdude  ok, fixed. sorry for the back and forth... while fixing i realized that `(case :a 1 :a)` should almost be caught as a warning by a different linter that says "case with only one test clause should be replaced by `if`?". not sure if that's something that happens a lot in the wild, i personally don't use case all that often.

**Comment by borkdude** [comment]
I haven't looked closely into the implementation yet, but just checking: are you also catching:

``` clojure
(case x
  (:foo :bar) :yolo
  :bar :hello)
```

mknoszlig:
that's a good catch, i've added the test case. will update here once i've fixed it.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -551,9 +551,25 @@
...
+                                 (if (= :list (:tag c))
+                                   (into acc (:children c))
+                                   (conj acc c)))
+                               []
+                               (take-nth 2 (if (even? (count clauses))
+                                             clauses
+                                             (butlast clauses))))]
     (analyze-expression** ctx matched-val)
-    (loop [[constant expr & exprs] (rest children)]
+    (doseq [[_ occs] (group-by identity test-constants)
```
borkdude, line 565: It might be safer to use group-by `str` instead since the node sometimes can contain keys that do not make them equal, while their string representation is equal. And we can then bind the string value at the place of the underscore and use it in the error messag.

borkdude: I think we can make this code more efficient by adding another loop variable below and then walk the constants as we are analyzing the expression and emit the warning there, instead of walking it twice, as it happens now.

mknoszlig: updated accordingly

borkdude:
Exactly, thank you!

borkdude:
@mknoszlig We're still missing one (admittedly silly) edge:

Screen Shot 2021-12-13 at 14 30 32

mknoszlig:
hm, in cases (hehe) where these groups become large it might not even be so silly, who knows ... i have a patch in my master, do i need to make an additional PR or ist there some magical way to attach to one that's already merged?

borkdude:
@mknoszlig More edge cases, but I'm not sure if we should attack that one in this linter, or that we should just warn against using quotes in case test constants instead.

When using a quoted test constant, this is expanded into essentially `(quote the-constant)`:

So the below returns `:quote-a`.

``` clojure
(prn (case 'quote
       ;; a :a
       'a :quote-a ;; duplicate
       ;; '[a] :foo
       ;; [a] :bar ;; duplicate
       1))
```

Issue #1149, opened 2021-01-28
Suggestion: warn if the name of a binding that *is* in use starts with an underscore


aviflax:
**Is your feature request related to a problem? Please describe.**

I find it very handy when clj-kondo warns me of unused bindings. Sometimes I’ll just prepend an underscore to the name of the binding, as it can sometimes be handy as a form of documentation.

Then, sometimes I will later end up using the binding for something, and sometimes I will forget to remove the underscore from the front of the name.

It’s an edge case of an edge case, but it feels to me like the kind of thing that would match cli-kondo’s “I’ve got your back” ethos.

**Describe the solution you'd like**

A warning if/when the name of a binding starts with an underscore (_) *and* the binding is in use (referenced).

**Describe alternatives you've considered**

It’s quite an edge case, so we could just skip it. It’s not a big deal at all.

**Additional context**

🤷

----

Thanks!

borkdude:
@aviflax This question has come up a couple of times, so I'll leave this open. Do you have a suggestion for this linter? Maybe `:used-underscored-binding`?

aviflax:
Yeah, that’ll work nicely. A slightly different possible variation might be `underscored-binding-in-use` but I’d think consistency with the rest of the linter rules is more important, and I’m not super familiar with them right now.

borkdude:
Released.

PR #1497, opened 2021-12-13
warn if the name of a binding that *is* in use starts with an underscore

mknoszlig:

- [x] This PR correponds to an #1149 

- [x] This PR contains a test

the config key to switch it off is `exclude-incorrectly-marked-unused`. i've kept it in the `:unused-binding` linter for now, it seems more like a feature of that linter than a separate linter.


borkdude:
@mknoszlig Can you check the lint diff? There are quite a lot of new warnings and I'm curious if these new warnings are justified. https://github.com/clj-kondo/clj-kondo/runs/4509193910?check_suite_focus=true
I'm inclined to make this a new linter, called `:used-underscore-binding` which is off by default, but which be opted into.

mknoszlig:
fwiw, here's a case that seems justified: https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/types.clj#L283
but in general it looks like it's picking up a lot var names that are autogenerated by macros. i assume these can be filtered out by checking `:clj-kondo.impl/generated`.
happy to move it to a dedicated linter.

borkdude:
OK, let's try `:clj-kondo.impl/generated` first and see what remains.

mknoszlig:
This already removed a lot of the messages. The remainder is mostly cljs-related:
one is an edge case that appears a couple of times that's along the lines of
```clojure
(let [_ 1] _)
```
but most appear in the diff as:
```
< cljs/core.cljs:8495:48: warning: used binding __hash marked as unused.
718d684
< cljs/core.cljs:8766:58: warning: used binding __hash marked as unused.
735d700
< cljs/core.cljs:8973:35: warning: used binding _meta marked as unused.
741d705
< cljs/core.cljs:9052:35: warning: used binding _meta marked as unused.
```
this which appears to trace back to `.-` calls in cljs at first glance. if you want to take a look yourself i can push my current version.

borkdude:
I suspect `_meta` was used to not shadow the core var. I'm still inclined to make this an optional linter since in general I don't want people to have to deal with new warnings in their CI, unless they are real serious errors. This should be a minor tweak to what you already have now.

mknoszlig:
same here :) ... not sure yet what to do about the cljs stuff, though

borkdude:
There's nothing left to do. The lint diff is just a sanity check if we didn't overlook anything, but it seems good now.

borkdude:
We could make an exception for double underscored bindings, like `__hash` as a way to avoid the lint warning when you really do want to have underscores for whatever reason. This avoids most of the CLJS warnings.

mknoszlig:
yeah, crossed my mind as well, but will leave the `_meta` warnings in... if we special case this, we're open for false negatives which would be a bummer.
ok for the three test cases to live in unused-binding-test for the time being or do you prefer a strict separation by linter?

borkdude:
I think it's a different lint rule, although closely related. Let's separate it. This also leaves more room for adding options to it later without nesting everything deeply.

src/clj_kondo/impl/linters.clj
```diff
@@ -475,6 +475,28 @@
            (node->line filename node
                        finding-type msg)))))))
 
+(defn lint-used-underscored-bindings! [ctx]
```
borkdude, line 478: I actually think we can keep what you did before, linting the used and unused bindings in one go. This is better for performance probably, so let's revert that. We might just rename the function to `lint-bindings`.

doc/linters.md
```diff
@@ -751,6 +738,32 @@ This will disable the warning in:
...
 
+### Used underscored bindings
+
+*Keyword:* `:used-underscored-binding`.
+
+*Description:* warn when a underscored (ie marked as unused) binding is used.
+
+*Default level:* `:off`.
+
+*Example trigger:* `(cond :else 1 (odd? 1) 2)`.
```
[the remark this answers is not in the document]
mknoszlig: also remove the part that basically duplicates the summary, the note of how it is enabled remains

src/clj_kondo/impl/linters.clj
```diff
@@ -475,19 +475,47 @@
...
     (let [ns-config (:config ns)
           ctx (if ns-config
                 (assoc ctx :config ns-config)
                 ctx)
           ctx (assoc ctx :lang (:lang ns))]
+      (when-not (identical? :off (-> ctx :config :linters :used-underscored-binding :level))
+        (doseq [binding (set (->> (:used-bindings ns)
+                                  (filter (comp not :clj-kondo.impl/generated))
+                                  (filter #(str/starts-with? (str (:name %)) "_"))))
+                :when (not= "_" (str (:name binding)))]
```
borkdude, line 490: what is the use case for this line?

mknoszlig: you're right, makes no sense to ignore the `(let [_ 1] _)` case... i'll update.

src/clj_kondo/impl/linters.clj
```diff
@@ -484,9 +484,10 @@
                 ctx)
           ctx (assoc ctx :lang (:lang ns))]
       (when-not (identical? :off (-> ctx :config :linters :used-underscored-binding :level))
-        (doseq [binding (set (->> (:used-bindings ns)
-                                  (filter (comp not :clj-kondo.impl/generated))
-                                  (filter #(str/starts-with? (str (:name %)) "_"))))]
+        (doseq [binding (into #{}
```
borkdude, line 487: Just realized, we don't need a set at all, we can add `distinct` in the transducer.

mknoszlig: too late now  ;)

mknoszlig: but given the fact that distinct internally uses a set (even if it's volatile), it's not super surprising. but i realized i only rarely use transducers so dusting that off wasn't too bad :)

src/clj_kondo/impl/linters.clj
```diff
@@ -475,19 +475,47 @@
...
+(defn lint-bindings!
   [ctx]
   (doseq [ns (namespace/list-namespaces ctx)]
     (let [ns-config (:config ns)
           ctx (if ns-config
                 (assoc ctx :config ns-config)
                 ctx)
           ctx (assoc ctx :lang (:lang ns))]
+      (when-not (identical? :off (-> ctx :config :linters :used-underscored-binding :level))
+        (doseq [binding (sequence (comp (distinct)
```
borkdude, line 487: sorry but let's keep `(into #{] ...)` since it's way faster :)

src/clj_kondo/impl/linters.clj
```diff
@@ -475,19 +475,47 @@
...
       (when-not (identical? :off (-> ctx :config :linters :unused-binding :level))
         (let [bindings (:bindings ns)
               used-bindings (:used-bindings ns)
               diff (set/difference bindings used-bindings)
               defaults (:destructuring-defaults ns)]
+          (when-not (-> ctx :config :linters :unused-binding
+                        :exclude-incorrectly-marked-unused)
+            (doseq [binding (set (->> (filter (comp not :clj-kondo.impl/generated) used-bindings)
+                                      (filter #(str/starts-with? (str (:name %)) "_"))))
+                    :when (not= "_" (str (:name binding)))]
```
borkdude, line 509: This is still here

mknoszlig: weird, i thought i'd removed that a while ago.

src/clj_kondo/impl/linters.clj
```diff
@@ -475,19 +475,46 @@
...
+            :row (:row binding)
+            :col (:col binding)
+            :end-row (:end-row binding)
+            :end-col (:end-col binding)})))
       (when-not (identical? :off (-> ctx :config :linters :unused-binding :level))
         (let [bindings (:bindings ns)
               used-bindings (:used-bindings ns)
               diff (set/difference bindings used-bindings)
               defaults (:destructuring-defaults ns)]
+          (when-not (-> ctx :config :linters :unused-binding
```
borkdude, line 505: Isn't this part stale as well?

mknoszlig: 🤦 that explains it

src/clj_kondo/impl/linters.clj
```diff
@@ -501,13 +515,14 @@
                   :end-row (:end-row binding)
                   :end-col (:end-col binding)}))))
           (doseq [default defaults
-                  :let [binding (:binding default)]
+                  :let [binding (:binding default)
+                        nm (:name binding)]
```
borkdude, line 519: This seems to be an unnecessary change (keyword lookup can be prevented when linter doesn't apply).

mknoszlig:
thanks for the in depth feedback!
we missed one line, though: https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/config.clj#L38
can you fix it in master directly?

borkdude:
Done.

Issue #1167, opened 2021-02-10
unresolved vars for potemkin-imported vars when linting single file


ericdallo:
**version**
`2021.01.21-20210207.143254-18`

**platform**

JVM

**editor**

Confirmed the issue with Calva, clojure-lsp and JVM only via terminal.

**problem**

when clj-kondo lint a single file it reports `unresolved-var` even if the whole classpath was previously linted and cached it.

**repro**

First, we add a new unknown function [here](https://github.com/clj-commons/rewrite-clj/blob/v1/test/rewrite_clj/zip_test.cljc#L7) to test the unresolved-var: `(z/unknown)`

Then, we lint the whole classpath: 
```bash
$ clj -Sdeps '{:deps {clj-kondo {:mvn/version "2021.01.21-20210207.143254-18"}}}' -m clj-kondo.main --lint src test

test/rewrite_clj/zip_test.cljc:10:40: warning: Unresolved var: z/unknown
linting took 4882ms, errors: 0, warnings: 1
```
But when lining the single file I get:
```bash
$ cat test/rewrite_clj/zip_test.cljc | clj -Sdeps '{:deps {clj-kondo {:mvn/version "2021.01.21-20210207.143254-18"}}}' -m clj-kondo.main --lint -

<stdin>:10:28: warning: Unresolved var: z/of-string
<stdin>:10:40: warning: Unresolved var: z/unknown
<stdin>:26:26: warning: Unresolved var: z/find-tag-by-pos
<stdin>:27:26: warning: Unresolved var: z/replace
<stdin>:28:26: warning: Unresolved var: z/append-child
<stdin>:29:25: warning: Unresolved var: z/down
<stdin>:30:25: warning: Unresolved var: z/remove
<stdin>:31:25: warning: Unresolved var: z/root-string
<stdin>:47:16: warning: Unresolved var: z/rightmost
<stdin>:48:16: warning: Unresolved var: z/child-sexprs
<stdin>:57:16: warning: Unresolved var: z/up
<stdin>:68:16: warning: Unresolved var: z/right
<stdin>:82:17: warning: Unresolved var: z/insert-child
<stdin>:92:28: warning: Unresolved var: n/map-qualifier-node
<stdin>:106:36: warning: Unresolved var: z/node
<stdin>:106:43: warning: Unresolved var: n/map-context-clear
<stdin>:116:18: warning: Unresolved var: z/sexpr
[...]

PR #1499, opened 2021-12-15
Improve potemkin analysis

ericdallo:

Fixes https://github.com/clojure-lsp/clojure-lsp/issues/555

I think besides that issue this PR fixes/improve clj-kondo to warn about the not required namespace or consider required namespaces as used and don't complain about unused ones if used inside potemkin macro :)

### DIFF
--- a/corpus/import_vars.clj
+++ b/corpus/import_vars.clj
@@ -3,7 +3,11 @@
 (defn foo [_x]) ;; 1 arg
 
 (ns app.api
-  (:require [potemkin :refer [import-vars]]))
+  (:require
+   [clojure.walk]
+   [clojure.data]
+   [app.core]
+   [potemkin :refer [import-vars]]))
 
 (import-vars
  [clojure.walk
--- a/test/clj_kondo/main_test.clj
+++ b/test/clj_kondo/main_test.clj
@@ -2383,8 +2383,6 @@ i/blank? (i/starts-with?) i/x
 
 (defn foo []) ;; non-empty to generated ns cache
 
-(defmacro baz [])
-
 ;; dynamically generated baz
 
 (ns foo (:require [foo.bar] [potemkin :refer [import-vars]]))
--- a/test/clj_kondo/main_test.clj
+++ b/test/clj_kondo/main_test.clj
@@ -2303,12 +2303,12 @@ foo/foo ;; this does use the private var
 (deftest import-vars-test
   (assert-submaps
    '({:file "corpus/import_vars.clj",
-      :row 19,
+      :row 23,
       :col 1,
       :level :error,
       :message "clojure.walk/prewalk is called with 0 args but expects 2"}
      {:file "corpus/import_vars.clj",
-      :row 20,
+      :row 24,
       :col 1,
       :level :error,
       :message "app.core/foo is called with 0 args but expects 1"})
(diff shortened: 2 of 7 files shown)


test/clj_kondo/main_test.clj
```diff
@@ -2374,26 +2373,11 @@ i/blank? (i/starts-with?) i/x
...
+          [clojure.string]))
 
 (import-vars [clojure.string blank?, starts-with?, ends-with?, includes?])"
                      {:linters {:unresolved-symbol {:level :error}}})))
-  (is (empty? (lint! "
-(ns foo.bar)
-
-(defn foo []) ;; non-empty to generated ns cache
-
-;; dynamically generated baz
```
ericdallo, line 2386: @borkdude I removed this test because AFAICS this code doesn't compile, I didn't see any dynamically generated baz that would result in a valid `foo.bar/baz`, am I missing something?

test/clj_kondo/main_test.clj
```diff
@@ -2355,8 +2355,7 @@ foo/foo ;; this does use the private var
       (rename-path ".clj-kondo.bak" ".clj-kondo")))
   (testing "aliases"
     (assert-submaps
-     ' ({:file "<stdin>", :row 4, :col 14, :level :warning, :message "namespace clojure.string is required but never used"}
```
ericdallo, line 2358: Not unused anymore :yay:

corpus/import_vars.clj
```diff
@@ -3,7 +3,11 @@
 (defn foo [_x]) ;; 1 arg
 
 (ns app.api
-  (:require [potemkin :refer [import-vars]]))
+  (:require
+   [clojure.walk]
+   [clojure.data]
+   [app.core]
```
ericdallo, line 9: Required add the namespace now (like it's required to compile successfully this file on a REPL)

test/clj_kondo/main_test.clj
borkdude: The dynamically generated baz is a hypothetical one: suppose there was a macro in this namespace that generated it. Then the below code should still work, although clj-kondo did not see the baz var.

borkdude: Via git blame you can see that this test was added for https://github.com/clj-kondo/clj-kondo/issues/1167.

ericdallo: I see, this is not possible anymore with the current code right? unless we ad the macro that would expand to a var-definition, what do you think?

borkdude: > I see, this is not possible anymore with the current code right? 

Explain?

ericdallo: my bad, it's possible, it is just that we would need to include the macro to the test keep working, do you think that would be enough?

borkdude: I think you would know better than me since you reported #1167 :)

borkdude: Feel free to change the test, ensuring that it still fixes that issues.

ericdallo: makes sense! will do

ericdallo: Done @borkdude

borkdude: @ericdallo That is still not a correct fix. The point was that there could be a dynamic creation of the `baz` var in the namespace. You are introducing a static creation of the `baz` var.

ericdallo: Hum, so I'm not sure I understood how this dynamic creation works

ericdallo: The issue is that we get a foo.bar/baz unresolved with my changes, that it seems correct given this code, doesn't ?

borkdude: The fix for #1167 was that we should just assume the imported var to be there, even if it didn't exist. I'm fine with dropping this behavior, but your reported it yourself, so it may backfire on you.

PR #1512, opened 2021-12-18
Warn against 2 argument version of reduce

### LINKED ISSUE #1064: Optional linter: always use the 3-argument version of reduce
See https://purelyfunctional.tv/issues/purelyfunctional-tv-newsletter-313-always-use-the-3-argument-version-of-reduce/.

mknoszlig:

- [x] This PR fixes #1064

- [x] This PR contains a test

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -392,6 +392,18 @@ Normally a call to this macro will give an invalid arity error for `(select-keys
 {:linters {:invalid-arity {:skip-args [silly-macros/with-map]}}}
 ```
 
+### Two argument usage of reduce
+
+**Keyword:** `:two-argument-reduce`.
+
+*Description:* warn when reduce is called with 2 instead of 3 arguments.
+
+*Default level:* `:warning`.
+
+*Example trigger:* `(reduce + (range 3))`.
+
+*Example message:* `clojure.core/reduce called with 2 arguments. 3 argument form is recommended.`.
+
 ### Loop without recur
 
 *Keyword:* `:loop-without-recur`.
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1525,7 +1525,7 @@
             (recur (inc attempt) (rest args))))))
     (analyze-children ctx children false)))
 
-(defn analyze-hof [ctx expr resolved-as-name]
+(defn analyze-hof [ctx expr resolved-as-name hof-ns-name hof-resolved-name]
   (let [children (next (:children expr))
         f (first children)
         fana (analyze-expression** ctx f)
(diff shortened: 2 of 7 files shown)


src/clj_kondo/impl/linters.clj
```diff
@@ -354,6 +354,20 @@
                   (and call?
                        (not (utils/linter-disabled? call :single-logical-operand))
                        (lint-single-logical-operand call))]]
+      (when (and (not (utils/linter-disabled? ctx :two-argument-reduce))
+                 (= fn-name 'reduce)
+                 (= 2 arity)
+                 (#{'clojure.core 'cljs.core} fn-ns))
```
borkdude, line 360: Although this looks nice, I try to avoid this kind of set construct because of performance reasons, it's much faster to just do the boring thing here and writing it fully with `or` and `=`.

I think it makes sense to do this check in `analyzer`, we're already "casing" on core things there.

mknoszlig: That's why i put it in last position in the `and`, but i'm also fine with the boring variant, np.

When first looking into it, I tried to put this into `analyze-hof` right above the arity checks, which didn't work... now i see i misunderstood the meaning of `arity` in that context... moving to analyzer.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1604,6 +1604,17 @@
                              (node->line filename f
                                          :invalid-arity
                                          (linters/arity-error nil fn-name arg-count fixed-arities varargs-min-arity))))))))))
+    (when (and (not (utils/linter-disabled? ctx :two-argument-reduce))
+               (= 'reduce resolved-as-name)
```
borkdude, line 1608: This should probably be resolved-name

borkdude: I mean, the originally resolved-name from analyze-call, not the one that is in scope here

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1604,6 +1604,17 @@
...
+               (or (= 'clojure.core resolved-namespace)
+                   (= 'clojure.cljs resolved-namespace))
```
borkdude, line 1610: resolved-namespace is the resolved-namespace of the higher order function, not of reduce itself

borkdude: we can probably just add extra arguments to add the resolved-ns + resolved-name from analyze-call if we don't have them here

src/clj_kondo/impl/analyzer.clj
mknoszlig: afaict it's the same `resolved-as-name` that is being used to determine transducer eligibility, see:  https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/analyzer.clj#L1556
it seems to be i'm using it in a consistent manner (which may still be wrong of course)

borkdude: fair enough. it's just that when people lint something as `reduce`, they will get this warning for this linter.

mknoszlig: that's a good point, this linter is just super specific...

doc/linters.md
```diff
@@ -392,6 +392,18 @@ Normally a call to this macro will give an invalid arity error for `(select-keys
 {:linters {:invalid-arity {:skip-args [silly-macros/with-map]}}}
 ```
 
+### Two argument usage of reduce
+
+**Keyword:** `:two-argument-reduce`.
+
+*Description:* warn when reduce is called with 2 instead of 3 arguments.
+
+*Default level:* `:warning`.
```
vemv, line 401: Should have been :off right?

Regardless, great addition 🙌

borkdude: Corrected.

PR #1513, opened 2021-12-20
Repeated arities in multi-arity function

### LINKED ISSUE #1136: Repeated arities in multi-arity function

Doesn't detect error for this code:

```
(defn foo
  ([x] x)
  ([y] (inc y)))
```

A warning when this is detected.

**Additional context**

Clojure will catch this on eval, so not a big deal really.

mknoszlig:

- [x] This PR correponds fixes #1136 
- [x] This PR contains a  test

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -359,8 +359,9 @@
               arg-list (sexpr arg-vec)
               arity (analyze-arity arg-list)
               ret (cond-> {:arg-bindings (dissoc arg-bindings :analyzed)
-                           :arity (with-meta arity (meta body))
+                           :arity arity
                            :analyzed-arg-vec (:analyzed arg-bindings)
+                           :arg-vec arg-vec
                            :args arg-tags
                            :ret return-tag}
                     (:analyze-arglists? ctx) (assoc :arglist-str (str arg-vec)))]
@@ -388,7 +389,7 @@
   (let [docstring (:docstring ctx)
         macro? (:macro? ctx)
         {:keys [:arg-bindings
-                :arity :analyzed-arg-vec :arglist-str]
+                :arity :analyzed-arg-vec :arglist-str :arg-vec]
          return-tag :ret
          arg-tags :args} (analyze-fn-arity ctx body)
         ctx (ctx-with-bindings ctx arg-bindings)
@@ -436,6 +437,7 @@
            (concat analyzed-arg-vec analyze-pre-post parsed)
            :ret return-tag
            :arglist-str arglist-str
+           :arg-vec arg-vec
            :args arg-tags)))
 
 (defn fn-bodies [ctx children body]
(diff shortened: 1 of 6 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -359,7 +359,7 @@
               arg-list (sexpr arg-vec)
               arity (analyze-arity arg-list)
               ret (cond-> {:arg-bindings (dissoc arg-bindings :analyzed)
-                           :arity arity
```
mknoszlig, line 362: I'm not sure this is fine. `parsed-bodies` doesn't really map back to the original expression which i need for location info, the best alternative i could think of would be doing `(reduce (fn [] ...) (mapv list parsed-bodies bodies)` which feels a bit awkward as well.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -523,16 +523,26 @@
...
-                                (let [arg-tags (when (some identity args)
-                                                 args)
-                                      v (assoc-some {}
-                                                    :ret ret :min-arity min-arity
-                                                    :args arg-tags)]
-                                  (if varargs?
-                                    [:varargs v]
-                                    [fixed-arity v]))))
-                      parsed-bodies)
+        arities (reduce (fn [acc {:keys [:fixed-arity :varargs? :min-arity :ret :args] :as arity}]
```
mknoszlig, line 526: we could extract this part into a single function used by both `analyze-fn` and `analyze-defn`. Not sure what the reasoning is behind the current duplication so i didn't do anything yet,  but happy to extract if desired.

borkdude: feel free to extract

src/clj_kondo/impl/config.clj
```diff
@@ -13,6 +13,7 @@
     ;; :config-in-comment {} config override for comment blocks
     :linters {:invalid-arity {:level :error
                               :skip-args [#_riemann.test/test-stream]}
+              :repeated-arity {:level :error}
```
borkdude, line 16: naming suggestion: conflicting-fn-arity or conflicting-fn-overload
in my opinion it communicates the problem better: you can't have more than one, whereas repeated is more ambigious

src/clj_kondo/impl/analyzer.clj
borkdude: Two things:

- I think pointing to the vector would be fine too.
- An alternative to the meta idea is to assoc `arg-vec` into the arity map and pull it out, only when you actually need the location in case of an error.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -448,6 +450,28 @@
...
+              (if varargs?
+                (assoc acc :varargs v)
+                (do
+                  (when (get acc fixed-arity)
+                    (findings/reg-finding! ctx
+                                           (node->line
+                                            (:filename ctx)
+                                            arg-vec
+                                            :conflicting-fn-arity
+                                            (format "More than one function body with arity %s."
```
borkdude, line 469: let's replace `body` with `overload` since body usually refers to the thing after the argument vector, see
https://github.com/clojure/core.specs.alpha/blob/7d67e2f863ea28406f03338a7b92e2026d555062/src/main/clojure/clojure/core/specs/alpha.clj#L79

src/clj_kondo/impl/analyzer.clj
mknoszlig: adapted to `arg-vec` ... i think it would be a bit confusing to just use the metadata this way on data that's not a direct parsing result. i kind of like the idea that derived data can point to a source location via metadata as well. if this were to be introduced it would probably make more sense in a larger scope and also documented, otherwise it'll be more confusing than helpful... do you think that would be worth pursuing?

borkdude: >  i think it would be a bit confusing to just use the metadata this way on data that's not a direct parsing result. 

Are you referring to something which is the case _after_ your latest commit? I don't understand.

mknoszlig: no, i was referring to my initial solution that transfers the metadata of the function body to `arity`. now i'm just passing around the `arg-vector`.

borkdude: Sometimes results aren't maps and then the only solution (without changing a lot of code) is to attach it to metadata. I don't have a strong opinion on this approach, but I think we should not introduce changes without a good reason (problem statement, issue).

borkdude:
@mknoszlig Thanks!

I've noticed that `extract-arity-info` runs twice for a single function. This has always been the case and I've wondered if there's way to do better. This is done because we need to know the arities beforehand (I don't recall the exact details). But I don't think we need to run the conflicting arity check twice. Perhaps something clever can be done about this to prevent unnecessary work with respect to this linter but also in general (the latter not being a critical part of this PR, but still welcome to look into it).

borkdude:
I'll just merge this now and we can look into the improvement as an optional follow up.

PR #1541, opened 2022-01-15
Fix #783 (keyword map type inference) and ret map for function calls

### LINKED ISSUE #783: :keys can't be used in  :ret position
**version**

clj-kondo v2020.02.15

**platform**

native

**editor**

command line

**problem**

the `:keys` works in function `:args`, but not in `:ret`:

**repro**

```bash
echo '(do (defn fun2 [m] (str (:a m))) (+ 1 (:a (fun2 {:a "1"}))))' | clj-kondo --lint - --config '{:linters {:type-mismatch {:namespaces {user {fun2 {:arities {1 {:args [{:op :keys, :req {:a :int}}], :ret {:op :keys, :req {:a :string}}}}}}}}}}'
<stdin>:1:53: error: Expected: integer, received: string.
linting took 14ms, errors: 1, warnings: 0
```

**expected behavior**

... the `:ret` should be identified as map with :a key of type `:string` and the `:a` lookup on it should return a `:string`.

pfeodrippe:
Please answer the following questions and leave them below in as part of your PR.

Fix #783 (keyword map type inference) and ret map for function calls.
 
### Main changes (or proposal as I'm very new to the codebase) 
- Instead of calling `lint-keyword-call!` when in a keyword call, call `analyze-call`
- Uses new `:kw-call` key inside `:call` to be used for dynamic type dispatch for keywords


src/clj_kondo/impl/analyzer.clj
```diff
@@ -2115,30 +2115,59 @@
...
+      (findings/reg-finding! ctx
+                             (node->line (:filename ctx) expr :invalid-arity
+                                         (format "keyword :%s is called with %s args but expects 1 or 2"
+                                                 kw-str
+                                                 arg-count))))
+    ;; With one argument to the keyword call, we can check if the return is a map.
+    ;; Another linter warning we could create is to check if the 2-arity version
+    ;; will never return the default value.
+    (when (= arg-count 1)
+      (let [analyzed (first (analyze-children (update ctx :callstack conj [nil :list]) (rest (:children expr))))
```
pfeodrippe, line 2146: Here I want to check the return type of the unique argument, is there a better way to do it?

src/clj_kondo/impl/analyzer.clj
```diff
@@ -2115,30 +2115,59 @@
...
+              (get (:lang call))
+              (get (:ns call))
+              :vars
+              (get (:name call))
+              :arities
+              (get (:arity call))
+              :ret
+              :val
+              (get kw)
+              :tag))))))
```
pfeodrippe, line 2170: Ugly, is there some better way to do it?

borkdude:
@pfeodrippe Thanks! I will take a look soon.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -2115,30 +2115,59 @@
...
+                             ;; because we couldn't resolve the namespaced
+                             ;; keyword, we print it as is
+                             (str ":" (namespace kw)))
+                         ;; if the keyword is namespace, but there is no
+                         ;; namespace, it's the current ns
+                         (:name ns))
+                       (namespace kw))
+        kw-str (if ?resolved-ns (str ?resolved-ns "/" (name kw))
+                   (str (name kw)))]
+    (when (and (not (config/skip? config :invalid-arity callstack))
```
borkdude, line 2134: why was this moved?

src/clj_kondo/impl/analyzer.clj
borkdude: This works most of the time but isn't always reliable. More reliable is to insert an :id on the node you're interested in and then check the `:id` on the returned value. This is done elsewhere in the code.

src/clj_kondo/impl/analyzer.clj
pfeodrippe: So we are able to use the existing bindings, or type warnings should not work if invalid arity is skipped?

src/clj_kondo/impl/analyzer.clj
pfeodrippe: Great! Will take a look on it later, thanks o/

src/clj_kondo/impl/analyzer.clj
borkdude: Yeah, if you call the invalid arity, you don't know which overload was called, so we don't emit type warnings then.

src/clj_kondo/impl/analyzer.clj
borkdude: Instead of looking this up at runtime, we postpone this until the `impl/linters.clj` phase, where we have a more complete picture of what is being called (e.g. from other namespaces).

pfeodrippe: Keyword calls do not keep track of the argtypes atom, so I was not finding a good way to do it there, but I will try to go deeper into the code, thanks o/

borkdude: Hmm, good point, I don't know the right answer either until I take a deeper look.

pfeodrippe: I will point to the locations of the `:resolver` key (the function which converts types dynamically) when you have some time o/

See https://github.com/clj-kondo/clj-kondo/pull/1541/files#diff-3f1a249587095acd795d57380e88790caca3cead9f2924b337637e2d905a3ff9R230-R247, which is used at https://github.com/clj-kondo/clj-kondo/pull/1541/files#diff-f0dd386221a7c2f1794f9ced61a738b313001f4fbc5336dc828b392e4d40e2bcR49.

borkdude:
@pfeodrippe In line with developer documentation, which according to your PR description you've read, could you stop using force push? Thanks :) See. https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#force-push

borkdude:
The linting diff is crashing:

> Linting and writing output to /tmp/clj-kondo-diff/branch.txt
Error: Exception in thread "main" java.lang.RuntimeException: java.lang.Exception: Not supported: class clj_kondo.impl.types$ret_tag_from_call$fn__17403

pfeodrippe:
> @pfeodrippe In line with developer documentation, which according to your PR description you've read, could you stop using force push? Thanks :) See. https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#force-push

Yes, I have read, sorry. I have to remove some old habits.

pfeodrippe:
> The linting diff is crashing. I will take another look after this is fixed.
> 
> > Linting and writing output to /tmp/clj-kondo-diff/branch.txt
> > Error: Exception in thread "main" java.lang.RuntimeException: java.lang.Exception: Not supported: class clj_kondo.impl.types$ret_tag_from_call$fn__17403

Ah, it is green, this is what I didn't look at it. Will take a look, thanks

borkdude:
I think the work in this branch is promising, although I still need to find the time to properly review it.

@pfeodrippe Is there anything that could be split up into a smaller PR with trivial changes that are easy to merge in already?

@ikitommi Perhaps you could test this branch with malli to see if map spec inference results in more warnings.

borkdude:
The `(def ...)` one is probably not one to fix in this PR but a more general one. This isn't caught yet:

``` clojure
(def y :foo)
(inc x)
```

borkdude:
@pfeodrippe I think I found a bug:

Screen Shot 2022-01-24 at 20 03 06

src/clj_kondo/impl/analyzer.clj
borkdude: Still waiting on this one, I think?

pfeodrippe:
> @pfeodrippe I think I found a bug:
> 
> Screen Shot 2022-01-24 at 20 03 06

Nice, thanks! I will take a look later when is night here

pfeodrippe:
> I think the work in this branch is promising, although I still need to find the time to properly review it.
> 
> @pfeodrippe Is there anything that could be split up into a smaller PR with trivial changes that are easy to merge in already?

I guess not so much, this PR is dependent on the ret map which was added. We have to consider if the addition of `:kw-call` is the right path to go o/

borkdude:
Indeed, that's the part I'm thinking about too.

pfeodrippe:
@borkdude Now I'm using `:kw-calls` to "get-in" over multiple keyword nested calls as only the outer call is properly analyzed. See https://github.com/clj-kondo/clj-kondo/pull/1541/commits/a4c4f3c7a32b1d07f6facec0d5451e304fc4a44f, it has some duplication in the cond, but we can refactor if we want, let me know what you think about the approach o/

The idea would be not to handle all the cases in this PR (as you already mentioned), but at least not to create regressions while adding these keyword typing inferences.

I don't want to get too complex in my first PR for `clj-kondo` o/ It's an interesting issue though.

borkdude:
@pfeodrippe Playing around with the RP . I'm trying to avoid going through `analyze-call` to not intertwine it too much with var calls (not sure yet if this is a good approach either, but just to get my feet wet with this PR). I've come up with this:

```
(defn analyze-keyword-call [ctx expr arg-count k]
  (let [arg-types (atom [])
        ctx (assoc ctx :arg-types arg-types)
        analyzed (analyze-children ctx (rest (:children expr)) false)
        m (meta analyzed)
        ns-name (-> ctx :ns :name)
        proto-call (merge {:type :call
                           :resolved-ns :clj-kondo/unknown-namespace
                           :ns ns-name
                           :name (keyword (name k))
                           :unresolved? true
                           :unresolved-ns true
                           :arity arg-count
                           :base-lang (:base-lang ctx)
                           :lang (:lang ctx)
                           :filename (:filename ctx)
                           :arg-types arg-types}
                          (meta expr))
        ret-tag (or (:ret m)
                    (types/ret-tag-from-call ctx proto-call expr))
        call (cond-> proto-call
               ret-tag (assoc :ret ret-tag))]
    (namespace/reg-var-usage! ctx ns-name call)
    (if m
      (with-meta (cons call analyzed)
        m)
      (cons call analyzed))))
```

Most tests are passing, but 7 of them are not. I'm probably overseeing something.

borkdude:
Hmm, got it down to 1 failing test now in the `pfeodrippe-fix-keyword-map-type-inference` branch on the clj-kondo repo.

borkdude:
I solved it a different way now by copying analyze-call and ripping out everything unrelated to keywords.

borkdude:
@pfeodrippe Slowly making progress in tweaking things here: https://github.com/clj-kondo/clj-kondo/pull/1561/files

I'll be updating that in the coming time until I think it's ready and then I'm going to ask you again for a final review of that.

pfeodrippe:
> I solved it a different way now by copying analyze-call and ripping out everything unrelated to keywords.

Thanks, it makes sense! I didn't analyzed the analyze-call function (no pun intended) in depth as it was too big, thanks for making it more reasonable for the keyword call o/ 

Sorry for creating work for you, man, didn't  want to bother

borkdude:
@pfeodrippe No problem! It's going to be worth it and it has to be done at some point anyway :-D

borkdude:
>  the analyze-call function (no pun intended) in depth as it was too big

This is one of the problems with that function, I'm not very proud of it.

borkdude:
Merged via other PR

PR #1551, opened 2022-01-20
match on :depends instead of defaulting to last element.

mknoszlig:

This was caused by an insufficient test case. The task maps only contained one element each, so the `last` used on the task definition entries always chose a entry for `:depends`. Fixed by extending the conditions in `doseq` to do a proper match.

### DIFF
--- a/src/clj_kondo/impl/linters/deps_edn.clj
+++ b/src/clj_kondo/impl/linters/deps_edn.clj
@@ -188,14 +188,16 @@
   (let [tasks (edn-utils/sexpr-keys expr)
         known-task? (set (keys tasks))]
     (doseq [[_ t-def] tasks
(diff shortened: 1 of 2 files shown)


src/clj_kondo/impl/linters/deps_edn.clj
```diff
@@ -188,8 +188,10 @@
   (let [tasks (edn-utils/sexpr-keys expr)
         known-task? (set (keys tasks))]
     (doseq [[_ t-def] tasks
-            dep-task   (:children (last (:children t-def)))
+            [td-key td-body]   (partition 2 (:children t-def))
```
borkdude, line 191: you could also use `sexpr/keys` here and just get the `:depends` key out of that perhaps.

PR #1561, opened 2022-01-26
Pfeodrippe fix keyword map type inference (draft PR to simplify existing PR, hopefully...)

borkdude:


borkdude:
@pfeodrippe I think the branch is ready for merge now... Can you take one last look?

PR #1562, opened 2022-01-26
do not use qualified keywords for built in linters.

mknoszlig:


borkdude:
Thanks! ❤️

PR #1565, opened 2022-01-31
Update setup-clj-kondo example

danielcompton:
The commit referenced in the example points to an old version of setup-clj-kondo.

Since https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
the old version of the Action no longer runs, but the updated version does run.

- [ ] This PR correponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).

### DIFF
--- a/doc/ci-integration.md
+++ b/doc/ci-integration.md
@@ -73,10 +73,10 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-    - name: Setup clj-kondo
-      uses: DeLaGuardo/setup-clj-kondo@822352b8aa37d5c94135e67f7b4e2f46c08008a8
+    - name: Install clj-kondo
+      uses: DeLaGuardo/setup-clj-kondo@afc83dbbf4e7e32e04649e29dbf30668d30e9e3e
       with:
-        version: '2020.04.05'
+        version: '2022.01.15'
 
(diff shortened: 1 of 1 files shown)


doc/ci-integration.md
```diff
@@ -73,10 +73,10 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-    - name: Setup clj-kondo
-      uses: DeLaGuardo/setup-clj-kondo@822352b8aa37d5c94135e67f7b4e2f46c08008a8
+    - name: Install clj-kondo
```
danielcompton, line 76: I can revert this line change if you want.

danielcompton:
Best security practice for Actions is to [pin a SHA](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) rather than a tag or branch, as those can be updated by the Actions author later. Based on the [README](https://github.com/marketplace/actions/setup-clj-kondo) though, it looks like `@master` should work.

The issue which prompted this was a one-time breaking change in Actions. I updated the clj-kondo version while I was there, but you shouldn't need to update the version after every clj-kondo release.

PR #1567, opened 2022-02-03
implement some more basic linting for bb.edn

### LINKED ISSUE #1560: lint task definition keys in bb.edn
It would be nice to have some support when writing tasks in bb.edn. following are a couple I have in mind:

- `unknown-key`: any key that is not in  `#{:doc, :requires, :task, :extra-paths, :extra-deps, :enter, :leave}`
- `task-key-required`: missing `:task`
- `doc-key-recommended`: optional, warn on missing `:doc`

mknoszlig:
fixes #1560 

- [x] This PR correponds to #1560 

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -286,10 +286,26 @@ Depending on undefined task: compile
          b {:depends [a]}}}
 ```
 
+### Bb.edn global requires
+
+*Keyword:* `:bb.edn-global-requires`
+
+*Description:* warn on top level `:requires` key.
+
+*Default level:* `:warning`
+
+*Example trigger:*
+
+`bb.edn`:
+
+``` clojure
+{:requires [[babashka.fs :as fs]]}
+```
+
 *Example message:*
 
 ```
-Cyclic task dependency: a -> b -> a
+Global :requires belong in the :tasks map.
 ```
 
 ### Docstring blank
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -292,11 +292,11 @@ Depending on undefined task: compile
 Cyclic task dependency: a -> b -> a
 ```
 
-### Bb.edn global requires
+### Bb.edn Unexpected key
 
-*Keyword:* `:bb.edn-global-requires`
+*Keyword:* `:bb.edn-unexpected-key`
 
-*Description:* warn on top level `:requires` key.
+*Description:* warn on unexpected keys in bb.edn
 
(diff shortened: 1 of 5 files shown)


doc/linters.md
```diff
@@ -286,10 +286,48 @@ Depending on undefined task: compile
...
+
+*Example message:*
+
+```
+Global :requires belong in the :tasks map.
+```
+
+### Bb.edn task docstring missing
+
+*Keyword:* `:bb.edn-task-docstring-missing`
```
borkdude, line 313: `bb.edn-task-missing-docstring` would me more consistent with the other `missing-docstring`

borkdude: I think this warning should be disabled by default for "private" tasks (the task name starts with `-` or has `:private true`).

borkdude: Good point. It seems the missing-docstring linter also warns for private functions, so let's just keep the behavior as you have it now.

mknoszlig: we could reverse it and have it complain for private by default. that way it's consistent with missing-docstring but could also be customized further?

mknoszlig: fwiw i'm halfway there already :)

borkdude: > we could reverse it and have it complain for private by default. that way it's consistent with missing-docstring

Let's just do that, for consistency and leave it at that for now.

mknoszlig: ok, so just added test cases for private tasks as well so the behaviour is covered

doc/linters.md
```diff
@@ -292,6 +292,50 @@ Depending on undefined task: compile
 Cyclic task dependency: a -> b -> a
 ```
 
+### Bb.edn global requires
+
+*Keyword:* `:bb.edn-global-requires`
```
borkdude, line 297: I think this could be a more general linter with a more general name. Instead of having a linter for each un-expected top level key, I think we could have `:bb.edn-unexpected-key` to lint for `:init`, `:requires`, and whatever more.

mknoszlig: Makes sense. What's your opinon on keeping the error message specific to the key in question? (eg being explicit about `:requires` belonging within the task map). i guess the mistake is quite clear for most cases, otoh i can't recall ever complaining about an error message or warning being too specific...

borkdude: Yeah, specific error message is helpful, agree

PR #1595, opened 2022-02-23
Support analysing re-frame/dispatch flavours and inject-cofx

benedekfazekas:
Following up the re-frame work with analysers for different flavours of dispatch and inject-cofx

- [ ] This PR correponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).


src/clj_kondo/impl/analyzer/re_frame.clj
```diff
@@ -40,6 +41,20 @@
           (common/analyze-children ctx args)))
       (common/analyze-children ctx children))))
 
+(defn analyze-inject-cofx [ctx expr ns]
+  (let [kns (keyword ns)
+        [cofx-id cofx-param] (next (:children expr))]
+    (common/analyze-children (assoc-in ctx [:context kns :cofx-ref] true) [cofx-id])
+    (when cofx-param
+      (common/analyze-children ctx [cofx-param]))))
```
borkdude, line 49: I would just lint all the children from `(next (:children exor))` here, even though inject-cofx only supports two arguments, this will analyze all the arguments the user passes and will making linting more correct.

borkdude: Same for below

benedekfazekas: pushed, hope i covered all cases

src/clj_kondo/impl/analyzer/re_frame.clj
```diff
@@ -49,6 +69,45 @@
...
+(defmulti analyze-dispatch-type (fn [_ctx _fq-def x] (:k (first x))))
+
+(defmethod analyze-dispatch-type :dispatch [ctx fq-def x]
+  (let [[disp-kw & _dispatch-vector] x]
+    (analyze-dispatch ctx {:children x} (str (namespace fq-def)))
+    [disp-kw]))
+
+(defmethod analyze-dispatch-type :dispatch-n [ctx fq-def x]
+  (let [disp-kw (:k (first x))]
+    (doseq [dispatch-vector (:children (first (drop 1 x)))]
```
borkdude, line 82: Isn't (first (drop 1 ...)) the same as second?

borkdude:
@benedekfazekas I'm a bit concerned about the postwalk in analyze-reg-event-fx and then analyzing during the postwalk + also analyzing the body later on. This results in some nodes being analyzed twice which could have averse consequences: double warnings, but also for performance we should try to prevent this. Is there a better way to accomplish the same results or perhaps you can explain why this wouldn't be a problem. Also I'd like to have some more comments in the code to understand what exactly is being done, as I find it hard to grok.

benedekfazekas:
comments: fair point. will add.

postwalk: I suppose postwalk or something similar is necessary here as the result map can be anywhere in the reg-event-fx body. on the up side any re-frame specific analysis is only triggered if we find a collection which starts with one of the keywords (`:dispatch`, `:dispatch-n`, `:dispatch-later`). also I think we only double analyse those dispatch keywords quoted above, maybe i could get rid of that double analysis by returning just `[]` from the defmethods? would that suffice? I can look into these on Monday/Wednesday next week.

borkdude:
I think that sounds reasonable. Thanks! No hurry from my side, take your time.

benedekfazekas:
coolio, I'd like to wrap this up soon, thanks for the good feedback!

benedekfazekas:
@borkdude docstring added and eliminated the dbl analysis as discussed

lassemaatta:
Fyi: there's also the [:fx](http://day8.github.io/re-frame/api-builtin-effects/#fx) variant of dispatching, which also deprecates `:dispatch-n`.

benedekfazekas:
:wave: @lassemaatta I think that is covered here, see https://github.com/clj-kondo/clj-kondo/blob/master/test/clj_kondo/analysis_test.clj#L1620 if you think I've missed something, can you pls provide a code snippet that I can put in a test as in the quoted test ns?

Issue #1240, opened 2021-03-27
Consider linting for namespace names that do not match source filename


lread:
**Is your feature request related to a problem? Please describe.**
While testing out bb 3.1 SNAPSHOT tasks on my rewrite-clj scripts, I noticed that a few of my bb scripts had wonky namespace names. For examples:
- I had accidentally used `clj_tests` instead of `clj-tests` for a `clj_tests.clj` script. 
- Or I had maybe renamed a script filename and not adjusted the namespace name within the file.

I noticed my error because wonky namespace names do not matter for bb scripts, but they do for bb 3.1 SNAPSHOT tasks.

**Describe the solution you'd like**
Warn when a namespace name does not match the dirs/filename.

A linter perhaps called `:namespace-name-mismatch` with message:
`Namespace name foo_booboo.bar does not match expected foo-boo.boo.bar`

**Describe alternatives you've considered**
Could do nothing: one discovers pretty quickly through usage when one's namespace names are wonky, but I think a linter would be nice here. You could discover your error more quickly in the editor.

**Additional context**
I suppose an editor linter integration would have to be smart enough to be aware of filename changes.

borkdude:
Related / duplicate: https://github.com/clj-kondo/clj-kondo/issues/842

But please keep it open since you have more details than the other one.

lread:
@borkdude, I just happened to notice this one is still open, can be closed ya?

borkdude:
Thanks

PR #1596, opened 2022-02-24
#1240: Add new linter :namespace-name-mismatch

svdo:

- [X] This PR correponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code). #1240 

### DIFF
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -418,7 +418,7 @@
         _ (let [filename (:filename ctx)
                 filename-to-periods (some-> filename
                                             (str/replace #"/" ".")
-                                            (.replaceAll ^String fs/file-separator "."))
+                                            (.replace ^CharSequence fs/file-separator "."))
                 munged-ns (str (munge ns-name))]
             (when-not (or (= "<stdin>" filename)
                           (= 'user ns-name)
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -423,9 +423,7 @@
             (when-not (or (= "<stdin>" filename)
                           (= 'user ns-name)
                           (and filename-to-periods 
-                               (or (str/ends-with? filename-to-periods (str munged-ns ".clj"))
-                                   (str/ends-with? filename-to-periods (str munged-ns ".cljs"))
-                                   (str/ends-with? filename-to-periods (str munged-ns ".cljc")))))
+                               (str/ends-with? (fs/strip-ext filename-to-periods) munged-ns)))
               (findings/reg-finding!
                ctx
                (node->line filename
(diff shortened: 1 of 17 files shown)


svdo:
@borkdude I think this is ready now as far as I'm concerned. Looking forward to your review comments!

borkdude:
@svdo I think the warning should be on the namespace name instead of the namespace form:

Screen Shot 2022-02-25 at 11 38 48

borkdude:
In babashka scripts and also Clojure scripts that are invoked from the command line or via `load-file` it is perfectly fine to use dashes in the file name. Is there a way we can avoid warnings for those?

Screen Shot 2022-02-25 at 11 43 15

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -414,6 +414,23 @@
...
                                     "namespace name expected"))))
                  'user)
+        _ (let [expected-file-path (-> (str (munge ns-name))
+                                       (str/replace #"\." "/"))]
+            (when-not (or (= "<stdin>" (:filename ctx))
+                          (= 'user ns-name)
+                          ;; Test for .indexOf because there can be something before
+                          ;; (e.g. 'corpus/' in the tests), and the "expected-file-path"
+                          ;; does not include the extension.
+                          (and (seq (:filename ctx)) (<= 0 (.indexOf ^String (:filename ctx) ^String expected-file-path))))
```
borkdude, line 424: Suggestions:

- Pull out `(:filename ctx)` to a let
- Don't call `seq` on the string, just test if the value is there
- Use `str/index-of` rather than interop

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -414,6 +414,23 @@
...
+              (findings/reg-finding!
+               ctx
+               {:type :namespace-name-mismatch
+                :message (str "Namespace name does not match file name: " ns-name)
+                :row (-> ns-name-expr meta :row)
```
borkdude, line 429: Pull out the ns-name-expr meta in let to avoid boilerplate.

src/clj_kondo/impl/analyzer/namespace.clj
borkdude: Another suggestion: just call `(str/index-of (str ...) ...)` on the filename: this should prevent any NPE issues

borkdude:
@svdo Windows tests are failing: https://ci.appveyor.com/project/borkdude/clj-kondo/builds/42706510

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -414,6 +414,21 @@
                                     :syntax
                                     "namespace name expected"))))
                  'user)
+        _ (let [expected-file-path (-> (str (munge ns-name))
+                                       (str/replace #"\." "/"))
```
borkdude, line 418: This probably causes the Windows tests to fail. Insteaf of hardcoding `"/"` use `babashka.fs/file-separator`

borkdude: I think it would be better to do this even:

- Always replace in the filepath "/" with "."
- On Windows, additionally replace in the filepath `babashka.fs/file-separator` with "."
- Munge ns-name
- Then check if filename ends-with? munged ns-name

svdo: Ok I think I made all the requested changes.

PR #1599, opened 2022-02-28
fix #1598: Give multi-arity fn args scope-end

### LINKED ISSUE #1598: [analysis] :scope-end-row is missing on multi-arity fn args
**version**

v2022.02.09

**platform**

JVM

**editor**

Emacs / emacs-lsp, although this probably isn't relevant.

**problem**

The locals defined by the args of a multi-arity function should contain `:scope-end-row` and `:scope-end-col`, but don't.

**repro**

```clojure
(defn foo ([a] (inc a)))
```

In the analysis, the local definition of `a` should include a `:scope-end-row` and `:scope-end-col`, to match the analysis of single-arity functions:

```clojure
(defn foo [a] (inc a))
```

mainej:
This gives args defined in multi arity fns the `:scope-end-col` and `:scope-end-row` properties, which args in single arity fns also have. Fixes #1598.

- [x] This PR correponds to #1598.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 For a list of breaking changes, check [here](#breaking-changes).
 
+## Unreleased
+
+### New
+
+### Fixed
+
+- [#1598](https://github.com/clj-kondo/clj-kondo/issues/1598): `:scope-end-row` is missing on multi-arity fn args ([@mainej](https://github.com/mainej))
+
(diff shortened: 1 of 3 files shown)


borkdude:
Merged via https://github.com/clj-kondo/clj-kondo/commit/8163afab67596ecc0bf384bb3a1951012c6db0aa. Thanks!

PR #1611, opened 2022-03-11
Support nested ^:replace overrides in config

dundalek:
Based on a [slack conversation](https://clojurians.slack.com/archives/CHY97NXE2/p1647025453421829)

Allows a workflow when one needs to replace linters, but keep default paths for custom hooks to be found.

### DIFF
--- a/src/clj_kondo/impl/utils.clj
+++ b/src/clj_kondo/impl/utils.clj
@@ -142,7 +142,8 @@
   ([])
   ([a] a)
   ([a b]
-   (cond (and (map? a) (map? b))
+   (cond (:replace (meta b)) b
+         (and (map? a) (map? b))
          (merge-with deep-merge a b)
          (and (sequential? a) (sequential? b))
          (into a b)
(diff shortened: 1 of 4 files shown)


src/clj_kondo/impl/utils.clj
```diff
@@ -142,7 +142,8 @@
   ([])
   ([a] a)
   ([a b]
-   (cond (and (map? a) (map? b))
+   (cond (:replace (meta b)) b
```
borkdude, line 145: the `:replace` lookup can be prevented when there is no meta

dundalek: Is `some->` a good option?

borkdude: This seems a tiny bit faster:

``` clojure
user=> (time (dotimes [i 100000000] (when-let [m (meta 1)] (:replace m))))
"Elapsed time: 1561.475506 msecs"
```

Issue #1638, opened 2022-03-30
Introduce re-frame testing code in corpus to prevent future breakage 


borkdude:
_Originally posted by @borkdude in https://github.com/clj-kondo/clj-kondo/issues/1637#issuecomment-1082735498_

@benedekfazekas I "fixed" this by reverting the analysis commit. Before introducing more analysis changes around re-frame we should first have some more re-frame testing in place that will prevent future breakage (this is the second time we experienced this).
I asked around on Twitter if there are open source projects using re-frame that we could use:

https://twitter.com/borkdude/status/1508933061534228483

Athens Research is one of them. I suggest we take their re-frame code and put it in corpus and lint for 0 warnings/errors. I'll make an issue of this.

benedekfazekas:
looking into this. athens (their `src/cljs` dir) does pass linting with the re-frame commit reverted and does report errors and warnings when linted with the re-frame commit

borkdude:
@benedekfazekas OK, let's add that to `corpus`, e.g. `corpus/re-frame/athens/src/cljs`

borkdude:
And then also add a test for it, I would say separately from the analysis tests: `test/clj_kondo/re_frame_test.clj` and then a test which should result in 0 warnings/errors linting that dir.

benedekfazekas:
PRed

benedekfazekas:
should be closed?

PR #1641, opened 2022-03-30
[#1638] Test to catch lint regressions caused by re-frame analysis

benedekfazekas:
add https://github.com/athensresearch/athens cljs code snapshot as is
now to lint against

### DIFF
--- a/test/clj_kondo/re_frame_test.clj
+++ b/test/clj_kondo/re_frame_test.clj
@@ -1,7 +1,40 @@
 (ns clj-kondo.re-frame-test
   (:require
+   [clj-kondo.core :as clj-kondo]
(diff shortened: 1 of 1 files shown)


borkdude:
@benedekfazekas Thanks! Wow, it's a bit much! Maybe we could do this another way: check out the athens repo at a specific commit and then run the tests over that.

borkdude:
@benedekfazekas We could just use tools.deps to define a dependency on athens and then lint their code maybe? This is already done in several places, e.g. check the analysis/java_test.

benedekfazekas:
will check analysis/java_test out

benedekfazekas:
maybe wrong but athens being an app (not a lib) I don't think it is on clojars or on maven central

PR #1658, opened 2022-04-16
Fix 1637

benedekfazekas:
Fix false positive caused by double analyzing some parts of a reg-event-fx at times and tests for the specific cases we saw those false positives with

### DIFF
--- a/src/clj_kondo/impl/analyzer/re_frame.clj
+++ b/src/clj_kondo/impl/analyzer/re_frame.clj
@@ -62,22 +62,24 @@
     (when cofx-params
       (common/analyze-children ctx cofx-params))))
 
-(defn analyze-dispatch
-  "Analyzes re-frame dispatch call.
-  Marks the event id, the first element of the parameter vector as an event reference so
+(defn- analyze-dispatch-event-id
+  "Marks the event id, the first element of the parameter vector as an event reference so
   it can be tracked back to the event registration. Marker is `[:context :re-frame.core :event-ref]`.
   https://day8.github.io/re-frame/api-re-frame.core/#dispatch"
   [ctx expr ns]
   (let [kns (keyword ns)
-        [farg & args :as children] (next (:children expr))]
-    (if (identical? :vector (utils/tag farg))
-      (let [[event-id & event-params] (:children farg)]
-        (common/analyze-children (assoc-in ctx [:context kns :event-ref] true) [event-id])
-        (when event-params
-          (common/analyze-children ctx event-params))
-        (when args
-          (common/analyze-children ctx args)))
-      (common/analyze-children ctx children))))
+        [farg & _args :as _children] (next (:children expr))]
+    (when (identical? :vector (utils/tag farg))
+      (let [[event-id & _event-params] (:children farg)]
+        (common/analyze-children (assoc-in ctx [:context kns :event-ref] true) [event-id])))))
+
+(defn analyze-dispatch
+  "Analyzes re-frame dispatch call.
+  Uses [[analyze-dispatch-event-id]] to analyze the event id. And analyzes the rest of the dispatch call
+  as usual."
+  [ctx expr ns]
+  (analyze-dispatch-event-id ctx expr ns)
+  (common/analyze-children ctx (next (:children expr))))
 
 (defn analyze-reg-sub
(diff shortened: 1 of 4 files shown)


src/clj_kondo/impl/analyzer/re_frame.clj
```diff
@@ -40,7 +50,46 @@
...
+      (common/analyze-children ctx cofx-params))))
+
+(defn- analyze-dispatch-event-id
+  "Marks the event id, the first element of the parameter vector as an event reference so
+  it can be tracked back to the event registration. Marker is `[:context :re-frame.core :event-ref]`.
+  https://day8.github.io/re-frame/api-re-frame.core/#dispatch"
+  [ctx expr ns]
+  (let [kns (keyword ns)
+        [farg & _args :as _children] (next (:children expr))]
+    (when (identical? :vector (utils/tag farg))
```
borkdude, line 72: @benedekfazekas What if the argument is not a vector, then it won't be analyzed? E.g.

```
(let [x ...]
  (dispatch x))
```

src/clj_kondo/impl/analyzer/re_frame.clj
```diff
@@ -49,6 +98,51 @@
...
+
+(defmulti analyze-dispatch-type (fn [_ctx _fq-def x] (:k (first x))))
+
+(defmethod analyze-dispatch-type :dispatch [ctx fq-def x]
+  (analyze-dispatch-event-id ctx {:children x} (str (namespace fq-def)))
+  x)
+
+(defmethod analyze-dispatch-type :dispatch-n [ctx fq-def x]
+  (let [disp-kw (:k (first x))]
+    (when (identical? :vector (utils/tag (second x)))
```
borkdude, line 110: Similar concern here

src/clj_kondo/impl/analyzer/re_frame.clj
benedekfazekas: essentially yes, this is a known limitation -- I think there was an issue about this as well and we discussed it a bit maybe on slack?! -- not sure. thinking about it but I don't think this is the scope if this PR. unless of course you think that without taking care of these cases this causes more harm than good

borkdude: I'm not talking about correct analysis, my concern is when skipping analysis for some parts, that will cause false positive lint warnings like "unused binding".

borkdude: This seems to work:
```
clojure -M:clj-kondo/dev --lint - <<< '(ns foo (:require [re-frame.core :as r])) (let [x nil] (r/dispatch x))'
linting took 34ms, errors: 0, warnings: 0
```

but why does that work?

benedekfazekas: because this fn just analyzes the event-id as the name shows. and this case it does not even analyzes that because it only understands the default format, i.e. a vector. the entry point for a direct dispatch call however -- see `analyze-dispatch` --, analyzes the whole form. does this make more sense?

borkdude: Makes sense

src/clj_kondo/impl/analyzer/re_frame.clj
```diff
@@ -40,7 +50,46 @@
...
+        (common/analyze-children (assoc-in ctx [:context kns :event-ref] true) [event-id])))))
+
+(defn analyze-dispatch
+  "Analyzes re-frame dispatch call.
+  Uses [[analyze-dispatch-event-id]] to analyze the event id. And analyzes the rest of the dispatch call
+  as usual."
+  [ctx expr ns]
+  (analyze-dispatch-event-id ctx expr ns)
+  (let [[name-expr & body] (next (:children expr))]
+    (common/analyze-children ctx (cons name-expr body))))
```
borkdude, line 83: Why are the args here destructured but then cons-ed again?

benedekfazekas: lazy refactor on my side. fixing...

benedekfazekas: done

borkdude:
Thanks!

PR #1674, opened 2022-04-29
Limit var analysis

mainej:
This PR adds config options to limit the amount of analysis that is done.
- `:var-usages false` excludes var-usages from the analysis output
- `:var-definitions {:shallow true}` includes var-definitions, but skips analysis of their bodies

If both options are set analysis is a little over 80% faster. This will be helpful for clojure-lsp, which can use them when scanning deps.

This PR also moves the `:analysis` config option to the top-level of `:config`, out of `:output`. Docs and tests have been updated to reflect this new location. To maintain backward compatibility, the code still accepts `:analysis` in the `:output` config.

Finally, this PR clarifies the meanings of `:analysis` and `:skip-lint`. The output will include `:analysis` only when the config includes `:analysis`. The `:findings` will be non-empty only if the config doesn't include `:skip-lint`. The code tries to minimize the work performed based on these settings.

I experimented with a `:var-definitions false` setting. It was easy to implement, and I have the code if you'd like it, but it doesn't affect performance much. The true performance gain comes from skipping the definition bodies, not from skipping their signatures.

- [x] This PR correponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code). See Slack discussion starting [here](https://clojurians.slack.com/archives/CHY97NXE2/p1651009819196579) and continuing [here](https://clojurians.slack.com/archives/C02RPGUCQNB/p1651060338202199).

### DIFF
--- a/src/clj_kondo/core.clj
+++ b/src/clj_kondo/core.clj
@@ -119,6 +119,7 @@
         files (atom 0)
         findings (atom [])
         analysis-cfg (get-in config [:output :analysis])
+        analyze-var-usages? (get analysis-cfg :var-usages true)
         analyze-locals? (get analysis-cfg :locals)
         analyze-keywords? (get analysis-cfg :keywords)
         analyze-protocol-impls? (get analysis-cfg :protocol-impls)
@@ -132,8 +133,8 @@
                             (not skip-lint))
                    (atom (cond-> {:namespace-definitions []
                                   :namespace-usages []
-                                  :var-definitions []
-                                  :var-usages []}
+                                  :var-definitions []}
+                           analyze-var-usages? (assoc :var-usages [])
                            analyze-locals? (assoc :locals []
                                                   :local-usages [])
                            analyze-keywords? (assoc :keywords [])
@@ -163,6 +164,7 @@
              :used-namespaces used-nss
              :ignores (atom {})
              :id-gen (when analyze-locals? (atom 0))
+             :analyze-var-usages? analyze-var-usages?
              :analyze-locals? analyze-locals?
              :analyze-protocol-impls? analyze-protocol-impls?
              :analyze-keywords? analyze-keywords?
--- a/src/clj_kondo/core.clj
+++ b/src/clj_kondo/core.clj
@@ -22,7 +22,7 @@
         output (cond-> {:findings findings}
                  (:summary output-cfg)
                  (assoc :summary summary)
-                 (or (:analysis config) (:analysis output-cfg))
+                 analysis
                  (assoc :analysis analysis))]
     (case fmt
       :text
@@ -125,8 +125,7 @@
         analyze-java-class-defs? (some-> analysis-cfg :java-class-definitions)
         analyze-java-class-usages? (some-> analysis-cfg :java-class-usages)
         analyze-meta? (or analysis-var-meta analysis-ns-meta)
-        analysis (when (and analysis-cfg
-                            (not skip-lint))
+        analysis (when analysis-cfg
                    (atom (cond-> {:namespace-definitions []
                                   :namespace-usages []
                                   :var-definitions []}
@@ -156,7 +155,6 @@
              :findings findings
              :namespaces (atom {})
              :analysis analysis
-             :output-analysis? (boolean analysis-cfg)
              :cache-dir cache-dir
              :used-namespaces used-nss
              :ignores (atom {})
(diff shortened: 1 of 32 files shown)


src/clj_kondo/impl/core.clj
```diff
@@ -321,7 +321,7 @@
 
 ;;;; threadpool
 
-(defn lint-task [ctx ^java.util.concurrent.LinkedBlockingDeque deque dev?]
+(defn analyze-task [ctx ^java.util.concurrent.LinkedBlockingDeque deque dev?]
```
mainej, line 324: This series of renames reflects the fact that these tools call `ana/analyze-input`, not one of the linter namespaces. They do, however, lint as a side effect of calling `ana/analyze-input`. Would `analyze-and-lint-task` be better?

borkdude: It's fine with me this way.

src/clj_kondo/core.clj
```diff
@@ -18,7 +18,12 @@
   subject to change."
   [{:keys [:config :findings :summary :analysis]}]
   (let [output-cfg (:output config)
-        fmt (or (:format output-cfg) :text)]
+        fmt (or (:format output-cfg) :text)
```
borkdude, line 21: It's unclear to me why this code was moved.

src/clj_kondo/core.clj
```diff
@@ -181,39 +179,42 @@
...
              :debug debug}
         lang (or lang :clj)
+        ;; primary file analysis and initial lint
         _ (core-impl/process-files (if parallel
                                      (assoc ctx :parallel parallel)
                                      ctx) lint lang filename)
         ;;_ (prn (some-> analysis deref :java-class-usages))
         ;; _ (prn :used-nss @used-nss)
-        idacs (when-not skip-lint
+        idacs (when (or dependencies (not skip-lint) analysis)
```
borkdude, line 188: Why does analysis need this step again?

src/clj_kondo/impl/analysis/java.clj
```diff
@@ -47,7 +47,7 @@
   (-> ctx :config ))
 
 (defn analyze-class-defs? [ctx]
-  (:analyze-java-class-defs? ctx))
+  (and (:analysis ctx) (:analyze-java-class-defs? ctx)))
```
borkdude, line 50: Why is this necessary? How can analyze-java-class-defs? be true while analysis is not true?

src/clj_kondo/impl/analysis/java.clj
```diff
@@ -67,7 +67,8 @@
                             (str jar ":" entry))}))))
 
 (defn analyze-class-usages? [ctx]
-  (and (:analyze-java-class-usages? ctx)
+  (and (:analysis ctx)
```
borkdude, line 70: Same

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1876,7 +1883,8 @@
                       expanded (assoc expanded :visited [resolved-namespace resolved-name])]
                 ;;;; This registers the original call when the new node does not
                 ;;;; refer to the same call, so we still get arity linting
-                  (when-not same-call?
+                  (when (and (:analyze-var-usages? ctx)
```
borkdude, line 1886: Same

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -172,20 +172,21 @@
...
-                                                                     :resolved-ns ns-name
-                                                                     :ns current-ns-name
-                                                                     :refer true
-                                                                     :lang lang
-                                                                     :base-lang base-lang
-                                                                     :filename filename
-                                                                     :config config
-                                                                     :expr %)))
-                                 opt-expr-children)
+                           (when (:analyze-var-usages? ctx)
```
borkdude, line 175: Same

src/clj_kondo/impl/analyzer/usages.clj
```diff
@@ -200,39 +200,40 @@
...
-                                                  :callstack (:callstack ctx)
-                                                  :config (:config ctx)
-                                                  :in-def (:in-def ctx)
-                                                  :context (:context ctx)
-                                                  :simple? simple?
-                                                  :interop? interop?
-                                                  ;; save some memory
-                                                  :expr (when-not dependencies expr)
-                                                  :resolved-core? resolved-core?})))))
+                       (when (:analyze-var-usages? ctx)
```
borkdude, line 203: Same

src/clj_kondo/core.clj
mainej: Sorry, I'll put this back... it was part of an intermediate refactoring where the logic for when to show the analysis got complicated and I didn't want to duplicate it. Now, the presence or absence of `analysis` is all that's needed so it'll be ok to duplicate.

src/clj_kondo/impl/analysis/java.clj
mainej: Same, will change.

src/clj_kondo/core.clj
borkdude: `l/lint-var-usages` is a bit of a misnomer. what actually happens there is resolving vars from the cached data and then linting which happens based on that retrieved data. So it could be called `post-analysis` / `post-cache` or something. It's just the naming that is confusing, but the logic makes sense (imo).

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1649,35 +1655,36 @@
...
-                                       :expr (when-not (:dependencies ctx) expr)
-                                       :simple? (simple-symbol? fsym)
-                                       :callstack (:callstack ctx)
-                                       :config (:config ctx)
-                                       :top-ns (:top-ns ctx)
-                                       ;; :arg-types (:arg-types ctx)
-                                       :interop? interop?
-                                       :resolved-core? resolved-core?
-                                       :in-def (:in-def ctx)}))
+            (when (:analyze-var-usages? ctx)
```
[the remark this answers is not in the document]
borkdude: Maybe we can make this more clearer because I'm pretty it will confuse _me_ in the future. What about `(not (:shallow-var-usages?))`?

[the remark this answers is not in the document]
borkdude: I'm trying of a better name, but trying to avoid `(:analyze-...? ctx)` since that kind of implies that `:analysis true` was used.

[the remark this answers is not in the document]
mainej: Maybe `(not (:skip-var-usage-analysis? ctx))` would be clearest? ("shallow" applies to var-defs, not var-usages.)

PR #1676, opened 2022-04-30
Add on-progress-update-fn callback

### LINKED ISSUE #1675: Support callback function for each file analyzed

As mentioned [here](https://clojurians.slack.com/archives/CHY97NXE2/p1651283825804139), it'd be super helpful if kondo could provide a callback function that would be called for each analyzed filename, so clojure-lsp could update progress bar smarter and precisely.

A custom callback function on clj-kondo JVM which would be called N times being N the number of filenames analyzed.
Note: if passing `src`, it'd be even better if kondo could call the callback for each file not only one time.

**Additional context**

I can open a PR, any suggestion on where to add that? maybe near existing custom-lint-fn, not sure all files are already available there to loop

ericdallo:
Fixes #1675 

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -2652,17 +2652,10 @@
                       filename ", "
                       (or (.getMessage ex) (str ex)))}])))
 
-(defn notify-progress [filename {:keys [on-progress-update-fn entries-to-call entries-call-count]}]
-  (when on-progress-update-fn
-    (when-let [entry (get @entries-to-call filename)]
-      (swap! entries-call-count update entry dec)
-      (when (= 0 (get @entries-call-count entry))
-        (on-progress-update-fn {:entry entry})))))
-
 (defn analyze-input
   "Analyzes input and returns analyzed defs, calls. Also invokes some
   linters and returns their findings."
-  [{:keys [:config] :as ctx} filename uri input lang dev?]
+  [{:keys [:config :file-analyzed-fn :files-count] :as ctx} filename uri input lang dev?]
   (when (:debug ctx)
     (utils/stderr "[clj-kondo] Linting file:" filename))
   (try
@@ -2711,7 +2704,10 @@
         (when (and (= :text (:format output-cfg))
                    (:progress output-cfg))
           (binding [*out* *err*]
-            (print ".") (flush)))))))
+            (print ".") (flush))))
+      (when file-analyzed-fn
+        (file-analyzed-fn {:filename filename
+                           :files-count @files-count})))))
 
 ;;;; Scratch
 
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -2652,10 +2652,17 @@
                       filename ", "
                       (or (.getMessage ex) (str ex)))}])))
 
+(defn notify-progress [filename {:keys [on-progress-update-fn entries-to-call entries-call-count]}]
+  (when on-progress-update-fn
+    (when-let [entry (get @entries-to-call filename)]
+      (swap! entries-call-count update entry dec)
+      (when (= 0 (get @entries-call-count entry))
+        (on-progress-update-fn {:entry entry})))))
+
 (defn analyze-input
   "Analyzes input and returns analyzed defs, calls. Also invokes some
   linters and returns their findings."
-  [{:keys [:config :file-analyzed-fn] :as ctx} filename uri total-files input lang dev?]
+  [{:keys [:config] :as ctx} filename uri input lang dev?]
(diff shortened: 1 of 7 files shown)


src/clj_kondo/impl/core.clj
```diff
@@ -408,21 +411,26 @@
...
                          :lang (if filename
                                  (lang-from-file filename default-language)
                                  default-language)} dev?)
           (classpath? path)
-          (run! #(process-file ctx % default-language canonical? filename)
-                (str/split path
-                           (re-pattern path-separator)))
+          (let [paths (str/split path
+                                 (re-pattern path-separator))]
+            (run! #(process-file ctx % default-language canonical? filename (count paths))
```
borkdude, line 432: It seems this count is supposed to be total-files, but it's actually the number of classpath entries here, so jars + directories.

ericdallo: yeah, indeed, any suggestion how to get all possible files here?

ericdallo:
Not sure why 3 tests are failing, still debugging it

src/clj_kondo/impl/core.clj
ericdallo: Also, I'm ok on we don't have all specific files, but something roughly, a different name would be better though if that was the case

borkdude: In parallel mode you don't know the number of files before hand. So I don't think we can reliably add this number at all.

ericdallo: RIght, I think I can try to make that on lsp side, probably will be easier and enough for what we want to fix: a better feedback, it doesn't need to be totally precise. I let you know if that works and then change this PR

ericdallo:
@borkdude It took me some time but it works as expected now, I did some changes and spent some time making the parallel work, also, I made the feature a little more specific, now it's called for each entry passed on :lint, not for every file as we can't know when :parallel as you said, this is enough for clojure-lsp have a way better progress report. LMK what you think

borkdude:
@ericdallo Can you explain in words to me how it now works and how this benefits clojure-lsp and potentially other projects?

ericdallo:
Sure, the idea is to have a callback function related to the progress of the analysis, the first idea was to call that for each filename but since this is not possible to know the total times that function would be called, it would make hard to clojure-lsp or any client create some kind of progress bar without knowing how many times the function would be called.

The way it works now is: it calls that callback for each entry passed which could be a file or a folder, check the tests where those cases are tested. The parallel mode made a little bit more trick to make it work, but in the end the logic is the same

ericdallo:
Alright @borkdude, made one more change, now code is way easier since we count total files before when that callback is passed, and the callback is `file-analyzed-fn` which will be triggered for every filename analyzed

borkdude:
I'll look again when the tests are fixed :)

ericdallo:
Ops, my bad, fixed @borkdude !

src/clj_kondo/impl/core.clj
```diff
@@ -361,6 +361,47 @@
 (defn classpath? [f]
   (str/includes? f path-separator))
 
+(defn ^:private entries-from-jar [^java.io.File jar-file]
```
borkdude, line 364: I think laziness should never escape `with-open`. Also, perhaps we can just call this `entries-from-jar-count`, and already count in there before closing?

src/clj_kondo/impl/core.clj
```diff
@@ -361,6 +361,47 @@
...
+              (let [path (str path)
+                    file (io/file path)
+                    canonical (when (.exists file)
+                                (.getCanonicalPath file))]
+                (cond
+                  (and canonical
+                       (str/ends-with? canonical ".jar"))
+                  (count (entries-from-jar file))
+
+                  (and canonical
```
borkdude, line 384: The canonical stuff doesn't really matter for counting files I think?

ericdallo: yes, this is just the path, maybe we could rename from canonical to other name?

borkdude: ok

borkdude:
Also Windows tests are failing.

borkdude:
There is some normalize-path function in the test-utils to fix that.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -2707,7 +2707,7 @@
             (print ".") (flush))))
       (when file-analyzed-fn
         (file-analyzed-fn {:filename filename
```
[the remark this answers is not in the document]
ericdallo: I think we can assoc both for now?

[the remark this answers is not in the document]
borkdude: OK

PR #1679, opened 2022-05-02
Add :files-done to file-analyzed-fn

ericdallo:
This adds `:files-done` to the `file-analyzed-fn`, this is useful because usually clients want to know how many files we already analyzed, not only the total.

@bokdude I know we already have the `files` atom, but on parallel it's not reliable for this feature as it will swap all files before starts analyzing it, it'd result in a wrong number when we trigger the file-analyzed-fn call.

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -2656,7 +2656,7 @@
 (defn analyze-input
   "Analyzes input and returns analyzed defs, calls. Also invokes some
   linters and returns their findings."
-  [{:keys [:config :file-analyzed-fn :total-files] :as ctx} filename uri input lang dev?]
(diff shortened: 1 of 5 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -2707,11 +2707,11 @@
           (binding [*out* *err*]
             (print ".") (flush))))
       (when file-analyzed-fn
-        (swap! files-analyzed-done inc)
+        (swap! files inc)
```
borkdude, line 2710: The files atom needs to be incremented regardless whether we're having a file-analyzed-fn

ericdallo: good catch!

ericdallo:
It should be ready to review @borkdude, everything working with clojure-lsp

PR #1692, opened 2022-05-15
Implement :redundant-call

NoahTheDuke:
As discussed on Slack, here's an initial version of `:redundant-call`, which lints against function/macro calls that return the given argument (generally a single-arity form). For the sake of ease and clarity, I didn't expand this to `as->`, but doing so would be trivial. I also didn't do any research on other functions this could apply to, only those that readily came to mind.


borkdude:
Feedback:

- Implementing the check in the `macroexpand` namespace seems bit out of place
- I don't think you have to base the check on the rewrite-clj node since we already have the number of children from the call available in `lint-usages`.

borkdude:
Looks great!

PR #1736, opened 2022-06-25
Add `nilable` as an option to map type specs.

### LINKED ISSUE #1735: Type definitions for maps are not nilable

A map spec typically looks like this:

```clojure
{:op :keys
 :req {:a-number :int}}
```
This spec is very helpful in Cursive, where you will get warnings if you try to pass in a map with the wrong type for `:a-number`. The problem is that there is currently no way to state that this map could be `nil`. That leaves us with `:any` (I think) which is way less useful.

[In this slack discussion](https://clojurians.slack.com/archives/CHY97NXE2/p1655894939666259), @borkdude suggested using the following format to make maps nilable:

```clojure
{:op :keys 
 :nilable true}
```

I could try to tackle this myself, a few pointers to relevant places in code would be helpful.

ingesolvoll:
Fixes #1735

The simplest solution that I could figure out right away. The only obvious weakness would be performance, there may be a way to run the nil checking code less often?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 ## Unreleased
 
 - [#698](https://github.com/clj-kondo/clj-kondo/issues/698): output rule name with new output option `:show-rule-name-in-message true`. See example in [config guide](https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#show-rule-name-in-message).
+- [#1735](https://github.com/clj-kondo/clj-kondo/issues/1735) Add support for nilable map type specs
 
 ## 2022.06.22
 
--- a/src/clj_kondo/impl/types.clj
+++ b/src/clj_kondo/impl/types.clj
@@ -380,7 +380,9 @@
         (emit-missing-required-key! ctx arg k)))))
 
 (defn lint-map! [ctx s a t]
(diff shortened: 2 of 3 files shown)


borkdude:
Thanks

ikitommi:
Could this be extended to all types? Can all types be presented as map with `:op`?

borkdude:
@ikitommi Simple types can be represented as `:nilable/string` and you can also use `#{:nilable/string :nilable/keyword}` for combinations of types.

ikitommi:
That I knew, but if any type could be presented also as `:op`-map, you would have unified AST with nilable for all types. E.g. `{:op :string, :nilable true}`,  `#{{:op :string, :nilable true}, {:op :keyword, :nilable true}}`. The `:nilable/string` would be just sugar on top of this. 

Have already mapped the `:nilable/xyz` already on malli side, this new this is only applied for `:keys` op. So, just thinking aloud 🙂

ingesolvoll:
> if any type could be presented also as `:op`-map, you would have unified AST with nilable for all types.

@ikitommi I thought about that while [doing this](https://github.com/metosin/malli/pull/715/files#diff-0c47d0cd013f6cb845efdab593e740088b1cbcdb0704959422641cdffd9b7b01R100) . Seems like all other `:op` types would have the same problem.

Issue #1782, opened 2022-08-21
clj-kondo might throw an InvalidPathException exception when analysing jar files on MS-Windows


ikappaki:
**version**

v2022.08.03

**platform**

MS-Windows, clj-kondo native.

**problem**

clj-kondo throws an invalid filename exception when performing parallel analysis on a jar file with a file-analyzed-fn and canonical paths set to true, e.g.:

``` clojure
#error {
 :cause "Illegal char <:> at index 60: C:\\Users\\ikappaki\\.m2\\repository\\babashka\\fs\\0.1.6\\fs-0.1.6.jar:META-INF\\leiningen\\babashka\\fs\\project.clj"
 :via
 [{:type java.nio.file.InvalidPathException
   :message "Illegal char <:> at index 60: C:\\Users\\ikappaki\\.m2\\repository\\babashka\\fs\\0.1.6\\fs-0.1.6.jar:META-INF\\leiningen\\babashka\\fs\\project.clj"
   :at [sun.nio.fs.WindowsPathParser normalize "WindowsPathParser.java" 182]}]
 :trace
 [[sun.nio.fs.WindowsPathParser normalize "WindowsPathParser.java" 182]
  [sun.nio.fs.WindowsPathParser parse "WindowsPathParser.java" 153]
  [sun.nio.fs.WindowsPathParser parse "WindowsPathParser.java" 77]
  [sun.nio.fs.WindowsPath parse "WindowsPath.java" 92]
  [sun.nio.fs.WindowsFileSystem getPath "WindowsFileSystem.java" 232]
  [java.io.File toPath "File.java" 2387]
  [babashka.fs$as_path invokeStatic "fs.cljc" 35]
  [babashka.fs$as_path invoke "fs.cljc" 32]
  [babashka.fs$exists_QMARK_ invokeStatic "fs.cljc" 124]
  [babashka.fs$exists_QMARK_ invoke "fs.cljc" 119]
  [babashka.fs$exists_QMARK_ invokeStatic "fs.cljc" 121]
  [babashka.fs$exists_QMARK_ invoke "fs.cljc" 119]
  [clj_kondo.impl.utils$__GT_uri invokeStatic "NO_SOURCE_FILE" 378]
  [clj_kondo.impl.utils$__GT_uri invoke "NO_SOURCE_FILE" 377]
  [clj_kondo.impl.analyzer$analyze_input invokeStatic "analyzer.clj" 2800]
[...]

PR #1789, opened 2022-08-30
Pass existing filename URI to file-analyzed-fn

ikappaki:
(cc'ing @ericdallo)

Hi,

could you please consider a patch to simplify the logic around what to pass as the filename URI to the `file-analyzed-fn` as discussed in #1782 starting with https://github.com/clj-kondo/clj-kondo/issues/1782#issuecomment-1228920239

The analyzer, without this patch, will pass a nil URI to the `file-analyzed-fn` when there is a request for a nested jar file. As discussed, It is better instead to pass in the already calculated input `uri` argument rather than construct on the fly from the `filename` which might leave the URI as nil.

As also discussed in the above thread, and  is now obvious from the additional test vector, the analyzer will return "unconventional" URIs for the nested jar files, such as `"jar:file:file:/.*/corpus/withcljdir.jar!/dirinjar.clj/arity.clj"` (double "file" in the URI scheme).

This can also be seen using the public API, e.g.:

``` clojure
(clj-kondo.core/run! {:parallel true :lint ["corpus/withcljdir.jar"] :config {:output {:canonical-paths true} } :file-analyzed-fn #(prn :analysis-results %)})
;; => :analysis-results {:filename "/home/ikappaki/src/clj-kondo/corpus/withcljdir.jar:dirinjar.clj/arity.clj", :uri "jar:file:file:/home/ikappaki/src/clj-kondo/corpus/withcljdir.jar!/dirinjar.clj/arity.clj", :total-files 1, :files-done 1}
;; => ...
```

I have also have a patch at hand to "fix" the above, i.e. remove the repeated "file" uri scheme. Let me know if you think this can be accommodated in this patch.

Will update the changelog in due time.

Thanks 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+  [#1782](https://github.com/clj-kondo/clj-kondo/issues/1782): Fix issue with jar URI missreporting to `file-analyzed-fn`, bump babashka/fs to 0.1.11
 - [#1780](https://github.com/clj-kondo/clj-kondo/issues/1780): Can not use NPM dependency namespaces beginning with "@" in consistent-linter alias
 - [#1771](https://github.com/clj-kondo/clj-kondo/issues/1771): don't crash on empty ns clauses: `(require '[])` and `(import '())` 
 - [#1774](https://github.com/clj-kondo/clj-kondo/issues/1774): Add support for sourcehut inferred git dep urls for the Clojure CLI
--- a/deps.edn
+++ b/deps.edn
@@ -4,7 +4,7 @@
         cheshire/cheshire {:mvn/version "5.11.0"}
         nrepl/bencode {:mvn/version "1.1.0"}
         org.babashka/sci {:mvn/version "0.3.2"}
-        babashka/fs {:mvn/version "0.1.2"}
+        babashka/fs {:mvn/version "0.1.11"}
         org.ow2.asm/asm {:mvn/version "9.2"}}
  :aliases {:clj-kondo/dev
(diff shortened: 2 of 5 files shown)


ericdallo (approved):
LGTM

borkdude:
@ikappaki OK!

ikappaki:
Thanks both, shall I also go ahead with fixing the double `file` in the jar uri scheme as mentioned here https://github.com/clj-kondo/clj-kondo/issues/1782#issuecomment-1232124264 in this PR, given it is related as per the new test results?

borkdude:
Sure!

borkdude:
Seems all good now. Did you add everything to the changelog that you wanted to add?

ikappaki:
> Seems all good now. Did you add everything to the changelog that you wanted to add?

Yes.

Just to summarize, this PR includes the following fixes as discussed on the parent issue:
1. Fixes issue with babshka/fs throwing an exception on MS-Windows when path includes stray `:`s (https://github.com/babashka/fs/pull/72); `babashka/fs` bumped to 0.1.11.
2. The URI passed to `file-analyzed-fn` is taken directly from the input variable with the same name; this is more reliable than trying to determine it from the file name.
3. Eliminates the double `file` scheme in jar file URIs.

The changelog entry is complete.

Though I've noticed that there is another babashka/fs dependency in `project.clj` that probably needs to be bumped to 0.1.11, but the `script/update-project.clj` file mentioned in the header is nowhere to be found?

``` clojure
;; GENERATED by script/update-project.clj, DO NOT EDIT
;; To change dependencies, update deps.edn and run script/update-project.clj.
;; To change other things, edit project.template.clj and run script/update-project.clj.
```

borkdude:
I'll take care of the fs bump after this PR.

PR #1793, opened 2022-09-04
Add support for filename-pattern on ns-group

### LINKED ISSUE #1790: Support `:file-pattern` on `:ns-group`
At Nubank, we want to use the `discouraged-var` linter only on test namespaces, the issue is that we can know for sure a ns is in a test source-path only checking the `ns`, we need to check the source-path.

if we support a `:file-pattern` on the `:ns-group` we would be able to exclude specific source-paths.

no alternatives AFAICS

**Additional context**
https://clojurians.slack.com/archives/CHY97NXE2/p1661880291392359

I can help with a PR

ericdallo:
Fixes #1790


borkdude:
Looks good, I wonder if this works on Windows too using the same regex.

ericdallo:
@borkdude good catch, maybe I can simplify the regex

borkdude:
Isn't that just hiding the problem? I think those regexes should work for *nix and Windows, else you would have to write some pretty hairy regexes to target both. We already have similar regexes for `:output {:exclude-file ...}`.

borkdude:
Hmm, since we don't have support for this in `:exclude-files` either, we will address that when necessary. So this PR is sufficient imo.

PR #1805, opened 2022-09-11
Colocate linter config defaults and documentation

NoahTheDuke:

---

As discussed briefly on Clojurian Slack, here's my initial attempt at putting both linter config and documentation into the same place and then using it in both `clj-kondo` and the docs.

Open questions:

* What other fields are worth tracking? Should I include "Version added" to each linter?
* I wrote a fake "spec" for each of the linters with configs, as the descriptions were sometimes brief and inspecific. `clj-kondo` doesn't use real spec or other schema system, so they're only good for human understanding. Is their inclusion worth it?
* In this form, "example trigger" and "example message" feel like they're not nearly enough documentation for what is or is not expected. Should these be expanded to look closer to Rubocop or eslint-style lists of "Good" and "Bad" examples? (Ideally, examples would be next to the lints themselves, but the lints aren't built like that in `clj-kondo`, so putting them here will work enough.)

I personally think this is a good way go to in general. Lack of indentation-aware multi-line strings in Clojure is a real pain, but otherwise, I think this demonstrates that the concept is sound and workable.


NoahTheDuke:
Closing as this is a solution in search of a problem.

PR #1808, opened 2022-09-13
Add :unknown-require-option

### LINKED ISSUE #1460: Add lint for incorrect :require forms
I had a typo in a require form that was pretty confusing, I had types `:s` rather than `:as`. This was hard to spot in a file with many imports.

```Clojure
(ns circle.http.api.private.test-user
  (:require [ring.middleware.params :s params]))
```

When I attempted to use a var from the namespace, I got an error on the line where I used it:

`Unresolved namespace params. Are you missing a require?`

but when I saw this message, I look up at the requires, and it all looked good - I _was_ requiring `ring.middleware.params`. I was pretty confused.

I would love to see an error message pointing at the `:s` symbol, saying that `:s` is not a valid require directive.

NoahTheDuke:
Warns on unknown `:require` option pairs. Includes no options as I can't think of any reason to include that at this time.

Closes #1460

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+- [#1460](https://github.com/clj-kondo/clj-kondo/issues/1460): Add a linter for incorrect `:require` options
 - [#1794](https://github.com/clj-kondo/clj-kondo/issues/1794): Add a linter for line length
 
 ## 2022.09.08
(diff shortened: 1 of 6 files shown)


borkdude:
@NoahTheDuke 

- Let's call this: ` :unknown-require-option`
- Let's start the message with a capital letter: `Unknown require option: :foo`

borkdude:
Otherwise, looks good!

borkdude:
Two more things:

- The default level should be `:warning`: an unrecognized option doesn't crash your program. It might cause other errors, but those will be reported elsewhere as errors.
- A test that shows that this warning can be ignored with `#_:clj-kondo/ignore` in case we (clj-kondo) made a mistake

borkdude:
Can you change the message to just: `Unknown require option: :s`. We don't use single quotes anywhere in other messages and I don't think printing the value is important.

PR #1809, opened 2022-09-14
Add :existing-alias lint

### LINKED ISSUE #1800: Lint: warn when using fully-qualified keyword or symbol when there is an alias defined for the same namespace

```clojure
(ns foo
  (:require
    [clojure.string :as str]
    [bar :as-alias b]))

;; show  `alias available` warning for both
(clojure.string/blank? "")
(identity :bar/baz)
```

If there is already an alias defined for a namespace, show a linter warning when using fully qualified keywords/symbols of that same namespace. 

The rationale is that such warnings will help keep code more consistent, easier to grep and refactor, etc.

NoahTheDuke:

---

~Pretty simple implementation~. Includes config to exclude namespaces, as I suspect some folks prefer to use the fully-qualified version of one or another namespace while checking others.

I chose to check symbols before performing `:resolve-as` or any `:hooks` expansion, as it seemed to me that we'd want to allow users to enforce namespace aliases even when they say "resolve this as `def-catch-all`". For keywords, I checked before registering that the namespace is used. I don't think this should affect anything, but it seems worth noting.

Closes #1800

EDIT: This has grown quite a bit, my apologies. It started small as I wrote above, but to handle clj-kondo's macroexpansion system, I had to make some deeper changes.

First, I replaced all uses of `token-node` with a new function `generated-token` which acts like `hooks-api/mark-generate`: return a new `token-node` for a given symbol with attached `:clj-kondo.impl/generated true` meta. I did this because clj-kondo's macroexpansion system creates a lot of fully-qualified symbols in place (`cond->`, etc) which are picked up by the new lint, leading to strange warnings with no line/col numbers. I considered doing this for all of the re-exported rewrite-clj functions but decided to hold off until I got some feedback for this approach.

Second, because `(:aliases (:ns ctx))` only goes one direction (alias -> namespace), I had to flip it directly in the linter implementation. I would have preferred to do this wherever `:aliases` is constructed (`:aliases` and `:ns->aliases` maybe), but I couldn't find it and gave up to demonstrate validity of the overall approach. I'd be happy to move it tho, this is fast enough but repeats a lot of work.


corpus/existing_aliases/as_alias_example.clj
```diff
@@ -0,0 +1,4 @@
+(ns as-alias-example
+  (:require [baz.qux :as-alias q]))
+
+(identity ::baz.qux/some-fn)
```
borkdude, line 4: This example isn't runnable in Clojure so maybe it's a different category than this error.

borkdude: Github is slow and doesn't let me edit my message. This error -> this linter.

borkdude:
> EDIT: This has grown quite a bit

I _think_ the implementation can be simplified greatly without touching any of the hook stuff and also without putting anything in `linters.clj`. (Disclaimer: I'm not 100% sure since I haven't implemented this myself yet).

The relevant bit is where a symbol or keyword is analyzed in `src/clj_kondo/impl/namespace.clj`'s `resolve-name`.  The namespace will be recognized as either: an already required namespace, an aliased namespace or an unresolved namespace. In the first case we can simply check if there's also an alias for that namespace and then emit the message, that's it. For keywords, I'm not 100% sure where we do this, I think in `impl/usages.clj` but the logic is similar.

corpus/existing_aliases/as_alias_example.clj
NoahTheDuke: Do you mean Clojure 1.11? I didn't think about that. Could we check the clojure version and only warn if 1.11 or higher?

borkdude: It also doesn't work in any other clojure version if you just use `:as`. You cannot use auto-resolved keywords with anything else than an alias.

NoahTheDuke: Oh, I didn't know that.

Selfishly, I'd really like this one cuz I work on a Clojure-only codebase. Can we limit it to  just Clojure? Or would you prefer to have a separate related linter that only works on Clojure?

NoahTheDuke:
I'll try it out.

> In the first case we can simply check if there's also an alias for that namespace and then emit the message, that's it.

To be clear, this function doesn't check the result of macro expansions? That's the issue I ran into: clj-kondo performs its own version of macro expansion.

For example, `cond->` is (roughly) transformed into a bunch of `(if ~test (clojure.core/-> ~g ~step) ~g)` forms. This normally isn't an issue, but it is if the namespace declaration includes `(:require [clojure.core :as cc])`. Without some way of tracking that clj-kondo has inserted the `clojure.core/->` token, how does clj-kondo know?

NoahTheDuke:
Okay, your suggestion simplified things as anticipated but without some notion of "this form is created by clj-kondo", I don't know how to distinguish between forms from the given text and forms created by clj-kondo.

Any ideas?

NoahTheDuke:
I know all of the tests pass, but if you look in the logs, it's printing errors:

```
test/clj_kondo/impl/analyzer_test.clj::: warning: An alias is defined for clojure.test: t
test/clj_kondo/impl/parser_test.clj::: warning: An alias is defined for clojure.test: t
```

corpus/existing_aliases/as_alias_example.clj
NoahTheDuke: Rereading this, seems I have fundamentally misunderstood the issue you raised. I've removed the whole test and removed all mention of keywords from the changelog and docs.

borkdude:
@NoahTheDuke You can assume a node was generated if:

1) It has no metadata
or:
2) `:clj-kondo.impl/generated` on the node is `true`

NoahTheDuke:
Something odd is happening with metadata, which wasn't *entirely) fixed by #1813.

Using `:analyze-call`, the `expr` passed to `resolve-name` has no metadata so this passes:

```clojure
(deftest dont-check-analyze-call-hook
  (is (empty? (lint!
                (str "(ns analyze-call-hook (:require [clojure.core :as cc]))\n"
                     "(defn new-> [x f] (f x))\n"
                     "(new-> 1 inc)")
                {:linters ^:replace {:existing-alias {:level :warning}}
                 :hooks
                 {:__dangerously-allow-string-hooks__ true
                  :analyze-call
                  {'analyze-call-hook/new->
                   (str "(require '[clj-kondo.hooks-api :as api])\n"
                        "(defn new-> [{:keys [node]}]\n"
                        "  (let [children (rest (:children node))\n"
                        "        node (list* (api/token-node 'clojure.core/->) children)]\n"
                        "    {:node (api/list-node node)}))")}}}))))
```

and prints this malformed report:

```
<stdin>::: warning: An alias is defined for clojure.core: cc
```

Using `macroexpand`, the `expr` passed to `resolve-name` has metadata so this fails:

```clojure
(deftest dont-check-macroexpansion
  (is (empty? (lint!
                (str "(ns macroexpansion (:require [clojure.core :as cc]))\n"
                     "(defn new-> [x f] (f x))\n"
                     "(new-> 1 inc)")
                {:linters ^:replace {:existing-alias {:level :warning}}
                 :hooks
                 {:__dangerously-allow-string-hooks__ true
                  :macroexpand
                  {'macroexpansion/new->
                   (str "(require '[clj-kondo.hooks-api :as api])\n"
                        "(defmacro new-> [x f]\n"
                        "  (list 'clojure.core/-> x f))")}}}))))

```

and prints this correct report:

```clojure
FAIL in (dont-check-macroexpansion) (NO_SOURCE_FILE:50)
expected: (empty? (lint! (str "(ns macroexpansion (:require [clojure.core :as cc]))\n" "(defn new-> [x f] (f x))\n" "(new-> 1 inc)") {:linters {:existing-alias {:level :warning}}, :hooks {:__dangerously-allow-string-hooks__ true, :macroexpand {(quote macroexpansion/new->) (str "(require '[clj-kondo.hooks-api :as api])\n" "(defmacro new-> [x f] (list 'clojure.core/-> x f))")}}}))
  actual: (not (empty? ({:file "<stdin>", :row 3, :col 1, :level :warning, :message "An alias is defined for clojure.core: cc"})))
```

Any other ideas?

borkdude:
In the first case: not all nodes have metadata, this just how it is, then we can't report. In the second case, you should check `:clj-kondo.impl/generated` on the node and then ignore that one.

NoahTheDuke:
Ignoring nodes with no meta works for me. In the second case, they don't have `:clj-kondo.impl/generated` 😅. Maybe this is a deeper bug?

borkdude:
@NoahTheDuke The ignore is on the symbol. In the case of a call, the `expr` is a list node. So something like this should work:

``` clojure
(defn ignored? [expr]
  (or (:clj-kondo.impl/generated expr)
      (some-> expr :children first :clj-kondo.impl/generated)))
```

``` clojure
  (when-not (or (contains? (:aliases (:ns ctx)) ns-sym)
                (linter-disabled? ctx :existing-alias)
                (config/existing-alias-excluded? (:config ctx) ns-sym)
                (ignored? expr))
```

borkdude:
I think the map inversion (namespace -> alias) could be done once when analyzing the namespace form, _iff_ the linter is enabled.

borkdude:
@NoahTheDuke In case you didn't notice:

https://github.com/clj-kondo/clj-kondo/pull/1809#issuecomment-1249344129

In case you did notice: consider this message not posted :)

NoahTheDuke:
My apologies, I don't think I follow.

I'm only generating the inverse map after checking if the linter level is not :off. That's what you asked for, right?

borkdude:
@NoahTheDuke Indeed. I think the inverse map could be calculated only once instead of every time encountering a symbol that should be checked, namely in the same spot where we store `:aliases` in the context.

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -340,6 +340,18 @@
                          {ns-name ns-name}
                          analyzed)
      :aliases (into {} (comp (filter :as) (map (juxt :as :ns))) analyzed)
+     :ns->aliases (when-not (-> ctx :config :linters :existing-alias :level
```
NoahTheDuke, line 343: @borkdude I moved it here.

src/clj_kondo/impl/namespace.clj
```diff
@@ -464,6 +464,45 @@
...
+  ;; Reasons to not check the given symbol:
+  ;; * it's already using an existing alias
+  ;; * this lint is disabled
+  ;; * this lint is configured to exclude the given namespace
+  ;; * it's generated by clj-kondo
+  (when-not (or (contains? (:aliases (:ns ctx)) ns-sym)
+                (linter-disabled? ctx :existing-alias)
+                (config/existing-alias-excluded? (:config ctx) ns-sym)
+                (generated? expr))
+    (when-let [ns->aliases (:ns->aliases (:ns ctx))]
```
NoahTheDuke, line 489: @borkdude from here

borkdude: Ah right, I only read the commit message, not the actual commit ;).

borkdude:
OK, I'm happy now with the code.

There are three things I want to think about more:

- The default level should be `:off` at least for a while, so we can test drive this linter for a few months, before we enable it. I usually do this to prevent other people's builds from failing because of bugs in clj-kondo.
- The name `:existing-alias` could maybe be improved. `:aliased-namespace` perhaps? For a fully written namespace that has an alias? 
- The original issue also references keywords like `:clojure.string/foo` vs `::str/foo`. Should we have two separate linters for this, or a config option? If we keep it split, should the name of this linter reflect this? `:aliased-namespace-symbol` and `:aliased-namespace-keyword`? Let's not support keywords as part of the current PR, to keep things focussed.

NoahTheDuke:
> The default level should be :off at least for a while

Cool, that works for me.

> `:aliased-namespace` perhaps?

I'm fine with that. I don't have strong feelings about this part.

> Should we have two separate linters for this, or a config option?

I think two separate linters is best here. Folks can set levels and exclusions etc for each. As for the name, I think including the type is clearest. That way, docs and examples and such will all co-exist.

NoahTheDuke:
I think the keyword part of this will be easy comparatively. Should be a single call in `clj-kondo.impl.analyzer.usages/analyze-keyword` to `lint-aliased-namespaces` when the keyword has a namespace but isn't `:namespaced?`, and adding `:as-alias` aliases to the `:ns->aliases` map.

borkdude:
Nonetheless, let's do a separate PR, we've had enough back and forth for one commit :)

borkdude:
I'll bikeshed a little bit more about the name in the clj-kondo channel, see if someone can improve upon what we came up with.

PR #1814, opened 2022-09-16
Check overrides during var-definition analysis

### LINKED ISSUE #1758: Overrides don't get applied to var-definitions during analysis
Version: v2022.06.22 

The [overrides](https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/impl/overrides.clj#L22-L25) are being applied to var usages but not var definitions during analysis. My understanding is that it works for usages because the call `(l/lint-var-usage ctx (override idacs))` ends up calling `analysis/reg-usage!` with the overrides applied, whereas `analysis/reg-var!` only receives the `analysis` map.

https://github.com/clj-kondo/clj-kondo/blob/master/src/clj_kondo/core.clj#L199-L208

You can see the inconsistency by comparing the definition and usage of `defn-`, which is defined in `overrides.clj` with a `:varargs-min-arity 2`:

```clojure
;; definition
{:ns clojure.core
 :name defn-
 :doc "same as defn, yielding non-public def"
 :varargs-min-arity 1
 :arglist-strs ["[name & decls]"]}

;; usage
{:name defn-,
 :filename "/tmp/foo.clj",
 :from foo,
 :macro true,
 :col 2,
 :arity 3,
 :varargs-min-arity 2,
 :row 8,
 :to clojure.core}
```

You can also reproduce this in your own namespace as long as you add the entry to `overrides.clj`:

```clojure
(ns foo)

(defn- bar [x & y])

(bar :_ :_)
```

```clojure
;; overrides.clj
(assoc-in '[:clj :defs foo bar] '{:ns foo
                                  :name bar
                                  :varargs-min-arity 2})
```

```clojure
{:var-definitions
 [{:end-row 3,
   :name-end-col 11,
   :name-end-row 3,
   :private true,
   :name-row 3,
   :ns foo,
   :name bar,
   :defined-by clojure.core/defn-,
   :col 1,
   :name-col 8,
   :end-col 20,
   :arglist-strs ["[x & y]"],
   :varargs-min-arity 1,
   :row 3}],
 :var-usages
 [{:end-row 3,
   :name-end-col 7,
   :name-end-row 3,
   :name-row 3,
   :name defn-,
   :from foo,
   :macro true,
   :col 1,
   :name-col 2,
   :end-col 20,
   :arity 2,
   :varargs-min-arity 2,
   :row 3,
   :to clojure.core}
  {:end-row 5,
   :name-end-col 5,
   :name-end-row 5,
   :name-row 5,
   :name bar,
   :from foo,
   :col 1,
   :name-col 2,
   :end-col 12,
   :arity 2,
   :varargs-min-arity 2,
   :row 5,
   :to foo}]}
```

My expectation is that overrides get applied to both definitions in a consistent manner, such that the records can be joined without conflict.

I took a little while to try to fix it, but I'm afraid I'm not familiar enough with the clj-kondo code to make a change of this size. There are some comments in place already about making improvements around the handling of overrides and about changing the structure to have the linter called from analysis rather than the other way around as it is now. I hope that by filing this issue the task of fixing it is made easier for someone more familiar with kondo than I am.

sheluchin:
Fixes #1758.


borkdude:
@sheluchin Is this PR done?

borkdude:
I don't know why CI is acting up, but it doesn't seem to be related to your PR.

sheluchin:
@borkdude Yes, I believe it is done. Would be nice if you could do a quick sanity check on the code, but my manual tests and unit tests all pass.

No idea what's happening with CI there. Happened with my last PR as well, but I too don't think anything in my changes could be responsible.

borkdude:
It does seem to be unique to your account. Do you maybe have a CircleCI account attached to your Github account or so?

borkdude:
Nope, it seems it's a circleci issue. I needed to re-login.

borkdude:
Thanks!

PR #1822, opened 2022-09-29
Include vectors in :exclude-destructured-as

### LINKED ISSUE #1821: :exclude-destructure-as doesn't check vectors

The linter `:unused-binding` checks for all bindings in destructured forms. This also checks the `:as X` portion. That check can be disabled with `:exclude-destructured-as` to allow for documentation purposes. However, the flag doesn't apply to vector destructuring forms that use `:as X`.

The flag `:exclude-destructured-as` applies to vector destructuring as well.

Do nothing, let `:exclude-destructured-as` only apply to maps.

**Additional context**
N/A

I have a fix ready, if desired.

NoahTheDuke:
Fixes #1821

---

Relatively simple change:
1. Check if the vector has an `:as` binding and if the `:exclude-destructured-as` is set.
2. If so, store the vector's last child as `as-sym`.
3. When looping over the vector's children, if the expr matches the `as-sym`, mark binding as used on context (same as how map destructuring works).

Closes #1821

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,15 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 - [#1758](https://github.com/clj-kondo/clj-kondo/issues/1758): Overrides don't get applied to var-definitions during analysis
 - [#1794](https://github.com/clj-kondo/clj-kondo/issues/1794): Add a linter for line length
-- [#1460](https://github.com/clj-kondo/clj-kondo/issues/1460): Add a linter for unknown `:require` options
+- [#1460](https://github.com/clj-kondo/clj-kondo/issues/1460): Add a linter for unknown `:require` options ([@noahtheduke](https://github.com/noahtheduke))
 - [#1807](https://github.com/clj-kondo/clj-kondo/issues/1807): false positive with map transducer in cljs
 - [#1806](https://github.com/clj-kondo/clj-kondo/issues/1806): false positive recur mismatch with letfn
 - [#1810](https://github.com/clj-kondo/clj-kondo/issues/1810): Fix printing error map as additional error
 - [#1812](https://github.com/clj-kondo/clj-kondo/issues/1812): Inconsistent handling of location metadata sometimes produces nil values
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#1800](https://github.com/clj-kondo/clj-kondo/issues/1800): Add `:aliased-namespace-symbol`, a linter that checks qualified symbols to see if they're using an existing alias.
 - [#1805](https://github.com/clj-kondo/clj-kondo/issues/1805): Ignore hints not being considered on protocol vars
 - [#1819](https://github.com/clj-kondo/clj-kondo/issues/1819): Fix "Too many open files" in java class definition analysis caused by files not being closed
+- [#1821](https://github.com/clj-kondo/clj-kondo/issues/1821): Include vectors in `:unused-binding` config `:exclude-destructured-as` flag.
```
borkdude, line 17: Feel free to add your github name like in previous changelogs

src/clj_kondo/impl/analyzer.clj
```diff
@@ -240,15 +240,27 @@
                         (str "unsupported binding form " expr))))
          :vector (let [children (:children expr)
                        all-tokens? (every? #(identical? :token %) (map :tag children))
+                       as-binding? (identical? :as (some-> (butlast children)
+                                                           last
+                                                           :k
+                                                           name
```
borkdude, line 246: I haven't tried locally yet, but does this call `name` on a rewrite-clj node? That seems a bit weird to me.

Is there a more efficient way to get to the but-last element of a seq without walking it almost twice? Maybe not, but maybe we could prevent doing that if we know the linter isn't enabled (which is checked after this, so should be moved up)

NoahTheDuke: I borrowed code from another part of the app. As far as I can tell, calling `(keyword (name (:k some-node)))` is unnecessary word. For keywords, the value of `:k` will be a keyword, and for all else, it will be `nil`.

You're right that I can do this slightly more efficiently. I've pushed a new version that first checks if the flag is sent, and if so, uses `drop` to get the last two items, and if the second-to-last item is `:as`, return the last item `as-sym`.

Thanks for the idea, glad to get it simpler and faster.

PR #1834, opened 2022-10-06
Fix :include-macros error in cljc and cljs

### LINKED ISSUE #1830: Ignore `:include-macros` option in the unknown require options linter
**version**

v2022.10.05

**platform**

Native & JVM

**problem**

The `:unknown-require-option` linter does not allow for the clojureScript `:include-macros` [optional argument](https://clojurescript.org/guides/ns-forms)

**repro**

```
$ echo '(ns user (:require [other.namespace :include-macros true]))' | clj-kondo --lang cljs --lint -
<stdin>:1:37: warning: Unknown require option: :include-macros
```

**expected behavior**

I would expect this keyword to be ignored as it is a valid CLJS option.

NoahTheDuke:

---

Closes #1830

Some jumbled thoughts:

~Went to fix the above and realized that this should probably be expanded to a new linter so I wrote it. Makes me wonder if we're gonna need either a "all :require lints" linter with various options or if this step-by-step process is fine.~

~There's a couple levels of granularity to the config, which feels messy. `:include-macros`, `:allow-clojure`, `:unknown-require-options`, and which dialect is being checked all affect this, meaning we have 16 different combinations (some of which have the same outcome). Is there a way to lower this?~

~I noticed that for `:refer-macros`, we're not throwing any warnings if in a `.clj` file, so that's an area for improvement later down the line. Maybe this shouldn't be `:include-macros` specific but `:dialect-require-options` or something?~

Excised the new linter, just throwing `:syntax` error if `:include-macros`'s opt isn't literal `true`. Also checks for cljc and cljs now.

I changed the `:unknown-require-option` default to `:off` as all new linters should be off by default.


test/clj_kondo/unknown_require_option_test.clj
```diff
@@ -15,3 +15,41 @@
   (is (= '()
          (lint! "(ns foo (:require #_:clj-kondo/ignore [bar :s b]))"
                 {:linters {:unknown-require-option {:level :warning}}}))))
+
+(deftest include-macros-cljs-test
+  (is (= '()
+         (lint! "(ns foo (:require [bar :include-macros true]))"
+                {:linters {:unknown-require-option {:level :warning}}}
+                "--lang" "cljs")))
+  (is (= '({:file "<stdin>",
```
borkdude, line 24: Can you use `assert-submaps` here?

test/clj_kondo/unknown_require_option_test.clj
```diff
@@ -15,3 +15,41 @@
   (is (= '()
          (lint! "(ns foo (:require #_:clj-kondo/ignore [bar :s b]))"
                 {:linters {:unknown-require-option {:level :warning}}}))))
+
+(deftest include-macros-cljs-test
+  (is (= '()
```
borkdude, line 20: Can you write `(is (empty? ...))` here?

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -225,6 +225,27 @@
...
+                (do
+                  (if (#{:cljc :cljs} base-lang)
+                    (when-not (true? opt)
+                      (findings/reg-finding!
+                        ctx
+                        (node->line
+                          filename
+                          child-expr
+                          :syntax
+                          "require form is invalid: :invalid-macros only accepts true")))
```
borkdude, line 238: In new message we try to stick to capitalized sentences

borkdude:
You probably need to correct the capital letter in the tests

NoahTheDuke:
Should be fixed now.

PR #1836, opened 2022-10-07
Fallback locations for hooks

borkdude:


borkdude:
See https://clojurians.slack.com/archives/CHY97NXE2/p1664897879130349 for more info

PR #1848, opened 2022-10-20
Add derived-name-location

ericdallo:


src/clj_kondo/hooks_api.clj
```diff
@@ -122,14 +122,18 @@
 (defn annotate
   {:no-doc true}
   [node original-meta]
-  (let [!!last-meta (volatile! (assoc original-meta :derived-location true))]
+  (let [!!last-meta (volatile! (assoc original-meta
+                                      :derived-location true
+                                      :derived-name-location true))]
```
borkdude, line 127: We don't have to add it here, we should only add this to vars when the name expression has `:derived-location` true

ericdallo: And what palaces we know are name-expressions?

ericdallo: And what palaces we know are name-expressions?

borkdude: @ericdallo The expression/node of which we get the `:name-row` and `:name-col`, that one should have `:derived-location-true`, so then we add `:derived-name-location` true to the analysis output.

ericdallo: Ok, just to make sure I got it: I should check inside this annotate function for nodes that have any name location with derived-location true, if so, I should assoc the derived-name-location, is that right?

borkdude: No, nothing has to change in the annotate function. Every node that doesn't have a location will get `:derived-location true`, including the nodes that represent the name of a defined or used var. 

But later on, we propagate `:col`, `:name-col` etc to the analysis. We do this in `analyze-call` for example in `analyzer.cljc`. _There_ we should check if the name node has `:derived-location true` (added by the annotation function), and if it has that, then we also add `:derived-name-location` to the analysis.

ericdallo: Got it, will try soon, thanks!

Issue #1863, opened 2022-11-04
Add Docker pre-commit hook


allie-jo:
**Is your feature request related to a problem? Please describe.**
We've been using clj-kondo as a pre-commit hook in many of our projects, but in order to preempt users from having to install all of the dependencies for each pre-commit hook (including clj-kondo) we've been using Docker images instead. Because we do this manually using the `local` section of our pre-commit-config, we have to update the version manually. It would be great if the repo provided a docker hook directly so we can have CI run `pre-commit autoupdate` to update versions periodically automatically. 

**Describe the solution you'd like**
Add a Docker hook in `.pre-commit-hooks.yaml`, pinned to the currently published docker version of clj-kondo. We've been using:

```yaml
  - id: clj-kondo-docker
    name: clj-kondo (via docker)
    entry: cljkondo/clj-kondo:2022.11.03
    args: [clj-kondo, --lint]
    language: docker_image
    types: [clojure]
```

In our projects successfully for a while now. Would need to templatize `.pre-commit-hooks.yaml` but otherwise is a small change. 

**Additional context**
How we're using clj-kondo in our pre-commits currently:

```yaml
repos:
  ...
  ...
  - repo: local
    hooks:
      - id: clj-kondo
        name: clj-kondo (via docker)
        entry: cljkondo/clj-kondo:2022.11.03
        args: [clj-kondo, --lint]
        language: docker_image
        types: [clojure]
```

Happy to do this PR myself :)

borkdude:
Does this mean that current users of the config suddenly will have to use Docker?

allie-jo:
nope! it's just another option.
if you wanna use the current executable based hook
```yaml
repos:
  - repo: https://github.com/clj-kondo/clj-kondo.git
    rev: v2022.11.02
    hooks:
      - id: clj-kondo
```
and if you wanna use the docker version
```yaml
repos:
  - repo: https://github.com/clj-kondo/clj-kondo.git
    rev: v2022.11.02
    hooks:
      - id: clj-kondo-docker
```

allie-jo:
[hadolint](https://github.com/hadolint/hadolint/blob/master/.pre-commit-hooks.yaml) does something similar, but they don't pin their docker version which causes some issues

borkdude:
> nope! it's just another option.

If that's the case, PR welcome!

PR #1864, opened 2022-11-04
Add docker pre-commit hook

allie-jo:

  - closes https://github.com/clj-kondo/clj-kondo/issues/1863

  - Don't believe there's tests for these build steps?

I've tested the hook and the template generation locally and it all seems to work as expected. I'm not super familiar with babashka tasks and their doesn't seem to be much documentation on the release process. Let me know if i missed anything!


allie-jo:
12 minute PR turnaround. hot damn

borkdude:
@allie-jo 🚀 I wonder if we should set the version to the latest release rather than the snapshot version though? Thoughts on this?

allie-jo:
you're generally not supposed to set the `rev` tag to `master` in pre-commit-configs, just to tags, because the hook wont get updated (pre-commit will warn if you try to do this). so if i understand the release process for clj-kondo, the next time you release, it'll bump to to the non-snapshot version and then anyone that adds the hook would use that release tag for their `rev` which points to the the stable docker image version. If people are pointing their `rev` keys to tags like they're supposed too, they'll never interact with a `SNAPSHOT` version. 

So actually since their's no tag that contains the pre-commit-hooks file with the docker hook yet, the docker-hook will only work when pointing to `master` until you cut a new release :/

borkdude:
We could manually change it to the previous release version for now?

allie-jo:
sure! we could also remove it from the ci integration docs until you cut a release so people don't try to use it until their's a tag they can set it too.

borkdude:
A PR which manually fixes the version seems sufficient I think

Issue #1891, opened 2022-11-29
Add possibility to override project config


staifa:
**Is your feature request related to a problem? Please describe.**
Currently I have no means to override clj-kondo project config (or I wasn't able to find out how). There is a `:config-paths` setting that lets us alter project config, buth the project config takes always precedence when merging. This bugs me for following reasons:

- I'd like to have an editor specific config - for example disabling conjure's log buffer linting is easy, robust and elegant with `:output {:exclude-files ["conjure-log*"]}`
- Our project config is quite benevolent. It's ok to not force certain coding habits/philosophies on others, but I'd like to use as many linters as possible. Brainfarts happen.

**Describe the solution you'd like**
Add possibility to override project config. We would add an `:extra-config-paths` setting to `config.edn`. It would be merged after project config, effectively overrriding it. It would be used the same way as `:config-paths`.

Example:
`{:extra-config-paths ["$XDG_CONFIG_HOME/clj-kondo"]}`

**Describe alternatives you've considered**
Since I'm using `clj-kondo` via `clojure-lsp`, there probably isn't any viable alternative apart from setting the config from `clojure-lsp` itslef.

**Additional context**
I actually created a PR for this on `clojure-lsp` side
https://github.com/clojure-lsp/clojure-lsp/pull/1394
and discussion over the related issue
https://github.com/clojure-lsp/clojure-lsp/issues/1393

I will create a PR if that's ok @borkdude

borkdude:
Sounds good

borkdude:
I've merged support to for the `CLJ_KONDO_EXTRA_CONFIG_DIR` to master. Can you test it there?

staifa:
Sure thing. I'll test it later today.

PR #1893, opened 2022-11-30
Add possibility to override project config #1891

staifa:

Closes https://github.com/clj-kondo/clj-kondo/issues/1891


staifa:
@borkdude set to draft, I realized that `:extra-config-path` overrides command line config and namespace local config too. I'll fix it. How do you feel about overriding the command line config? It may be a good thing in my opinion - you don't have control over it when using `clojure-lsp` for example. The user should be able to completely configure his dev env...

borkdude:
> you don't have control over it when using clojure-lsp for example. The user should be able to completely configure his dev env

yes, but then again, the developer is going to provide extra config via the command line as well, which then should override the extra-configs ;)

I don't think it will often be necessary to override the command line config of clojure-lsp, right? Do you have a good example of that?

staifa:
I don't :) No overriding command line config then!

staifa:
Ready for review @borkdude

borkdude:
Can you make another dummy commit? I'm not able to trigger CI via the admin interface.

src/clj_kondo/impl/core.clj
```diff
@@ -61,10 +61,9 @@
 (defn opts [^java.io.File cfg-file]
   (let [include #(read-fn cfg-file read-edn-file %)]
     {:readers
-     {'include include
+     {'include include}}))
```
borkdude, line 64: Can you revert this? I'd like to keep those comments within the map 🤣

staifa: Sorry, parinfer :)

staifa:
Done.

borkdude:
I'm not sure what's up with CircleCI here, but I'll get back to this PR in the next few days.

One thought I had: maybe it would be nice to be able to provide a config override that doesn't have to be visible in the checked-in code? E.g. via en environment variable or so?

staifa:
> I'm not sure what's up with CircleCI here, but I'll get back to this PR in the next few days.
> 
> One thought I had: maybe it would be nice to be able to provide a config override that doesn't have to be visible in the checked-in code? E.g. via en environment variable or so?

Neat idea, I pushed a commit implementing env var config paths.

doc/config.md
```diff
@@ -51,6 +51,11 @@ Clj-kondo can be configured in several ways:
 - project config: a `config.edn` file in the `.clj-kondo` directory (see
   [project setup](../README.md#project-setup)).
 - `:config-paths` in project `config.edn`: a list of directories that provide additional config.
+- `:extra-config-paths` in project `config.edn`: a list of directories that provide additional config
```
borkdude, line 54: Should we still have the `:extra-config-paths` if we have the environment variable?

Is it easy to set the environment variable for you editor on a project by project basis, have you tested this?

staifa:
Ok, only env var handling is implemented now. I hand tested ant everything works as expected. I removed tests for now - I noticed you don't use `with-redefs`, how would you test setting and unsetting the env variable?

borkdude:
I feel like we need to think a little more about this one in the coming weeks, so I haven't included in the current release yet, but I will come back to this.

PR #1925, opened 2022-12-21
Use backwards-compatible option for xargs

Heliosmaster:
From the manual
```
     -r      Compatibility with GNU xargs.  The GNU version of xargs runs the utility argument at least once, even if xargs input is empty, and it supports a -r
             option to inhibit this behavior.  The FreeBSD version of xargs does not run the utility argument on empty input, but it supports the -r option for
             command-line compatibility with GNU xargs, but the -r option does nothing in the FreeBSD version of xargs.
```

### DIFF
--- a/doc/ci-integration.md
+++ b/doc/ci-integration.md
@@ -32,7 +32,7 @@ else
     against=$(git hash-object -t tree /dev/null)
 fi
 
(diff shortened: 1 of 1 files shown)


borkdude:
Thank you!

Issue #1920, opened 2022-12-16
Encourage idiomatic closure definition


andreyorst:
I've seen some examples of misuse of `def` + `let` + `fn` combinations like this one:

```clojure
(def add-three
  "Function that adds three to the given number"
  (let [x 3]
    (fn [y] (+ x y))))
```

Which can be written in a more idiomatic way:

```clojure
(let [x 3]
  (defn add-three 
    "Function that adds three to the given number"
    [y]
    (+ x y)))
```

This has the additional benefit of providing useful metadata with the function name, such as function arglist.

## Solution 1

I've made a hook, that I've included in my local config, and I've thought that this can be one of the default linting rules for clj-kondo, as the code is more concise with `let` + `defn`:

```clojure
(ns hooks.def-let-fn
  (:require [clj-kondo.hooks-api :as api]))

(defn def-let-fn [{:keys [node]}]
  (let [body (rest (:children node))]
    (when-let [[name docstring body]
               (cond (= 3 (count body)) body
                     (= 2 (count body)) [(first body) nil (last body)]
                     :else nil)]
      (when (and (api/list-node? body)
                 (= 'let (first (api/sexpr body))))
        (let [[bindings body] (rest (:children body))]
          (when (and (api/list-node? body)
                     (= 'fn (first (api/sexpr body))))
            (api/reg-finding!
             (assoc (meta (first (:children body)))
                    :message
                    (format "consider using 'defn' inside 'let' instead of 'fn' inside 'let' inside 'def': (let %s %s)"
                            (api/sexpr bindings)
                            (list* 'defn
                                   (api/sexpr name)
                                   (if docstring
                                     (list* (api/sexpr docstring) (rest (api/sexpr body)))
                                     (rest (api/sexpr body)))))
                    :type :hook))))))))
```

It works, and even suggests the corrected variant, that you can copy from the message, only requiring you to reformat the code afterward.

## Solution 2

I can, of course, keep it as a local hook, or propose its inclusion into the [configs](https://github.com/clj-kondo/configs) project, but I think it would be better as a default rule.

I'm not sure how built-in checks are implemented, but I'm willing to try adding this myself if it is in the scope of the project.

Thanks!

borkdude:
It would be interesting to collect some data on how often this occurs in code bases.
I think https://github.com/borkdude/grasp could be used to detect some libraries that are doing the first pattern, where the second pattern should be preferred.

borkdude:
[...]

PR #1946, opened 2023-01-14
add a new def-fn linter

andreyorst:

Hi and sorry for the delay!

As discussed in https://github.com/clj-kondo/clj-kondo/issues/1920 I've added a new linter for this case. I've added tests for the common cases, and some cases that you've listed in the comments/

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -866,16 +866,15 @@
           :def-let-fn)))
 
 (defn- reg-def-fn! [ctx expr filename kind]
-  (when-not (linter-disabled? ctx :def-fn)
-    (findings/reg-finding!
-     ctx
-     (node->line
-      filename
-      expr
-      :def-fn
-      (case kind
-        :def-fn "consider using 'defn' instead of 'fn' inside 'def'"
-        :def-let-fn "consider using 'defn' inside 'let' instead of 'fn' inside 'let' inside 'def'")))))
+  (findings/reg-finding!
+   ctx
+   (node->line
+    filename
+    expr
+    :def-fn
+    (case kind
+      :def-fn "Use defn instead of def + fn"
+      :def-let-fn "Use let + defn instead of def + let + fn"))))
 
 (defn analyze-fn [ctx expr]
   (let [ctx (assoc ctx :seen-recur? (volatile! nil))
@@ -911,7 +910,8 @@
         varargs-min-arity (when arities (get-in arities [:varargs :min-arity]))
         arglist-strs (when arities (into [] (keep :arglist-str) (vals arities)))
         parsed-bodies (mapcat :parsed parsed-bodies)]
-    (when-let [kind (def-fn? ctx)]
+    (when-let [kind (and (not (linter-disabled? ctx :def-fn))
+                         (def-fn? ctx))]
       (reg-def-fn! ctx expr filename kind))
     (with-meta parsed-bodies
       (when arities
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -866,15 +866,16 @@
           :def-let-fn)))
 
 (defn- reg-def-fn! [ctx expr filename kind]
-  (findings/reg-finding!
-   ctx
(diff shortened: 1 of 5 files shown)


borkdude:
@andreyorst I think `:def-fn` should be disabled by default. This will likely also fix the failing tests.

andreyorst:
> @andreyorst I think `:def-fn` should be disabled by default (only very serious lint issues are enabled by default, since a while, since it would break people's builds if new linters like this were enabled by default). This will likely also fix the failing tests.

Rather than disabling, maybe it should be an `info` level lint?

borkdude:
The policy is that stylistic linters are off by default.

borkdude:
Can you fix the remaining failing tests? This can be done by supplying the explicit config. See other tests for examples.

andreyorst:
> Can you fix the remaining failing tests? This can be done by supplying the explicit config. See other tests for examples.

sure

andreyorst:
Sorry, got a bit messed up by my `~/.config/clj-kondo/config.edn` which included custom hooks, that I've mentioned in the parent issue, which messed with running tests in the project. Wasn't expecting that the user config will be picked up.

andreyorst:
@borkdude, I'm not sure what's up with CircleCI, but everything else seems to work

borkdude:
Thanks! I don't know what's up with CircleCI either, but this happens for some users, whereas for others it doesn't. Very weird. Anyway, I'll have a look later today. Thanks a lot!

andreyorst:
> but this happens for some users, whereas for others it doesn't

May have something to do with this: https://support.circleci.com/hc/en-us/articles/360043679453-CircleCI-Terms-of-Service-Violation-Sanctioned-Country

borkdude:
Looks good, but there's two things left to do:

1. For linters that aren't on by default, we check if they are disabled and then skip the check (to save some cycles). There is a function `utils/linter-disabled?` that you can use for this.

2. Update the `linters.md` page with the documentation for this new linter.

andreyorst:
Done.

What do you think about changing messages from:

- `consider using 'defn' instead of 'fn' inside 'def'`
- `consider using 'defn' inside 'let' instead of 'fn' inside 'let' inside 'def'`

To:

- `consider using 'defn'`
- `consider using 'defn' inside 'let'`

---

Also, in the hooks I've provided in the issue ticket, I've constructed a fixed version of the code, but as far as I know, Clj-kondo doesn't have an API for suggestions. LSP servers often have such a thing that they can fix your code based on some suggestion from the linter, so I was wondering if there's something like that?

borkdude:
> LSP servers often have such a thing that they can fix your code based on some suggestion from the linter

This would have to be done as part of clojure-lsp, let's address that separately after we get this merged

src/clj_kondo/impl/analyzer.clj
```diff
@@ -886,6 +911,8 @@
         varargs-min-arity (when arities (get-in arities [:varargs :min-arity]))
         arglist-strs (when arities (into [] (keep :arglist-str) (vals arities)))
         parsed-bodies (mapcat :parsed parsed-bodies)]
+    (when-let [kind (def-fn? ctx)]
```
borkdude, line 914: I would put the linter-disabled? call here so you can avoid the call to `def-fn?` when it's not necessary

borkdude:
About the change in messages. Proposal, just:

```
Use defn instead of def + fn
```

in all cases, without single quotes (these aren't used anywhere else in the messages either).

andreyorst:
Changed messages to:

- `Use defn instead of def + fn`
- `Use let + defn instead of def + let + fn`

Though I don't think `let + defn` conveys the idea clearly enough

borkdude:
I'd say let's keep it simple and have only one message. We can add to the docs how to rewrite `let + def + fn` to `defn`. We can refer to those docs in case people have questions. They will probably read those docs anyway since the linter isn't enabled by default.

andreyorst:
should be good now

borkdude:
Merged via https://github.com/clj-kondo/clj-kondo/commit/51ccf596e42d7a4bc43af87ed2371d74d2059371 - thanks a lot!

PR #1947, opened 2023-01-14
Avoid side-effects on `clj-kondo.impl.core` ns load

ericdallo:
This is more of a refactor to avoid loading a resource on ns load

### DIFF
--- a/src/clj_kondo/impl/core.clj
+++ b/src/clj_kondo/impl/core.clj
@@ -18,9 +18,9 @@
 
 (def dev? (= "true" (System/getenv "CLJ_KONDO_DEV")))
 
-(def version
+(defn version []
   (str/trim
-   (slurp (io/resource "CLJ_KONDO_VERSION"))))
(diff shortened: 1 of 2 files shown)


borkdude:
What problem does this solve?

ericdallo:
I was refactoring clojure-lsp jar build and accidentally found that I missed to include some resources, then I found that clj-kondo does that on ns loading, so I thought would be nice to avoid side effects to io during ns load, no particular issue

borkdude:
Doing this at ns load time helps with GraalVM, it's a compile time constant, so that you don't need to include it in your GraalVM resource configuration. As such, I suspect this would be a breaking change. You can test this by compiling this locally and then doing `clj-kondo --version` which I think won't print the right version anymore. Since this doesn't really solve a problem, but introduces one, I think it's better to close this PR. The PR template is there for a reason.

ericdallo:
No problem, feel free to close it

Issue #1362, opened 2021-08-19
unresolved-symbol for all-lowercase class name


vemv:
The following code:

```clj
(def ^name.fraser.neil.plaintext.diff_match_patch dmp (diff_match_patch.))
```

will result in `error: Unresolved symbol: name.fraser.neil.plaintext.diff_match_patch`

(I place the fully-qualified type hint for 1.7 compat)

`name.fraser.neil.plaintext.diff_match_patch` is a valid class name as found in `[org.clojars.brenton/google-diff-match-patch "0.1"]`

**Environment**

`[clj-kondo "2021.08.06"]`, JVM at runtime

borkdude:
Thanks!

Can you explain what you meant with 1.7 compat?

Just a note to self, the type hint in that position on a var can take care of reflection warnings, e.g.:

```
user=> (set! *warn-on-reflection* true)
true
user=> (def x nil)
#'user/x
user=> (defn foo [] (.length x))
Reflection warning, NO_SOURCE_PATH:1:14 - reference to field length can't be resolved.
#'user/foo
user=> (def ^String x nil)
#'user/x
user=> (defn foo [] (.length x))
#'user/foo
```

vemv:
> Can you explain what you meant with 1.7 compat?

Clojure <= 1.7 generally needs type hints to be fully qualified, else they often will be wrong when consumed from another ns

borkdude:
You mean fully qualified vs the imported classname right?

borkdude:
The reason this currently is a false positive in clj-kondo has to do with the heuristic of how classes are detected. E.g. this one works:

```
(def ^name.fraser.neil.plaintext.Diff_match_atch dmp (diff_match_patch.))
```

because of the capital letter.

vemv:
> You mean fully qualified vs the imported classname right?

Here's a description of the behavior pre clojure 1.8 https://github.com/metrics-clojure/metrics-clojure/issues/66

i.e. it didn't matter if one `:import`ed a class

vemv:
🍻

PR #1985, opened 2023-02-06
Lint constructor calls providing unresolved-symbol

### LINKED ISSUE #1984: unresolved-symbol for non imported classes when using constructor

clojure-lsp suggest to add missing import class using `java-class-definition` if clj-kondo reports a class as unresolved symbol or unresolved-namespace:
```clojure
(ns foo)
(File/createTempFile ...) ;; unresolved-namespace then clojure-lsp suggests add import java.io.File 
File ;; unresolved-symbol then clojure-lsp suggests add import java.io.File 

(File. ...) ;; no clj-kondo lint report, so no clojure-lsp code action
```

As you can see, clj-kondo doesn't report for classes being initialized via constructor which makes the user don't know if that class is really working because was imported or not and if not imported, clojure-lsp doesn't suggest to add import.

Report classes constructors that are not java.lang or clojure.lang or defrecord/deftype etc

Do nothing?

**Additional context**

I can help with a PR if pointed to where should be changed

ericdallo:
Fixes #1984

During development I noticed clj-kondo already not mark as unresolved java.lang classes, so AFAICS we don't need to change anything, I added a test to test java.lang is linted correctly.

### DIFF
--- a/test/clj_kondo/unresolved_symbol_test.clj
+++ b/test/clj_kondo/unresolved_symbol_test.clj
@@ -1,6 +1,6 @@
 (ns clj-kondo.unresolved-symbol-test
   (:require
-   [clj-kondo.test-utils :refer [lint! assert-submaps2] :rename {assert-submaps2 assert-submaps}]
+   [clj-kondo.test-utils :refer [assert-submaps2 lint!] :rename {assert-submaps2 assert-submaps}]
    [clojure.edn :as edn]
    [clojure.java.io :as io]
    [clojure.test :refer [deftest is testing]]))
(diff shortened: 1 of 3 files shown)


test/clj_kondo/unresolved_symbol_test.clj
```diff
@@ -241,8 +250,13 @@
  (:foo x))
 
 " '{:linters {:unresolved-symbol {:exclude [(slingshot.slingshot/try+)]}}})))
-  (is (empty? (lint! "(def ^name.fraser.neil.plaintext.diff_match_patch dmp (diff_match_patch.))"
-                     '{:linters {:unresolved-symbol {:level :error}}}))))
+  (assert-submaps
+   '({:file "<stdin>", :row 1, :col 56, :level :error, :message "Unresolved symbol: diff_match_patch"})
+   (lint! "(def ^name.fraser.neil.plaintext.diff_match_patch dmp (diff_match_patch.))"
```
borkdude, line 255: I think this test was added to test that `foo.bar.foo_bar` is detected as a class name despite not having capital letters. Could you leave that in? The constructor call can be removed.

borkdude: See https://github.com/clj-kondo/clj-kondo/issues/1362

ericdallo: I saw that and thought that we could lint only the constructor, but makes sense to remove the constructor since it's not what the issue is related to, will do

ericdallo: Done! @borkdude

PR #2008, opened 2023-03-12
analysis: support java member definitions

ericdallo:

- [x] Update documentation in `analysis.md`

## What was done

This is a base first iteration for #1983, it adds a new bucket `java-member-definitions` which have both methods and fields of classes analyzed with ASM, the tests should catch all code changes, and I tested manually with clojure-lsp branch here adding support for completion of static fields like the image:

image

## What was not done

- Completions starting from the slash like `(AtomicDouble/|)` doesn't work because clj-kondo doesn't return a `java-class-usage` for AtomicDouble, we should fix it in another PR
- ASM doesn't support include docstrings and locations since those are not included in the .class, that's a pity as would be really helpful for LSP to show the documentation and being able to find definition of those members (we can still go to the class definition).
- I couldn't implement easily in clojure-lsp support for completions that are not static members yet, but I think if kondo provide a way to add that a `instance-invocation` is a usage of class X, then probably would be possible.


ericdallo:
There is a reflection issue for graal related to `org.objectweb.asm.Opcodes`, any tips?

borkdude:
@ericdallo It's not a reflection issue but a build-time vs runtime init issue. I think you can fix it by adding to this line:

```
--initialize-at-build-time=com.fasterxml.jackson
```

in native-image properties, or, maybe better: use a delay instead of a top level class instance. I prefer the latter.

src/clj_kondo/impl/analysis/java.clj
```diff
@@ -45,28 +75,39 @@
...
-                                (class->class-name file)
-                                (str/ends-with? file ".java")
-                                (source->class-name file))))]
-      (swap! (:analysis ctx)
-             update :java-class-definitions conj
-             {:class class-name
-              :uri (->uri jar entry file)
-              :filename (or file
-                            (str jar ":" entry))}))))
+    (let [class-is (or (and jar entry (.getInputStream jar entry))
```
borkdude, line 83: I think we need to wrap this in a `with-open` to else we will get too many files open exceptions

src/clj_kondo/impl/analysis/java.clj
```diff
@@ -19,18 +27,40 @@
...
+(def ^:private opcode->flags
+  {Opcodes/ACC_PUBLIC #{:public}
+   Opcodes/ALOAD #{:public :field :static}
+   Opcodes/SIPUSH #{:public :field :final}
+   Opcodes/LCONST_0 #{:public :method :static}})
+
+(defn ^:private class-is->class-info
+  "Parse class-bytes using ASM."
+  [^InputStream class-is]
+  (let [class-reader (ClassReader. (input-stream->bytes class-is))
```
borkdude, line 39: We may need `with-open` here too? Not sure.

borkdude:
@ericdallo LGTM aside from the `with-open` stuff. I think it'd be good to run this against a large amount of jar files to see if we don't leak any resources (file handles).

One more thing to add: docs in `analysis.md`.

borkdude:
> I couldn't implement easily in clojure-lsp support for completions that are not static members yet

In the absence of type hints, perhaps just auto-complete all instance methods? Perhaps as a second iteration indeed.

src/clj_kondo/impl/analysis/java.clj
ericdallo: I don't think we need here because that private function already does that [here](https://github.com/clj-kondo/clj-kondo/pull/2008/files#diff-40ca05e82ee1bafb0e01402be6bc6119871dc41c1b65f9b830634d94f3ac1b10R18), only returning the byte array

ericdallo:
> @ericdallo LGTM aside from the with-open stuff. I think it'd be good to run this against a large amount of jar files to see if we don't leak any resources (file handles).

I was using clojure-lsp as has a considerable amount of jars, but I will try with bigger projects.

> One more thing to add: docs in analysis.md.

Well remembered!

ericdallo:
> In the absence of type hints, perhaps just auto-complete all instance methods? Perhaps as a second iteration indeed.

the issue is that a `instance-invocation` that kondo provides ATM, doesn't contain any info about the class, so we don't even the class to seek members

borkdude:
> so we don't even the class to seek members

I know, but we could search all classes, just as a dumb first try?

ericdallo:
> I know, but we could search all classes, just as a dumb first try?

Yes, but probably will affect performance since Usually each class has more than 1 member

ericdallo:
@borkdude I added docs for all missing analysis we added in the past and forgot 😅 LMK what you think

borkdude:
Looks good. I'm not sure why the pod tests are failing. Perhaps your PR isn't up to date with master? Not sure.

deps.edn
```diff
@@ -4,8 +4,7 @@
         cheshire/cheshire {:mvn/version "5.11.0"}
         nrepl/bencode {:mvn/version "1.1.0"}
         org.babashka/sci {:mvn/version "0.7.38"}
-        babashka/fs {:mvn/version "0.2.16"}
```
borkdude, line 7: This probably explains the issue with the pod test.

ericdallo: my bad, fixing it

ericdallo:
I will want to make https://github.com/clojure-lsp/clojure-lsp/issues/204 work somehow, but we can discuss in a next iteration

borkdude:
Reflection warning, clj_kondo/impl/analysis/java.clj:84:5 - reference to field close can't be resolved

PR #2009, opened 2023-03-12
Linter for vars that differ only by case

### LINKED ISSUE #2006: Linter for vars that differ only in case
If you (or one of your dependencies) have two vars that differ only in case (e.g. an example I ran into recently: [clojure2d.color/to-lch](https://github.com/Clojure2D/clojure2d/blob/c34b435ef86941c8121dc4802e0ccbc0a6ce949f/src/clojure2d/color.clj#L903) and [clojure2d.color/to-LCH](https://github.com/Clojure2D/clojure2d/blob/c34b435ef86941c8121dc4802e0ccbc0a6ce949f/src/clojure2d/color.clj#L1761)), building an uberjar will fail on case insensitive filesystems, because one class file will overwrite the other.

It would be good to have a linter that detected this.

I'd be happy to write a PR for this, but might need a couple of pointers to get started.

emlyn:
I've made a start on this PR that fixes #2006.

It basically works:
```
❯ clojure -M:clj-kondo/dev --lint - <<< "(def bar 1) (def Bar 2)"
<stdin>:1:13: warning: Bar differs only in case from bar
linting took 45ms, errors: 0, warnings: 1
```

But has a couple of false positives (`(definterface Foo (foo []))`, `(defprotocol Foo (foo []))`, and possibly other similar cases). What's the best way to detect we're inside an interface or protocol definition and skip the check?

I also noticed that this triggers even if one or both vars are in a comment form.
```
❯ clojure -M:clj-kondo/dev --lint - <<< "(comment (def bar 1)) (def Bar 2)"
<stdin>:1:23: warning: Bar differs only in case from bar
linting took 44ms, errors: 0, warnings: 1
```
I could bypass the check when `(:in-comment ctx)` is true, but that only works if the uncommented one is first. If a commented var is first, I can't see an obvious way to detect that from the second declaration - I think I'd have to add some data into the `vars` map? Or maybe it's fine as it is?


borkdude:
About the vars in comments: we could probably just ignore those.

About names of protocols and their methods: do I understand that there is not a case sensitivity problem there? Why not?

borkdude:
I this this problem only happens for vars that refer to functions, such as `(ns foo) (defn foo []) (defn Foo [])`.

When I compile a namespace like this:

```
(ns scratch)

(defn Foo1 [])

(defn FOO1 [] (Foo1))
```

I only see one `scratch$Foo1.class` file and not a second `scratch$FOO1.class` which surprises me.

borkdude:
Ah, macOS APFS is also case-insensitive by default.

So only `(defn foo [])` / defmacro / defn-, etc and `(def f (fn []))` will be problematic when they have overlapping names wrt case sensitivity.

So let's only collect those names and check for those.

emlyn:
I ran a check to work out what constructs write their own class files, and found this:
```clojure
;; These don't write a class file:
(def test-def :test)
(defonce test-defonce :test)
(def test-def-fn (fn test-fn [] :test))
(defmulti test-multi (fn [a] a))
(defmethod test-multi :default [a] a)
(defstruct test-struct :test)

;; These do:
(defn test-defn [] :test)
(defn- test-defn- [] :test)
(defmacro test-macro [] :test)
(definline test-definline [_] :test)
(deftype test-type [_])

;; These do, but `reg-var!` also gets called on some methods and/or constructors:
(definterface test-interface (testinterfacemethod []))
(defprotocol test-protocol (testprotocolmethod [_]))
(defrecord test-record [a] test-protocol (testprotocolmethod [_] :test))
;; (for the methods/constructors we always have :fixed-arities set on the metadata, so can use that to exclude them)
```

I think it's getting closer now - at least I'm not breaking any of the tests anymore :-)

borkdude:
I ran the `script/diff` script which runs the current branch compared to the master branch over a bunch of code. I'm finding this:

```
1678d1677
< cljs/vendor/cognitect/transit.clj:139:1: warning: writer differs only in case from Writer
1684d1682
< cljs/vendor/cognitect/transit.clj:290:1: warning: reader differs only in case from Reader
1804d1801
< clojure/core.clj:7762:1: warning: eduction differs only in case from Eduction
1821d1817
< clojure/core/reducers.clj:255:1: warning: cat differs only in case from Cat
1894d1889
< clojure/data.clj:124:1: warning: diff differs only in case from Diff
1920d1914
< clojure/gvec.clj:170:1: warning: Vec differs only in case from vec
2229,2230d2222
< clojure/spec/alpha.clj:158:1: warning: specize differs only in case from Specize
< clojure/spec/alpha.clj:368:1: warning: spec differs only in case from Spec
2422d2413
< cognitect/transit.clj:139:1: warning: writer differs only in case from Writer
2428d2418
< cognitect/transit.clj:290:1: warning: reader differs only in case from Reader
2778d2767
< sci/impl/test.cljc:32:1: warning: Foo differs only in case from foo
3015c3004
```

I think we should look whether these are serious cases or not.

borkdude:
I also started a discussion about this on Clojurians Slack and pinged you there.

emlyn:
I've modified it to check both cases separately, hopefully that fixes the false positives.

emlyn:
I ran the diff script, and the output is:
```
Cloning into 'clj-kondo'...
remote: Enumerating objects: 18430, done.
remote: Counting objects: 100% (1762/1762), done.
remote: Compressing objects: 100% (484/484), done.
remote: Total 18430 (delta 1460), reused 1413 (delta 1219), pack-reused 16668
Receiving objects: 100% (18430/18430), 13.32 MiB | 11.85 MiB/s, done.
Resolving deltas: 100% (11136/11136), done.
2767d2766
< sci/impl/test.cljc:32:1: warning: Foo differs only in case from foo
3004c3003
< linting took 6970ms, errors: 300, warnings: 2703
---
> linting took 7061ms, errors: 300, warnings: 2702
```

emlyn:
Looks like that's coming from [here](https://github.com/babashka/sci/blob/617bf809d1810b48cab24980864b85249c2a9d8b/src/sci/impl/test.cljc#L29-L33). I don't think that's a false positive - according to the details we worked out, that could cause issues if AOT compiled (which presumably isn't an issue for SCI).

borkdude:
Indeed, I can easily remove that code. In test namespace this check might not be important though, but maybe it's a good habit to avoid it everywhere.

emlyn:
I think the logic is correct now, but it is recalculating the lists of lower-cased var names each time the check is run. Do you think it would be worth keeping track of them to avoid the recalculations and improve performance?

borkdude:
@emlyn I haven't looked in detail at the code but if we can speed it up, I think it's worth it. Small things add up over time.

emlyn:
I'll have a look into it, the current implementation does feel a bit inefficient - I think it's running O(n^2) `str/lower-case`s, which isn't great, even if `n` is not that big.

borkdude:
We could speed it up by keeping a set of lowercased strings prepared and if we find a collision, only then we will search for the prior existing var's real name, something like that?

borkdude:
One other thing to consider: in ClojureScript (and perhaps other dialects) the above linter isn't strictly necessary. Should we enable it by default nonetheless for those?

emlyn:
Either a set of lower-cased var names, or a map of lower-cased name to original name (or rather, to vector of original names, as there could be more than one).

I was also wondering if this should be restricted to clj mode only. I'm not really sure what's best.

borkdude:
map sounds good

emlyn:
I've just realised I was reading the timings backwards, so it was apparently running slightly faster with my check added (I'm not suggesting the extra code is making it faster, it's probably a statistical fluke, or maybe a cache warming effect if the new code is run second).

In any case, I've made the change, and I think it might even be a little cleaner like this. Timing is still fine (now apparently 2.9% faster):

```
Cloning into 'clj-kondo'...
remote: Enumerating objects: 18430, done.
remote: Counting objects: 100% (1878/1878), done.
remote: Compressing objects: 100% (483/483), done.
remote: Total 18430 (delta 1571), reused 1532 (delta 1336), pack-reused 16552
Receiving objects: 100% (18430/18430), 13.32 MiB | 14.12 MiB/s, done.
Resolving deltas: 100% (11135/11135), done.
2767d2766
< sci/impl/test.cljc:32:1: warning: Foo differs only in case from foo
3004c3003
< linting took 6678ms, errors: 300, warnings: 2703
---
> linting took 6874ms, errors: 300, warnings: 2702
```

borkdude:
This looks good. What remains is updating the `doc/linters.md` page with the new linter. I think I'd prefer adding `-name` into the linter name: `:var-same-name-except-case` or something similar?

borkdude:
It would be good to add an explanation for this new (non-optional) linter in `linters.md`. I feel that I can't properly explain this to CLJS users so in `(:base-lang :cljs)` we might just turn this off.

emlyn:
Yes, that all makes sense. I've wrapped the check in `when (identical? :clj lang)` is that OK? What's the difference between `lang` and `base-lang`?

Another small thing: I think the wording of the message could be a bit confusing (e.g. what if the var was called "something"), would it make sense to quote the var name like: `(str "#'" ns-sym "/" var-sym " differs only in case from " (str/join ", " clashing-vars))`

borkdude:
- is that OK?

I think that's actually what we want yes, `:lang :clj` means either source code from a `.clj` file or code from a `:clj` branch in a`.cljc` file.

I think we can change the wording in:

```
Var name Foo differs only in case from: foo(, ...)
```

borkdude:
Thanks!

Issue #2030, opened 2023-03-27
Add the ability to warn against usage of certain tagged literals


GlassAndOneHalf:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/clj-kondo/clj-kondo/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

**Is your feature request related to a problem? Please describe.**

I have been converting a codebase to use the new Java time APIs, replacing `java.util.Date`. To discourage developers from using `java.util.Date` in the future, I would like to be able to warn against any usage of the `#inst` tagged literal, as this constructs `java.util.Date` instances.

**Describe the solution you'd like**

Ideally a linter that allows me to specify a tagged literal that should be flagged as a warning/error whenever it is used.

For example, in the `clj-kondo` config:
```
:linters {
  :discouraged-tag { inst { :message "Prefer #java-time/instant over #inst" } }
}
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #2031, opened 2023-03-27
Add a lint for discouraged tag literals.

GlassAndOneHalf:
Follow up to #2030.

Add a new `:discouraged-tag` linter.

===

### DIFF
--- a/test/clj_kondo/discouraged_tag_test.clj
+++ b/test/clj_kondo/discouraged_tag_test.clj
@@ -1,46 +1,51 @@
 (ns clj-kondo.discouraged-tag-test
   (:require
-   [clj-kondo.test-utils :as tu :refer [lint! assert-submaps]]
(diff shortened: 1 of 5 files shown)


test/clj_kondo/discouraged_tag_test.clj
```diff
@@ -0,0 +1,46 @@
+(ns clj-kondo.discouraged-tag-test
```
borkdude, line 1: Thanks. I think using assert-submaps2 gives better error message, so let's use that. I'm slowly phasing assert-submaps(1) out.

discouraged-var also supports setting the :level per var. I think that would be good to support here as well. 
Happy to help or take over, when you're stuck.

GlassAndOneHalf: Added in  e9538ec :)

borkdude:
Excellent @GlassAndOneHalf !

PR #2036, opened 2023-04-03
False positive `:def-fn` on `def` + `reify`

sumbach:

    - I mentioned this issue at https://github.com/clj-kondo/clj-kondo/issues/1920#issuecomment-1474551676 but haven't opened a separate GH issue for it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#2030](https://github.com/clj-kondo/clj-kondo/issues/2030): Add a new `:discouraged-tag` linter for discouraged tag literals. See the [docs](https://github.com/clj-kondo/clj-kondo/blob/master/doc/linters.md#discouraged-tag).
 - Add `:gen` support on `clojure.spec.alpha/keys`
 - [#1647](https://github.com/clj-kondo/clj-kondo/issues/1647): support `:exclude-patterns` in `:unresolved-symbol` linter
+- [#2036](https://github.com/clj-kondo/clj-kondo/pull/2036): False positive `:def-fn` on `def` + `reify`
 
 ## 2023.03.17
 
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -874,6 +874,9 @@
       varargs-min-arity (assoc :varargs-min-arity varargs-min-arity)
       (seq arglist-strs) (assoc :arglist-strs arglist-strs))))
 
+(defn- reify? [x]
+  (one-of x [[clojure.core reify] [cljs.core reify] [cljs.core specify!]]))
+
 (defn- def? [x]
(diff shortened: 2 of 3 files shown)


sumbach:
@borkdude I took a quick attempt at writing a failing test demonstrating the issue. I could definitely use your input on how to _resolve_ the issue, though 😉 

I think I see the cause:

* `reify` form is ultimately run through `analyze-fn`: https://github.com/clj-kondo/clj-kondo/blob/f1b13914ee192953f1ef/src/clj_kondo/impl/analyzer.clj#L1442-L1444
* `analyze-fn` reports `:def-fn` finding if parents of current form are `def` or `def` + `let`: https://github.com/clj-kondo/clj-kondo/blob/f1b13914ee192953f1ef/src/clj_kondo/impl/analyzer.clj#L931-L933

One approach I can think of is to add a check to `def-fn?` to ensure that the current form is, in fact, `fn`. This could take the form of an allow list including:

* `[clojure.core fn]` / `[cljs.core fn]`
* `[clojure.core fn*]` / `[cljs.core fn*]`
* `[schema.core fn]` (maybe?)
* `[clojure.core bound-fn]` (maybe?)

Or a block list of forms to exclude:

* `[clojure.core reify]` / `[cljs.core reify]`
* `[cljs.core specify!]` (didn't know about this one until today 😬 )

NOTE: The block list approach might still report `:def-fn` findings on weird stuff like `defmethod`, `deftype`, `extend-protocol`, `extend-type` within `def` -- I don't think any of these would be useful and probably shouldn't occur within normal code.

Do you have any concerns with this approach? Any preference on allow list vs block list? Any other alternatives to consider?

borkdude:
Yes, I think the allow list sounds good, just include core fn and see if any tests with :def-fn are breaking and change accordingly.

sumbach:
Thanks for the instant feedback @borkdude ! 😀 

I tried both the block list and the allow list (on separate commits within this PR).

I was initially convinced the allow list was better (since it should eliminate all false positives, known and unknown). But there's only one known false positive, `reify`, and quite a few different fn-ish forms that clj-kondo treats as fn. With the allow list, we would need to be vigilant about adding any new fn-ish forms to the `fn-form?` predicate.

For example, should `fn-form?` include `[day8.re-frame.tracing fn-traced]`?

borkdude:
Thinking more about it, I think adding an extra callstack element is better. I pushed this to master now and together with your PR (test + changelogs) this will be the full solution. Can you update your PR?

sumbach:
Yes! The extra element in callstack looks like an excellent solution to me as well.

I've pushed some updates to the branch, including tests for `{:lint-as {some.ns/my-fn clojure.core/fn}}` and `{:lint-as {some.ns/my-reify clojure.core/reify}}`.

Please let me know you want the git history cleaned up. I'm happy to rebase/squash/etc, but wanted to check with you before doing a force push.

Thanks!!!

borkdude:
Thanks!

PR #2053, opened 2023-04-16
analysis: parse java files via javaparser to produce java-member-definitions

### LINKED ISSUE #1983: analysis: java-method-definitions support

I couldn't find an open issue, so this is a continuation of #1330.
The feature most Cursive users miss from clojure-lsp is the java analysis, clojure-lsp ATM has ok support for classes using clj-kondo `java-class-definition` and `java-class-usages`, this is enough to make navigation between classes work which is awesome, but we still miss method analysis, mainly to:
- complete java methods, returning all public methods of a java class `(Thread/|)`
- present hover information showing signature and docstring

We discussed in #1330 about using some tool like ASM to analyze a .java (and/or .class) file and parse the analysis information, we could use that and provide a `:java-metho-definition` and maybe `:java-method-usage` in the future.
I created a demo as a POC: https://github.com/ericdallo/asm-sample

I considered using some external tool on clojure-lsp itself but would not be convenient as all analysis come from kondo, we would need to manage one more tool and manage both analysis when quering.

**Additional context**

I can help working in a PR

ericdallo:

Fixes #1983

### DIFF
--- a/src/clj_kondo/impl/analysis/java.clj
+++ b/src/clj_kondo/impl/analysis/java.clj
@@ -6,7 +6,7 @@
    [clojure.set :as set]
    [clojure.string :as str])
   (:import
-   [com.github.javaparser JavaParser Range]
+   [com.github.javaparser JavaParser ParserConfiguration Range]
    [com.github.javaparser.ast
     CompilationUnit
     Modifier
@@ -22,6 +22,8 @@
    [com.github.javaparser.ast.comments Comment]
    [com.github.javaparser.ast.expr SimpleName]
    [com.github.javaparser.metamodel PropertyMetaModel]
+   [com.github.javaparser.symbolsolver JavaSymbolSolver]
+   [com.github.javaparser.symbolsolver.resolution.typesolvers ReflectionTypeSolver]
    [java.io File InputStream]
    [java.util.jar JarFile JarFile$JarFileEntry]
    [org.objectweb.asm
@@ -100,10 +102,10 @@
                    Modifier {:flags (set (map #(modifier-keyword->flag (.getKeyword ^Modifier %)) value-or-list))}
                    Comment (some->> ^Comment value-or-list .asString (hash-map :doc))
                    VariableDeclarator {:name (.asString (.getName ^VariableDeclarator (first value-or-list)))
-                                       :type (.asString (.getType ^VariableDeclarator (first value-or-list)))}
+                                       :type (.describe (.resolve (.getType ^VariableDeclarator (first value-or-list))))}
                    Parameter {:parameters (mapv #(.toString ^Parameter %) value-or-list)}
                    SimpleName {:name (.asString ^SimpleName value-or-list)}
-                   com.github.javaparser.ast.type.Type {:return-type (.asString ^com.github.javaparser.ast.type.Type value-or-list)}
+                   com.github.javaparser.ast.type.Type {:return-type (.describe (.resolve ^com.github.javaparser.ast.type.Type value-or-list))}
                    nil))
         member (reduce merge {} member)]
     (when-not (contains? (:flags member) :private)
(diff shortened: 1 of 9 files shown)


test/clj_kondo/analysis/java_test.clj
```diff
@@ -105,12 +105,53 @@
...
+          :uri #"file:.*/corpus/java/sources/foo/bar/AwesomeClass.java"
+          :flags #{:public :field :final}
+          :name "bar2"
+          :type "Double"
+          :row 13 :col 5 :end-row 13 :end-col 35}
+         {:class "foo.bar.AwesomeClass"
+          :uri #"file:.*/corpus/java/sources/foo/bar/AwesomeClass.java"
+          :flags #{:public :static :field :final}
+          :name "bar3"
+          :type "Double"
```
borkdude, line 131: Should this type be fully qualified, as it can refer to any class?

```
java.lang.Double
```

src/clj_kondo/impl/analysis/java.clj
```diff
@@ -64,18 +79,52 @@
...
-          (if-let [next-line (read-line)]
-            (if-let [[_ package] (re-matches #"\s*package\s+(\S*)\s*;\s*" next-line)]
-              (str package "." class-name)
-              (recur))
-            class-name))))))
+(defn ^:private node->flag-member-type [node]
+  (condp = (type node)
+    FieldDeclaration :field
+    MethodDeclaration :method
+    ConstructorDeclaration :method))
```
borkdude, line 86: Why is this `:method` and not `:constructor`?

ericdallo: I followed the same rule we had at ASM, call it method and know it by the name, should we change it?

test/clj_kondo/analysis/java_test.clj
ericdallo: Yeah, I'm not sure if java parser does that only for java.langs, but I couldn't think on how to get the full package there, any suggestions?

ericdallo: For that to work, javaparser need to resolve the symbol which is possible via its javaparser-symbol-solver lib, now it's properly working!

src/clj_kondo/impl/analysis/java.clj
ericdallo: From ASM eyes, a constructor is a method as well, we could infer that and make javaparser returns constructor as well, but maybe do in another PR, the way it's is consistent always returning method for both asm and javaparser

borkdude:
@ericdallo I'm afraid the resolver isn't very useful because it relies on runtime analysis.

You can see this when you add something like:

```
import dude.foobar.FooClazz;

    public static FooClazz dude() {
        return null;
    }
```

to `AwesomeClass`. This will fail with:

```
Unsolved symbol : FooClazz
```

The symbol solver stuff is not particularly "static analysis" but uses reflection / javassist and stuff like that, that probably won't be very useful to clj-kondo / clojure-lsp.

As this is a bigger change I think it's best to test-drive this locally for a while, before merging this into clj-kondo, so we can discover issues like the above.

borkdude:
Btw, adding the symbol solver stuff added another 3mb to the image size (increase in total is then 6mb).

ericdallo:
I see, I was afraid something like that could happen, there is the JavaTypeResolver class which you can pass a File to lookup for Java classes, maybe we could pass the root of the jar being analyzed or something, but I agree this make things complex.

I'm wondering how bad this is for clojure-lsp having only the simple class name instead of full package for :type and :return-type, maybe it won't be even necessary as we care about location and available method names, I 'll test locally and confirm

ericdallo:
Good news @borkdude, in the end clojure-lsp won't care if `:type` and `:return-type` are full qualified and this info will be mostly used for hover and completion so having it as simple names is quite enough :)
I made one commit adding the `:method-name` to `:java-class-usages` so now we officially and finally can check docs of methods with clojure-lsp's improve-java-analysis branch 🙌🏻 

image

PR #2076, opened 2023-05-10
Fix #2063: introduce :defined-by->lint-as + always keep original in :defined-by

### LINKED ISSUE #2063: `:defined-by` in analysis output uses `:lint-as` rather than symbol itself

I'm trying to find uses of a particular var, when defined inside another custom type of var.  In this case, a `is` within `defspec` (which isn't valid, and I want to find out how rampant this problem is).

`:defined-by` shouldn't use `nil` for `lint.as/def-catch-all`, nor should it copy the value from `:lint-as` to `:defined-by`.

That is, for:

```clojure
{:lint-as {ns/a clj-kondo.lint-as/def-catch-all ns/b clojure.core/def}}
```

`:defined-by` should be `ns/a` and `ns/b` respectively for any uses.  Rather than `nil` and `clojure.core/def` as it is now.

I'll probably take another stab at solving my problem with grasp and postwalk.

**Additional context**

https://clojurians.slack.com/archives/CHY97NXE2/p1683106125566899

borkdude:


borkdude:
@ericdallo Can you review this and also test on clojure-lsp?

src/clj_kondo/impl/analysis.clj
```diff
@@ -187,6 +189,7 @@
                 :impl-ns impl-ns
                 :filename filename
                 :defined-by defined-by
+                :defined-by->lint-as defined-by->lint-as
```
ericdallo, line 192: This formatting looks odd, too much extra spaces?

ericdallo: There is the same in other places, I suppose you used some find-replace?

ericdallo (approved):
Looks good!
I will only have time to test with LSP in the weekend unfortunately

src/clj_kondo/impl/analysis.clj
borkdude: I'm not sure what you mean

PR #2094, opened 2023-05-30
Fix reflect config for javaparser

ericdallo:

I didn't refactor the tests to use the binary, although I agree should be done, but I manually compiled kondo and tested with the AwesomeClass.java that the tests use.
I guess we can leave the refactor for another PR.

Also, those reflect configs were added based on the output of running `java  -agentlib:native-image-agent=config-merge-dir=... -jar ...` to know what classes were being used.


borkdude:
I think it's better to adapt the test before merging else we'll never be motivated again to add the test.

borkdude:
I pushed a start of such a test to branch ericdallo-fix-javaparser-reflect-config but not fully working yet. Will check tomorrow, but feel free to help.

PR #2110, opened 2023-06-15
analysis: :arglists meta applies to :arglist-strs

### LINKED ISSUE #2096: analysis data: arglists that come from metadata are not necessarily edn friendly and not necessarily content preserving
## version

```
❯ clj-kondo --version
clj-kondo v2023.05.26
```
## Intro
Clj-kondo analysis data, when requested to do so, includes `:arglist-strs`. 
This string encoding of arglists nicely preserves the arglists as written in the code and is edn-friendly.

## Reference
[Discussion on Slack](https://clojurians.slack.com/archives/CHY97NXE2/p1685662451257919?thread_ts=1685661153.172639&cid=CHY97NXE2) with @borkdude and @ericdallo.

## Problem
When using clj-kondo's analysis data, `:arglist-strs` only shows the original arglists.
It does not incorporate `:arglists` specified as metadata. 

This is problematic for tooling that would like to show arglists as documentation as the author of the code intended.  Tooling can always refer to var analysis `:meta` `:arglists`, but this variant is not stringified and therefore does not always preserve arglists as written by the author, and is not necessarily edn-friendly.

## Repro
No `:arglists` metadata, all is good: `(defn foo [orig-arg])` -> `:arglist-strs ["[orig-arg]"]`
```Clojure
❯ clj-kondo --config '{:config-paths ^:replace [] :output {:format :edn :analysis {:var-definitions {:meta true} :arglists true}}}' \
        --lint - <<< $'(defn foo [orig-arg])' \
        | jet --query ':analysis :var-definitions first' --pretty
{:arglist-strs ["[orig-arg]"],
 :col 1,
 :defined-by clojure.core/defn,
 :defined-by->lint-as clojure.core/defn,
 :end-col 22,
 :end-row 1,
 :filename "<stdin>",
 :fixed-arities #{1},
 :meta nil,
 :name foo,
 :name-col 7,
 :name-end-col 10,
 :name-end-row 1,
 :name-row 1,
 :ns user,
 :row 1}
```

But for `(defn foo {:arglists '([new-arg])} [orig-arg])` -> `:arglist-strs ["[orig-arg]"]`.
We do see `:meta {:arglists (quote ([new-arg]))}` but it is not applied to `:arglist-strs`:
```Clojure
❯ clj-kondo --config '{:config-paths ^:replace [] :output {:format :edn :analysis {:var-definitions {:meta true} :arglists true}}}' \
        --lint - <<< $'(defn foo {:arglists \'([new-arg])} [orig-arg])' \
        | jet --query ':analysis :var-definitions first' --pretty
{:arglist-strs ["[orig-arg]"],
 :col 1,
 :defined-by clojure.core/defn,
 :defined-by->lint-as clojure.core/defn,
 :end-col 47,
 :end-row 1,
 :filename "<stdin>",
 :fixed-arities #{1},
 :meta {:arglists (quote ([new-arg]))},
 :name foo,
 :name-col 7,
 :name-end-col 10,
 :name-end-row 1,
 :name-row 1,
 :ns user,
 :row 1}
```

## Expected Behaviour
`:arglist-strs` should always reflect arglists even when overridden by metadata.
We can leave var analysis `:meta` `:arglists` as is.
Because `:arglists` metatada cannot be trusted, analysis data arities values should remain as is.

## Assumption
When overridden by metadata, no analysis data users are interested in original arglists.

#### option a) Always wrap in vector. 
Given `:arglists :fubar` we could encode as `:arglist-strs [":fubar"]`.
This also handles the case where user forgets the wrapping list, ex `:arglists [x y z]` becomes `:arglist-strs ["[x y z]"]`.
#### choice
option a seems like a better choice to me.

### 2) Handling of `^{arglists ...}`
Clojure does not seem to apply `^{:arglists ...}` for a `defn`.
```Clojure
user=> (defn ^{:arglists '([new])} foo [x])
#'user/foo
user=> (meta #'foo)
{:arglists ([x]), :line 1, :column 1, :file "NO_SOURCE_PATH", :name foo, :ns #object[clojure.lang.Namespace 0xbcb09a6 "user"]}
```

But it does do so for a `def`:
```Clojure
user=> (def ^{:arglists '([new])} foo)
#'user/foo
user=> (meta #'foo)
{:arglists ([new]), :line 1, :column 1, :file "NO_SOURCE_PATH", :name foo, :ns #object[clojure.lang.Namespace 0x42deb43a "user"]}
```

#### choice
Uh. Dunno. Probably doesn't matter too much? a?

### 3) Multiple `:arglists` are specified
This would be unusual but we should probably behave like Clojure does.

We could potentially have:
```Clojure
(defn ^{:arglists '([x1])} foo {:arglists '([x2])} ([x]) {:arglists '([x3])})
```
(but see nuance 2 for `^{:arglists ...}`).
As we learned for `:doc`, Clojure applies meta from left to right. So we should do the same for `:arglists`.

#### choice
no options here, just a nuance 

## Next steps 
We need to hash out nuances (and any other feedback), after that, happy to do up a PR if that is helpful.

lread:
Highlights:
- `:arglists` metadata only overrides `:arglist-strs` if it looks valid
- `:fixed-arities` and `:varargs-min-arity` are never affected by `:arglists` metadata, they always describe original arg lists params, if present

Also includes:
- minor analysis doc update & corrections

Closes #2096

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#2105](https://github.com/clj-kondo/clj-kondo/issues/2105): Consider .cljd files when linting.
 - [#2101](https://github.com/clj-kondo/clj-kondo/issues/2101): false positive with `if-some` + `recur`
 - [#2109](https://github.com/clj-kondo/clj-kondo/issues/2109): `java.util.List` type hint corresponds to `:list` or nil
+- [#2096](https://github.com/clj-kondo/clj-kondo/issues/2096): apply `:arglists` metadata to `:arglist-strs` for analysis data ([@lread](https://github.com/lread))
 
 ## 2023.05.26
 
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -21,7 +21,7 @@ Further analysis can be returned by providing `:analysis` with a map of options:
 
 - `:locals`: when truthy return `:locals` and `:local-usages` described below
 - `:keywords`: when truthy return `:keywords` described below
-- `:arglists`: when truthy return `:arglists` on `:var-definitions`
+- `:arglists`: when truthy return `:arglist-strs` on `:var-definitions`
 - `:protocol-impls`: when truthy return `:protocol-impls` described below
 - `:symbols`: when truthy, return `:symbols` described below
 - `:java-class-definitions`: when truthy, return `:java-class-definitions` as described below.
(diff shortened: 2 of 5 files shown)


borkdude:
Thanks!

PR #2156, opened 2023-08-18
support = in arguments

### LINKED ISSUE #2150: Support `=` in options
**version** 

clj-kondo v2023.07.13

**problem**

The behaviour of `--fail-level=error` and `--fail-level error` is different at the moment, which might be surprising since many other CLI tools see them as equivalent.

**expected behavior**

it should either be supported or just error when you pass something that's not expected.

AndreaCrotti:
Fixes #2150

### DIFF
--- a/src/clj_kondo/main.clj
+++ b/src/clj_kondo/main.clj
@@ -83,7 +83,7 @@ Options:
     :scalar))
 
 (defn- parse-opts [options]
-  (let [opts (loop [options options
+  (let [opts (loop [options (mapcat #(str/split % #"=") options)
                     opts-map {}
                     current-opt nil]
                (if-let [opt (first options)]


borkdude:
What's wrong about this approach is that it also splits arguments that contain EDN values, etc.
It's more safe to do it this way:

``` diff
index c62bb105..cf47afde 100644
--- a/src/clj_kondo/main.clj
+++ b/src/clj_kondo/main.clj
@@ -83,17 +83,19 @@ Options:
     :scalar))

 (defn- parse-opts [options]
-  (let [opts (loop [options (mapcat #(str/split % #"=") options)
+  (let [opts (loop [options options
                     opts-map {}
                     current-opt nil]
                (if-let [opt (first options)]
                  (if (starts-with? opt "--")
-                   (recur (rest options)
-                          ;; assoc nil value to indicate opt as explicitly added via cli args
-                          (case (opt-type opt)
-                            :scalar (assoc opts-map opt nil)
-                            :coll (update opts-map opt identity))
-                          opt)
+                   (let [[opt val] (str/split opt #"=" 2)]
+                     (recur (if val (cons val (rest options))
+                                (rest options))
+                            ;; assoc nil value to indicate opt as explicitly added via cli args
+                            (case (opt-type opt)
+                              :scalar (assoc opts-map opt nil)
+                              :coll (update opts-map opt identity))
+                            opt))
                    (recur (rest options)
                           (update opts-map current-opt (fnil conj []) opt)
                           current-opt))
```

borkdude:
Fixed via other PR, thanks!

AndreaCrotti:
Ok great thanks a lot

Issue #2159, opened 2023-08-21
New linter idea: threaded-fn-form


tomdl89:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/clj-kondo/clj-kondo/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

**Is your feature request related to a problem? Please describe.**
I've noticed newcomers to Clojure often don't understand how the threading macros work. As a result, anonymous functions are sometimes put in threading forms, expecting that the function will be applied to the threaded value. For `->` ("thread-first"), what actually happens it that the threaded value is supplied to the "name" slot of the function, and either a function is returned (named with the threaded symbol), or macroexpansion fails (in all other cases, when the threaded value is not a symbol). I've never seen anyone intentionally thread into the name slot of a function, so I think it's fine to treat as a lint warning.

**Describe the solution you'd like**
I'd like functions supplied as one of the `forms` to `->` ("thread-first") to be a lint warning, unless they are "double wrapped".

**Describe alternatives you've considered**
N/A

**Additional context**
I mentioned on #clj-kondo on clojurians slack, and got a couple of upvotes and a message of support. We could also add a lint warning for `->>` ("thread-last"), but maybe this is slightly less rare? Although I think the thread-first case is basically always a mistake, I can just about imagine people threading into the last place of a function body on purpose, wanting to return a function. Still seems pretty rare though.

I've mocked together a PR, which I'll link

borkdude:
I've never encountered this mistake myself, but as you mention, some beginners do. I'll let this idea + PR gather some more upvotes before I merge it. 

One thing to think about is, whether this mistake applies only to `->` and not to other threading macro. So the question is: could this be a more general linter than this very specific situation?

Naming: what naming is used for other threading-related macros? (I'd have to check myself, but consistence would be nice)

Implementation: the detection of `fn` is done in a basic way. It's unlikely but it could happen that people have their own `fn` var that does something completely different, or that `fn` is a local. A perhaps better way to implement this is to do this in the `fn` analysis and check via `(-> ctx :callstack)` if we're directly inside a threading macro. And only in that case, emit the warning.

tomdl89:
Thanks @borkdude.

[...]

PR #2160, opened 2023-08-21
Add threaded-fn-form linter

tomdl89:

- [✓] I have read the [developer documentation](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md).

- [✓] This PR corresponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).
https://github.com/clj-kondo/clj-kondo/issues/2159

- [✓] This PR contains a [test](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#tests) to prevent against future regressions

- [✓] I have updated the [CHANGELOG.md](https://github.com/clj-kondo/clj-kondo/blob/master/CHANGELOG.md) file with a description of the addressed issue.

As mentioned in the issue, I think a thread-last linter could be useful as well, but I could see that it's slightly less likely to be a mistake, so maybe if added it would be off by default?

In terms of implementation, I'm checking the unqualified symbols are `fn` or `fn*` rather than checking their qualified forms, as this didn't seem available to me in the `ctx` or `expr`. Otherwise, I think it works as expected.

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -1369,9 +1369,9 @@ Mismatched bracket: found an opening [ on line 1 and a closing )
 
 ### Threaded function form
 
-*Keyword:* `threaded-fn-form`.
+*Keyword:* `:threaded-fn-form`.
 
-*Description:* warn when directly supplying a function literal as one of the `forms` (n.b. not the first "expression" argument) to the `->` "thread-first" macro. Although this can compile (when a symbol literal is threaded), threading into the `name` slot of an anonymous function is almost always not intended. "Double wrapping" is usually what is intended (e.g. `(-> foo ((fn [x] ...)))`).
+*Description:* warn when directly supplying a function literal as one of the `forms` (n.b. not the first "expression" argument) to the `->` "thread-first" or `some->` macro. Although this can compile (when a symbol literal is threaded), threading into the `name` slot of an anonymous function is almost always not intended. "Double wrapping" is usually what is intended (e.g. `(-> foo ((fn [x] ...)))`).
 
 *Default level:* `warning`.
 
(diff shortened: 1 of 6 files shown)


doc/linters.md
```diff
@@ -1366,6 +1367,18 @@ Mismatched bracket: found an opening [ and a closing ) on line 1
 Mismatched bracket: found an opening [ on line 1 and a closing )
 ```
 
+### Threaded function form
+
+*Keyword:* `threaded-fn-form`.
```
tomdl89, line 1372: This needs a colon. Will add on next pushed commit

doc/linters.md
```diff
@@ -1366,6 +1367,18 @@ Mismatched bracket: found an opening [ and a closing ) on line 1
 Mismatched bracket: found an opening [ on line 1 and a closing )
 ```
 
+### Threaded function form
+
+*Keyword:* `:threaded-fn-form`.
+
+*Description:* warn when directly supplying a function literal as one of the `forms` (n.b. not the first "expression" argument) to the `->` "thread-first" or `some->` macro. Although this can compile (when a symbol literal is threaded), threading into the `name` slot of an anonymous function is almost always not intended. "Double wrapping" is usually what is intended (e.g. `(-> foo ((fn [x] ...)))`).
+
+*Default level:* `warning`.
```
borkdude, line 1376: New linters are usually disabled by default, to incubate while people test it out and not break existing setups for people.

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1627,6 +1627,21 @@
...
+                      (and (list-node? node)
+                           (contains? '#{fn fn*} (:value (first (:children node)))))))
+        fn-nodes (filter fn-node (nnext (:children expr)))]
+    (run! (fn reg-node! [node]
+            (findings/reg-finding!
+             ctx
+             (node->line (:filename ctx)
+                         node
+                         :threaded-fn-form
+                         "threading a value into the name slot of an anonymous fn")))
```
borkdude, line 1642: New lint warnings follow the following convention:

Sentences start with a capital letter

tomdl89:
Closing in favour of #2211

PR #2193, opened 2023-10-05
Add end-row and end-col template variables

### LINKED ISSUE #2192: Support end-row and end-col in :pattern output format
As in

```
clj-kondo --lint src/foo.clj --config \
'{:output {:pattern "{{LEVEL}} {{type}} {{filename}} {{row}} {{end-row}} {{col}} {{end-col}} | {{message}}"}}'
```

Not vital, but it'd be cool.

It'd need to go in config.md#print-results-with-a-custom-format , too. (Which I think is also missing the very useful `{{type}}`.)

I'd happily submit a PR for this one — I assume it's just `impl.core` and ~~wherever the docs live (haven't yet looked)~~ config.md.

joshgelbard:
Document 'type' template variable, too, while we're at it.

fixes #2192


borkdude:
Thank you!

Issue #1753, opened 2022-07-15
Linter: Warn when a namespace use `_` instead of `-` on its name.


ericdallo:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/clj-kondo/clj-kondo/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

**Is your feature request related to a problem? Please describe.**
Today I saw someone facing a error which clj-kondo could probably help, warn when the ns is using `_` instead of `-` like `(ns foo_bar)` instead of `(ns foo-bar)`

**Describe the solution you'd like**
A new linter complaining about that

**Describe alternatives you've considered**
Do nothing

**Additional context**

dharrigan:
Hi,

I was toying with the idea of doing this one. I put together a simple repo here `https://github.com/dharrigan/issue-1753` to see what would happen before doing anything on clj-kondo.

Looks like `foo_bar` is still valid, i.e., in `bar` I can bring in the `foo_bar` namespace and eval (at line 5) to get 3.

Would a warn that `_` in a namespace isn't best practice still be okay?

ericdallo:
I like it

borkdude:
Yeah

borkdude:
I think we should add a rationale to #2209 in the docs: why is it a good idea to avoid `_` in namespace names?

PR #2209, opened 2023-11-04
Implemented no underscore linter - #1753

cosineblast:

This adds a linter `:underscore-in-namespace` which warns on the presence of the `_` character in `ns` declarations, as requested in issue #1753.

This also modifies the `corpus/no_unused_namespace.clj` test sample file so that the namespace is named `no-unused-namespace`  instead of `no_unused-namespace`.

### DIFF
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -550,6 +550,7 @@
                                (str "Namespace name does not match file name: " ns-name)))))))
 
         _ (when (and (not (identical? :off (-> ctx :config :linters :underline-in-namespace :level)))
+                     (symbol? ns-name)
                      (.contains ^String (name ns-name) "_"))
             (findings/reg-finding!
              ctx
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -549,15 +549,15 @@
                                :namespace-name-mismatch
                                (str "Namespace name does not match file name: " ns-name)))))))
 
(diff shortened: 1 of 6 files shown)


borkdude:
@cosineblast Hi! I think this PR mostly looks good. I do prefer the word "underscore" for the `_` character (https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/u/underline-vs-underscore). Can you adapt the name?

About `(.contains ^String ...)` -> I prefer to just use the Clojure `clojure.string/includes?` function for this.

cosineblast:
Done

NoahTheDuke:
Might want to update the docs about this one as well.

src/clj_kondo/impl/analyzer/namespace.clj
```diff
@@ -548,6 +548,17 @@
...
+
+        _ (when (and (not (identical? :off (-> ctx :config :linters :underscore-in-namespace :level)))
+                     (symbol? ns-name)
+                     (str/includes? (name ns-name) "_"))
+            (findings/reg-finding!
+             ctx
+             (node->line filename
+                         ns-name-expr
+                         :underscore-in-namespace
+                         (str "Avoid underscore in namespace name: " (namespace-munge ns-name)))))
```
borkdude, line 560: Why is `namespace-munge` called here?

cosineblast: Oh, my bad, I forgot to remove it when testing the namespace messages. I guess my tests didn't cover this case since they didn't mix `-` and`_` in the namespace names. I'm updating both

borkdude:
Thanks!

PR #2211, opened 2023-11-06
Add invalid-fn-name linter

### LINKED ISSUE #2159: New linter idea: threaded-fn-form

tomdl89:
Fixes #2159 paritially

- [✓] I have read the [developer documentation](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md).

- [✓] This PR corresponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).
https://github.com/clj-kondo/clj-kondo/issues/2159

- [✓] This PR contains a [test](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#tests) to prevent against future regressions

- [✓] I have updated the [CHANGELOG.md](https://github.com/clj-kondo/clj-kondo/blob/master/CHANGELOG.md) file with a description of the addressed issue.

This, effectively supersedes #2160, although as noted in the test file, there is still room for a linter which catches bound vars threaded into fn name position.


src/clj_kondo/impl/analyzer.clj
```diff
@@ -934,6 +934,21 @@
...
+    (not (valid? ?name-expr))))
+
+(defn- reg-invalid-fn-name! [ctx expr filename]
+  (findings/reg-finding!
+   ctx
+   (node->line
+    filename
+    expr
+    :invalid-fn-name
+    "First arg of fn should be a symbol, arg vector or body list")))
```
tomdl89, line 950: "body list" isn't quite right. I'm not sure what we call the (multiple) parenthesized lists... arities?

tomdl89:
I noticed that the `%` arg needed adding to the bindings, as that isn't done by expand fn. I also noticed that`#()` in `->>` forms needs expanding too, and fn* needed adding to checks, in order to surface unsued-value lints which otherwise don't show. So a bit of scope drift, but all related.

tomdl89:
During this PR, I have spotted this false positive, which is interesting. I haven't addressed it, but it may be worth doing:
image
n.b. `hello` is not bound, but doesn't need to be. the warning says it's an unresolved symbol

tomdl89:
Merged master back in so it's easier to see it can be merged. Also updated the error message to be a bit more accurate. I think the only question that needs answering before it can be merged is if it's ok to set the default config for this to be `:error` or `:off`? @borkdude lmk if there's anything else you'd like to see before merging. Thanks!

borkdude:
Hey @tomdl89 ! I appreciate the effort, but I think the PR fixes more than one issue and partially fixes the main issue it was written for? If this PR is only for checking that the name of a `fn` should be simple symbol (unqualified), then I think there is a much simpler PR possible.

tomdl89:
Hey @borkdude yes the PR definitely has some scope creep, no doubt. But I'm fairly sure it isn't otherwise over-engineered.

The only bit of code that checks for invalid fns directly is `invalid-fn-name?` and `reg-invalid-fn-name!` (and the supporting `vector-node?` helper fn).

The rest of the code is expanding function reader literals (i.e. `#(...)` ) before the threading macros as expanded (they weren't before, which means some linting couldn't happen). I can split this out if you would prefer?

That, and tests to support the above work.

Some project prefer separate PRs, but I felt that it was connected enough to warrant one PR. Let me know what you'd like and I'll get on it :)

borkdude:
Yes, let's split it up into tinier issues. With the expanding of fn literals, please give a test case that would otherwise fail with the current order.

borkdude:
@tomdl89 I quickly bashed together a PR to show how I would have done it:

https://github.com/clj-kondo/clj-kondo/pull/2243

Happy to make you co-author of the PR if it gets merged. Feedback welcome.

tomdl89:
Ah, cool. I just extracted mine too. Lemme compare the approaches...

tomdl89:
So, your version is arguably a bit neater, but it does miss things like this:

```clojure
(fn {:foo :bar} [x] (+ x 42))
```

which [mine](https://github.com/tomdl89/clj-kondo/tree/invalid-function-name) catches, because it explicitly says the first arg has to be a symbol, arg vec or fn arity:

```clojure
({:file "<stdin>",
  :row 1,
  :col 1,
  :level :error,
  :message "First arg of fn should be a symbol, params vector or arity clause"})
```

otherwise, good catch on needing to lint it for fns and defns. I'll add that.

tomdl89:
I'll close this to reduce clutter. Can spin out the threading-macro-related functionality in a later PR.

PR #2213, opened 2023-11-09
1990 minimum clj kondo version

### LINKED ISSUE #1990: Specify minimum clj-kondo version in config.edn

Many of our devs rarely update their clj-kondo client. When our config starts employing more recent features, the older clients start to emit warnings that do not make sense to those devs. This has led to some fatigue for clj-kondo, where linter warnings are not taken seriously.

If possible, specifying a `min-version` in the config edn would be nice. Getting a warning about your outdated client will make it clear enough that you need to act on it.

snasphysicist:

This PR aims to resolve #1990

Tests are added in 21f932e1cb09cf5f1f7a0874a8d5c6c238df51df for

- minimum version = current version (no warning)
- minimum version < current version (no warning)
- minimum version > current version (warning)

See 82fb9b625a2b41f86597f0a87dab8102901b9e3e

## Reviewer Notes

I moved `version` out of `impl.core` to avoid circular import issues. I believe `impl.core` transitively depends upon `linters.config`. Happy to move it somewhere else if you prefer.

I put as the default minimum version the first non alpha release. Happy to change this if you prefer.

### DIFF
--- a/src/clj_kondo/impl/config.clj
+++ b/src/clj_kondo/impl/config.clj
@@ -4,6 +4,7 @@
   (:require
    [clj-kondo.impl.findings :as findings]
    [clj-kondo.impl.utils :as utils :refer [deep-merge map-vals]]
+   [clj-kondo.impl.version :as version]
    [clojure.set :as set]
    [clojure.walk :as walk]))
 
@@ -184,7 +185,7 @@
              ;; is appended to the end of the default pattern as " [{{type}}]"
              :linter-name false
              :canonical-paths false} ;; set to true to see absolute file paths and jar files
-    :min-clj-kondo-version "2019.10.26"}) 
+    :min-clj-kondo-version "2019.10.26"})
 
 (defn expand-ignore
   ":ignore true / [:unresolved-symbol] can only be used in
@@ -520,6 +521,36 @@
   (defn deprecated-namespace-excluded? [config required]
     (let [cfg (delayed-cfg config)]
       (contains? cfg required))))
+
+(defn ^:private compare-versions
+  "Returns a finding message if the current version
+   is below the minimum version"
+  [{minimum :minimum
+    current :current}]
+  (let [earlier-version (fn
+                          [v1 v2]
+                          (first (sort [v1 v2])))]
+    (when
+     (not=
+      minimum
+      (earlier-version
+       current
+       minimum))
+      (str
+       "Version "
+       current
+       " below configured minimum "
+       minimum))))
+
+(defn check-minimum-version
+  "Prints a warning if the version is below the configured minimum"
+  [ctx]
+  (let [minimum-version (-> ctx :config :min-clj-kondo-version)
+        warning (when minimum-version
+                  (compare-versions {:minimum minimum-version
+                                     :current version/version}))]
+    (when warning
+      (println "[clj-kondo] WARNING:" warning))))
 
 ;; (defn ns-group-1 [m full-ns-name]
 ;;   (when-let [r (:regex m)]
--- a/src/clj_kondo/impl/config.clj
+++ b/src/clj_kondo/impl/config.clj
@@ -157,7 +157,8 @@
               :protocol-method-varargs {:level :error}
               :unused-alias {:level :off}
               :self-requiring-namespace {:level :off}
-              :condition-always-true {:level :off}}
+              :condition-always-true {:level :off}
+              :underscore-in-namespace {:level :warning}}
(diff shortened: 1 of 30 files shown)


src/clj_kondo/impl/linters/config.clj
```diff
@@ -47,6 +48,58 @@
                      :clj-kondo-config
                      (str "Linter config should go under :linters")))))))
 
+(defn ^:private minimum-version-node
```
borkdude, line 51: I would move the processing of `:min-clj-kondo-version` to just the `impl/config.clj` namespace or so, or even `impl/core.clj`. This namespace is about verifying the `config.edn` file for syntactical errors.

src/clj_kondo/impl/config.clj
```diff
@@ -183,7 +183,8 @@
              ;; if below :linter-name is set to true, type (linter name) of reported the finding
              ;; is appended to the end of the default pattern as " [{{type}}]"
              :linter-name false
-             :canonical-paths false}}) ;; set to true to see absolute file paths and jar files
+             :canonical-paths false} ;; set to true to see absolute file paths and jar files
+    :min-clj-kondo-version "2019.10.26"})
```
borkdude, line 187: Why have this as the minimal clj-kondo version? I'd just comment this out, having no minimum clj-kondo version

src/clj_kondo/impl/linters/config.clj
borkdude: One more reason to not have the check here is that configurations might be merged and version can be overridden. As such we should only look at the effective config and not at a single config file.

snasphysicist: That's reasonable - I didn't realise that the linting here would apply only to the file.

This does raise a question though of how to register findings  not associated with a file. When registering a finding, we have to give a map containing `:filename`, `:row` and `:col`. `:row` and `:col` might make sense from configuration from any source, but `:filename` may not, e.g. a configuration given on the command line. I'm also not sure that it will be clear to this code where the `:min-clj-kondo-version` came from, i.e. was it a `config.edn`, was it from the command line, etc... Any thoughts on how to handle such a case?

src/clj_kondo/impl/config.clj
snasphysicist: Happy to comment this out, was just looking for a sensible default value. I'll make these changes when I hear back on the other review comment.

src/clj_kondo/impl/linters/config.clj
borkdude: That's a good point! Maybe this tips the balance in favor of just printing to stderr?

snasphysicist: Yes, thinking about it a bit more, this would also require (probably substantial) changes to config reading to track provenance (e.g. to know if the minimum version came from a file, from the command line, and in which row/col of the appropriate one) as well as changes to interfaces around findings - probably not worth it for this feature.

Also good to note that it's not too difficult to reverse this decision in future if `clj-kondo` changes in such a way that it makes more sense to have this check as a finding.

I've drafted up this approach in 9e5e395c897e9b67e2c6d187be3eed654ea4e89e & 475ce63b92060e428571eb1f657b2e51812c2424 - let me know what you think!

src/clj_kondo/impl/config.clj
snasphysicist: Fixed in 14cd00636de4c7fe531644a4edf198ae66241c98

src/clj_kondo/impl/linters/config.clj
snasphysicist: I'll have to look later this week into the test failures. It passes locally, I haven't yet figured out why it fails in CI.

borkdude:
I'll review this PR soon-ish hopefully

snasphysicist:
Thanks! No rush, I'm trying some other methods to capture what would be printed to stderr, which haven't worked in CI yet. I'll also need to merge in master.

snasphysicist:
Okay, I think I figured this out. I saw that in other tests that probe stderr, they are only run for the "non native" cases (I assume this means the JVM tests, not the tests that use binaries). I updated the tests for the minimum version message only to be run `when-not native?`, so I think the testing problem is solved. I would suggest to ignore 1bba512cf4ebe78209c59ec3f4dcc6408b561eb8 to 06063b75a059712af26dfb4d9c6a17b7b111ace6 where I explored some dead ends to try to get around this (the last commit there reverts the changes in the earlier ones).

borkdude:
This looks great, thank you!

PR #2222, opened 2023-11-28
Fix #2215: passthrough hook should not affect linting

### LINKED ISSUE #2215: `:redundant-let` linter not working inside `deftest` when it has a custom hook
**version**

```
clj-kondo v2023.09.07
```

**platform**

Native

**problem**

We have a custom `:analyze-call` hook defined for `deftest` that looks like this:

```clj
(defn deftest  [{:keys [node], :as arg}]
  ;; you can actually comment this line out and the problem still happens
  (deftest-check-not-horrifically-long node)
  arg)
```

With this config, the `:redundant-let` linter does not work inside `deftest` any more:

image

**repro**

```clj
;; .clj-kondo/config.edn
{:hooks {:analyze-call {clojure.test/deftest hooks.clojure.test/deftest}}}
```

```clj
;; .clj-kondo/hooks/clojure/test.clj
(ns hooks.clojure.test)

(defn deftest [arg]
  arg)
```

```clj
;; test/fail.clj
(ns fail
  (:require [clojure.test :refer [deftest]]))

(deftest my-test
  (let [x 1]
    (let [y 2]
      (assert (not=  x y)))))
```

```sh
# output
clj-kondo --lint test/fail.clj
linting took 7ms, errors: 0, warnings: 0
```

**expected behavior**

Should still warn about redundant-let even if a custom hook is in use

borkdude:
Fixes #2215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#1753](https://github.com/clj-kondo/clj-kondo/issues/1753): new linter `:underscore-in-namespace`
 - [#2207](https://github.com/clj-kondo/clj-kondo/issues/2207): New `:condition-always-true` linter, see [docs](doc/linters.md)
 - [#2013](https://github.com/clj-kondo/clj-kondo/issues/2013): Fix NPE and similar errors when linting an import with an illegal token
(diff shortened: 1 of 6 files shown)


borkdude:
TODO: add changelogs

PR #2233, opened 2023-12-05
Fix typos in linters.md

mrkam2:
Fix a few typos.


borkdude:
These aren't strictly typos since Clojure accepts this notation, but I think it's better to go with the mainstream notation, thanks.

mrkam2:
> These aren't strictly typos since Clojure accepts this notation, but I think it's better to go with the mainstream notation, thanks.

You're right, I did a poor job checking. Thanks for merging!

PR #2243, opened 2023-12-15
Lint invalid fn name

borkdude:


tomdl89:
Having taken a look, I'm not sure what you meant by:
> What is missing from this PR (and from mine) is the check for (fn foo/bar []): the error should emphasize that the symbol should be simple
As I think this PR does do that, right?

borkdude:
Yes, indeed, sorry for the confusion

PR #2244, opened 2023-12-15
Add invalid-fn-name linter

tomdl89:
Alternative to #2243

Covers cases such as this:

```clojure
(fn {:foo :bar} [x] (+ x 42))

({:file "<stdin>",
  :row 1,
  :col 1,
  :level :error,
  :message "First arg of fn should be a symbol, params vector or arity clause"})
```

Spun out of #2211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
-- ...
+- [#2159](https://github.com/clj-kondo/clj-kondo/issues/2159): New linter, `:invalid-fn-name`
 
 ## 2023.12.15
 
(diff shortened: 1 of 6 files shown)


borkdude:
Any reason we shouldn't just lump this together with other smaller `:syntax` checks? 

#2243 checks whether a name is a simple symbol (simple symbol means that it is not qualified, as checked with `simple-symbol?`), not whether the fn syntax is valid. With `(fn {})` I think it's pretty clear that the user didn't intent to write `{}` as the function's name and I would expect a different error message (possible also under the `:syntax` umbrella).
What is missing from this PR (and from mine) is the check for `(fn foo/bar [])`: the error should emphasize that the symbol should be _simple_.

tomdl89:
Yeah, fair point - it does seem to better suit a syntax check. As you said, no one's going to intend `{}` to be a function name. I'll close this and look at adding the check to #2243 if I can.

PR #2253, opened 2024-01-04
Add `datomic-type-extensions.api/q` to datalog syntax checking

magnars:
This change allows clj-kondo to also lint datalog queries when wrapping datomic.api with type extensions:

https://github.com/magnars/datomic-type-extensions

This commit also includes new tests for all the different datalog libraries.

- [ ] This PR does not correspond to an issue. I considered it a quick fix. I'll be happy to open an issue if necessary.

- [ ] I haven't updated the changelog, again because of the small size and scope of the change.


magnars:
That is already possible :) 

```clj
{:lint-as {datomic-type-extensions.api/q datomic.api/q}}
```

borkdude:
And then linting works?

magnars:
Yes, that seems to do the trick. The same can be said for all the other libs in the list.

borkdude:
Alright then ;)

magnars:
Thanks 🙏😊

Issue #1923, opened 2022-12-20
Check defn names for simple symbol without dots


borkdude:
Bad:

``` clojure
(defn foo.bar []) ;; not enforced by spec, but `(foo.bar)` doesn't work, although `(user/foo.bar)` does
(defn foo/bar []) ;; enforced by spec
```

See https://github.com/babashka/sci/issues/811

PR #2282, opened 2024-02-18
Expand fn literals at threading macroexpand time

tomdl89:

- [x] This PR corresponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code). ([1923](https://github.com/clj-kondo/clj-kondo/issues/1923))

---

This PR takes the functionality from #2211 which wasn't implemented in #2243.
Namely the work to thread into fn literals (`#()` forms) so invalid names and unused values can be detected.

Before this PR, the following code would not raise a linting error:
```clojure
(-> :foo #(name %))
```
But after it, it raises a "Function name must be simple symbol but got: :foo" error.
This makes sense because the macroexpansion of the above form looks like so:
```clojure
(macroexpand '(-> :foo #(name %)))
;; => (fn* :foo [p1__23143#] (name p1__23143#))
```

Also, before this PR, the following code would not raise a linting error:
```clojure
(->> :foo #(name %))
```
But after it, it raises an "Unused value" error.
This makes sense because the macroexpansion of the above form looks like so:
```clojure
(macroexpand '(->> :foo #(name %)))
;; => (fn* [p1__23147#] (name p1__23147#) :foo)
```

The PR is split into two commits, which should be squashed:

1. Expand fn literals at threading macroexpand time
This is the functional changes:
a. Add separate treatment for `:fn` forms vs other non-lists in `expand->` and `expand->>`
b. Move `%` anaphoric argument handling earlier in `analyze-fn`
c. Add `fn*` to the list of core-syms checked by `analyze-usages2` and `lint-var-usage` (as this is what fn literals expand to)
2. Move expand-fn code & add tests
I think it's clearer to move the `expand-fn` code and its dependencies as a separate step, to show that nothing in it is changed. Added tests & updated the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ For a list of breaking changes, check [here](#breaking-changes).
 <!-- - [ ] update lein-clj-kondo -->
 <!-- - [ ] update carve -->
 
+## Unreleased
+- [#1923](https://github.com/clj-kondo/clj-kondo/issues/1923): Lint invalid fn name for threaded fn literals
+
 ## 2024.02.12
 
 - [#2274](https://github.com/clj-kondo/clj-kondo/issues/2274): Support clojure 1.12 new type hint notations
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -957,9 +957,11 @@
     "Use defn instead of def + fn")))
 
 (defn analyze-fn [ctx expr]
-  (let [ctx (assoc ctx :seen-recur? (volatile! nil))
-        protocol-fn (:protocol-fn expr)
-        ctx (assoc ctx :protocol-fn protocol-fn)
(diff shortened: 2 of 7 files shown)


borkdude:
It seems kind of arbitrary to expand a fn literal specifically in `->` or `->>` (only such that it gives a better message in some cases). I'll have to think a bit more about this.

tomdl89:
Thanks for taking a look @borkdude. Fair point that it's arbitrary to expand fn literals only in `->` and `->>`. I've worked on a new approach which expands the literals earlier in the call stack: before the enclosing list is processed. This means values can be added to the `fn*` body by any macro which takes the fn literal as a direct child, which covers both these threading macros, and any others like `cond->`. Let me know if that's what you were thinking of: https://github.com/clj-kondo/clj-kondo/compare/master...tomdl89:clj-kondo:fn-literal-earlier-expansion

PR #2298, opened 2024-03-05
Fix #1732: warn on shadowed fn param

### LINKED ISSUE #1732: Misleading warning in case of duplicate symbols in function's parameters declaration
**version**

clj-kondo v2022.05.31

**platform**

native on Mac OS and JVM version

**problem**

When by mistake the function's parameter declaration form contains identical symbols in different positions clj-kondo reports about "Unused binding" referring to all but the last such symbols. Clojure does not complain about that and the function will be compiled correctly but potentially it might cause problems which kondo can highlight. However, the current message might bring some confusion.

```clojure
(defn foo [x x] ;; => warning: unused binding x
  (inc x))
```

The same problem happens with destructuring used together with `:as` directive.

Original discussion begins in [slack thread](https://clojurians.slack.com/archives/C053AK3F9/p1655839370235429)

**repro**

```shell
echo '(defn foo [x x] (+ x x))' | clj-kondo --lint -
```

**expected behaviour**

The message could be different stating that there are identical bindings in parameters which will shadow each other.

I'm interested in contributing a PR if this issue seems relevant.

borkdude:


src/clj_kondo/impl/config.clj
```diff
@@ -47,6 +47,7 @@
                                :exclude-destructured-keys-in-fn-args false
                                :exclude-destructured-as false
                                :exclude-defmulti-args false}
+              :shadowed-fn-param-name {:level :warning}
```
NoahTheDuke, line 50: Looks like this should be `:shadowed-fn-param`.

borkdude: You're right, I'll fix it

PR #2333, opened 2024-05-26
Report locations for more reader errors

### LINKED ISSUE #2322: Location missing on literal syntax errors

**version**

clj-kondo v2024.03.13

**platform**

native

**editor**

NA

**problem**

The following syntax errors cause clj-kondo to report an null / incorrect error location and interfere with the reporting of all other linting errors in the file. (this makes it trickier than usual to pinpoint the source of the error)

**repro**

```bash
$ echo '\u12345' | clj-kondo --lint -
<stdin>::: error: Invalid unicode literal: \u12345.
linting took 41ms, errors: 1, warnings: 0

$ echo '\o8' | clj-kondo --lint -
<stdin>::: error: Invalid digit 8 in unicode character \o8.
linting took 66ms, errors: 1, warnings: 0
```
Note the missing line and column locations (this confuses the Emacs flycheck checker)
Also `\o` denotes an octal, not unicode escape.

```bash
$ echo '##' | clj-kondo --lint -
<stdin>:1:3: error: EOF while reading.
linting took 47ms, errors: 1, warnings: 0

$ echo '99r0' | clj-kondo --lint -
<stdin>:1:: error: Radix out of range
linting took 32ms, errors: 1, warnings: 0

$ echo "oops\n\n2r123" | clj-kondo --lint -
<stdin>:1:: error: For input string: "123"
linting took 73ms, errors: 1, warnings: 0
```

In these cases the line number 1 is reported regardless of where the error occurs, and other lint warnings (eg. unresolved symbol) before or after the parsing errors are not reported.
Note too the incomplete error message in the last line.

**config**
NA

**expected behavior**

clj-kondo should report the line/column location of these syntax errors, similar to what it does in other similar situations:
```bash
$ echo "oops\n1.0/3\n\\spcae" | clj-kondo --lint -
<stdin>:1:1: error: Unresolved symbol: oops
<stdin>:2:1: error: Invalid number: 1.0/3.
<stdin>:3:1: error: Unsupported character: spcae.
linting took 71ms, errors: 3, warnings: 0
```

yuhan0:
Fixes #2322
As mentioned in https://github.com/clj-kondo/clj-kondo/issues/2322#issuecomment-2132237967 I added an additional clause to check for `:eof` reader-exception types, which surfaces the correct line numbers for invalid literals such as `##`.

The last commit removes those checks altogether (which I felt was unnecessary and brittle) and refactors the logic for better readability. (ie. subjective, can be left out)

Invalid-radix number literals, also mentioned in the original issue, seem to have a deeper root cause in the inlined tools.reader code which might be better reported upstream.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 ## 2024.05.24
 
 - Imports were copied to `.clj-kondo/imports` but weren't pick up correctly. Thanks [@frenchy64](https://github.com/frenchy64) for reporting the bug.
+- [#2333](https://github.com/clj-kondo/clj-kondo/issues/2333): Add location to invalid literal syntax errors
 
 ## 2024.05.22
 
--- a/corpus/invalid_characters.clj
+++ b/corpus/invalid_characters.clj
@@ -1,0 +1,7 @@
+(ns invalid-characters)
+
+\u12345
+\uxyz
+\o12345
+\o400
+\spcae
--- a/corpus/invalid_literals.clj
+++ b/corpus/invalid_literals.clj
@@ -1,0 +1,10 @@
+(ns corpus.invalid-literals)
+
+:
+::
+:foo:
+foo:
+foo/
+:foo/
+##
+##NAN
--- a/corpus/invalid_unicode.clj
+++ b/corpus/invalid_unicode.clj
@@ -1,3 +1,0 @@
-(ns invalid-unicode)
-
-\u12345
--- a/parser/clj_kondo/impl/rewrite_clj/parser/token.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/parser/token.clj
@@ -52,16 +52,11 @@
           (symbol-node reader v s)
           (node/token-node v s)))
       (catch Exception e
-        (if r/*reader-exceptions*
-          (do (let [{:keys [:type :ex-kind]} (ex-data e)]
-                (if (and (= :reader-exception type)
-                         (or (= :reader-error ex-kind)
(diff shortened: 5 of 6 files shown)


yuhan0:
The first commit's change was due to *print-length* being set to 100 in my local CIDER setup, which caused lots of tests to fail with confusing test errors ("map must contain even number of forms") - I eventually tracked it down to the huge `clj-kondo.impl.config/default-config` map being truncated with a stray `...` when being stringified and read back via `main`.

Not sure if there's a more suitable place further up the call stack to wrap with the `binding` form, do let me know if I should amend or remove it from this PR.

borkdude:
I prefer the binding print-length to nil not be part of this PR, as this seems a local dev setup problem. I've never personally encountered this, but if it's an actual problem, it's a separate issue with a repro + PR I'd say

borkdude:
The rest of the PR seems good to me

yuhan0:
Ok, I've removed the commit and added a couple more examples of invalid chars to the corpus.

yuhan0:
Side note: I noticed the inconsistent spacing in the error message for the invalid octal character, which it turns out was fixed in upstream clojure.tools.reader a few years ago: https://github.com/clojure/tools.reader/commit/0b984f95ba9f7e89afb441b4995e27c87b378a4d

Which made me wonder if there have been more important fixes/changes made upstream since it was forked back in 2019? The patches were made in a [monolithic commit](https://github.com/clj-kondo/clj-kondo/pull/296/commits/a11417ab116c3131defa2c2e59d0c9c3aa79e5f2) so it's somewhat difficult to understand how significantly it's diverged from the main repo.

borkdude:
Porting any improvements that actually matter to the forked version of tools.reader are welcome.

PR #2334, opened 2024-05-28
Analyse types of do and doto expressions

yuhan0:
I found these changes that I made some time ago while poking around the type analyzer, figured they would be useful :)

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -916,7 +916,7 @@
       (findings/reg-finding!
        ctx
        (node->line filename expr :redundant-do "redundant do"))))
-  (analyze-children ctx (next (:children expr))))
+  (analyze-children ctx (next (:children expr)) false))
 
 (defn lint-two-forms-binding-vector! [ctx form-name expr]
   (let [num-children (count (:children expr))]
--- a/src/clj_kondo/impl/macroexpand.clj
+++ b/src/clj_kondo/impl/macroexpand.clj
@@ -83,14 +83,16 @@
              (with-meta x (assoc (meta x) :clj-kondo.impl/generated true)))
         ret (list-node
              (list* (token-node 'clojure.core/let) (vector-node [gx x])
-                    (map (fn [f]
-                           (with-meta-of
-                             (let [t (tag f)]
-                               (if (= :list t)
-                                 (let [fc (:children f)]
-                                   (list-node (list* (first fc) gx (next fc))))
-                                 (list-node [f gx])))
-                             f)) forms)))]
+                    (conj
+                     (mapv (fn [f]
+                             (with-meta-of
+                               (let [t (tag f)]
+                                 (if (= :list t)
+                                   (let [fc (:children f)]
+                                     (list-node (list* (first fc) gx (next fc))))
+                                   (list-node [f gx])))
+                               f)) forms)
+                     gx)))]
     ret))
 
 (defn expand-dot-constructor
--- a/src/clj_kondo/impl/types/clojure/core.clj
+++ b/src/clj_kondo/impl/types/clojure/core.clj
@@ -690,7 +690,8 @@
                       3 {:args [:vector :nat-int :nat-int]
                          :ret :vector}}}
    ;; 3831 'with-open
-   ;; 3852 'doto
+   ;; 3852 
+   'doto {:fn first}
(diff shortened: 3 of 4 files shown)


NoahTheDuke:
The "regression" comes from the deliberate discarded value in [read-discard](https://github.com/clojure/tools.reader/blob/0d3a7cd946e0ef377b7fe2c85df6cf940ff39c81/src/main/clojure/clojure/tools/reader/edn.clj#L299-L302):

```clojure
(defn- read-discard
  [rdr _ opts]
  (doto rdr
    (read true nil true)))
```

yuhan0:
That's a strange one... I shrunk it down to this minimal repro:

```clj
(ns clojure.tools.reader.edn
  (:refer-clojure :exclude [read]))

(declare read)
(doto nil (read))
```
The "unused value" warning only appears when the ns is specifically `clojure.tools.reader.edn`, changing it to any other name causes it to go away. Maybe something to do with the cache? The actual `read` function defined below does lots of side-effecting things like throwing exceptions, but somehow the analyser is treating it as a pure one.

Which also brings up the question, should this be marked with an unused-value warning:

`(doto 1 (inc))`

The master branch does not, but this PR analyses the doto 'more accurately', allowing the analyser to see that the pure function `inc` is being used in an unreachable position.

Not sure how to continue debugging from here

yuhan0:
Note that `(doto 1 inc inc)` produces a warning on the master branch, the difference is that the analyser now appends the `1` at the end of the doto expansion so it recognizes the first `inc` as being unused.



Maybe it's reasonable to assume that every function call in a `doto` is there for side-effecting reasons (that the user 'knows what they're doing')? 

Either way it seems unrelated to the PR.. here's a command line version of the above repro

```bash
$ echo '(ns clojure.tools.reader.edn) (declare read) (doto nil (read))' | ./clj-kondo --lint -
<stdin>:1:56: warning: Unused value
linting took 9ms, errors: 0, warnings: 1
```

For some reason I could not reproduce it within a deftest, eg. this passes:

```clj
(deftest weird-test
  (assert-submaps ()
   (lint!
    "(ns clojure.tools.reader.edn) (declare read) (doto nil (read))"
    {:linters {:unused-value {:level :warning}
               :type-mismatch {:level :warning}}})))
```

borkdude:
I thought it might be related to this one: https://github.com/clj-kondo/clj-kondo/pull/2334/files/eb75edc95d1f4be2ebb1a0432d66c70f66ee56b2#diff-f591b825c81f488c7fa9426a07ba9dd30f948cc52e445be929d5fd8299bc7dde

but I'd have to take a look.

With clj-kondo master I get:

```
$ echo '(ns clojure.tools.reader.edn) (declare read) (doto nil (read))' | clj-kondo --lint -
linting took 42ms, errors: 0, warnings: 0
```

> Either way it seems unrelated to the PR

That seems a bold claim since it doesn't occur without this PR ;)

borkdude:
What issue number does this PR correspond to btw? Maybe it's better to split the PR up into two parts. The one for `do` which seems trivial and then another one for `doto` on top of that.

yuhan0:
Here's the false positive observed on the master branch:

```clj
(do
  (require '[clojure.tools.reader.edn :as ctr.edn])
  (ctr.edn/read) ;; <--
  :ok)
```

```bash
$ echo '(do (require \'[clojure.tools.reader.edn :as ctr.edn]) (ctr.edn/read) :ok)' | clj-kondo --lint  -
<stdin>:1:55: warning: Unused value
linting took 9ms, errors: 0, warnings: 1
```

There isn't a corresponding issue for this PR, do you need one to be opened?

borkdude:
Alright, since there isn't an issue corresponding to this PR, changes aren't thought through properly (relying on me to guess why things have changed), etc, I'm inclined to close this PR. Feel free to create separate issues for each problem that was identified and optionally a PR.

yuhan0:
Sure, sorry about that - will open corresponding issues later for these.

borkdude:
No problem, I just find that easier to track/reason about. Thanks for the help!

borkdude:
I found the issue with clojure.tools.reader.edn/read:

https://github.com/jonase/eastwood/issues/451

borkdude:
Submitted separate PRs for do and doto, which fixes all of the above issues and made @yuhan0 co-author since he already wrote most of the code already in this PR.

PR #2411, opened 2024-10-14
Add new `--report-level` CLI option

### LINKED ISSUE #2410: Add new `--report-level` cli flag; particularly useful for CI

We run clj-kondo as a CI job and fail the build if there are any warnings or errors. We have some `info` level diagnostics that clutters the CI output and makes it harder than necessary for people investigating to spot the errors / warnings they have to fix among all the `info` level noise.  (One of our projects has ~650 Clojure files, and over 120kLOClj - so a couple infos per file adds up.) 

Yet we don't want to turn these off in config, as they're useful information (for some of us, at least) when programming.

_A clear and concise description of what you want to happen._

It would be nice if clj-kondo had a `--report-level` that we could set to the same level as `--fail-level`, so we would not have to play whack-a-mole with new info-level diagnostics in the `.clj-kondo/ci-overrides.edn` file.

_A clear and concise description of any alternative solutions or features you've considered._

I added a separate `.clj-kondo/ci-overrides.edn` that I specify with `--config .clj-kondo/ci-overrides.edn` for the ci job, and change the reporting-level for the noisy info to `:off` there. But that still means we have to update this every time a new linter is added/enabled/starts "violating" to avoid it showing up in the output.

**Additional context**

[ Optional: if you are interested in doing a PR yourself, please leave a note. ]

stig:
This is useful for CI where you may not want to display info-level notices if you only require folks to fix warnings or above.

The default level is `info`.

With `--report-level error` will also omit the `, warnings: N` part of the output summary.

If the report level is not recognised, prints the help.

Fixes #2410

### DIFF
--- a/src/clj_kondo/core.clj
+++ b/src/clj_kondo/core.clj
@@ -24,12 +24,8 @@
   [{:keys [:config :findings :summary :analysis :report-level]}]
   (let [output-cfg (:output config)
         report-level? (if report-level
-                        (let [report-level (some-> report-level keyword)]
-                          (if-let [levels (->> [:info :warning :error]
-                                               (drop-while #(not= report-level %))
-                                               seq)]
-                            (set levels)
-                            (constantly true)))
+                        (let [report-level (keyword report-level)]
+                          (set (drop-while #(not= report-level %) [:info :warning :error])))
                         (constantly true))
         fmt (or (:format output-cfg) :text)]
     (case fmt
--- a/src/clj_kondo/core.clj
+++ b/src/clj_kondo/core.clj
@@ -23,10 +23,12 @@
   subject to change."
   [{:keys [:config :findings :summary :analysis :report-level]}]
   (let [output-cfg (:output config)
-        report-level (some-> report-level keyword)
-        report-level? (or (set (drop-while #(not= report-level %)
-                                           [:info :warning :error]))
-                          (constantly true))
+        report-level? (let [report-level (some-> report-level keyword)]
+                        (if-some [levels (->> [:info :warning :error]
+                                              (drop-while #(not= report-level %))
+                                              seq)]
+                          (set levels)
+                          (constantly true)))
         fmt (or (:format output-cfg) :text)]
     (case fmt
       :text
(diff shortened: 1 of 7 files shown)


src/clj_kondo/core.clj
```diff
@@ -21,8 +21,10 @@
 (defn print!
   "Prints the result from `run!` to `*out*`. Returns `nil`. Alpha,
   subject to change."
-  [{:keys [:config :findings :summary :analysis]}]
+  [{:keys [:config :findings :summary :analysis]} report-level]
```
borkdude, line 24: Since this is a public API function adding another argument would be a breaking change to existing caller. Just make :report-level another option in the map passed as the argument.

src/clj_kondo/core.clj
```diff
@@ -21,8 +21,11 @@
 (defn print!
   "Prints the result from `run!` to `*out*`. Returns `nil`. Alpha,
   subject to change."
-  [{:keys [:config :findings :summary :analysis]}]
+  [{:keys [:config :findings :summary :analysis :report-level]}]
   (let [output-cfg (:output config)
+        report-level (keyword report-level)
+        report-level? (set (drop-while #(not= report-level %)
```
borkdude, line 27: I think when report-level isn't set, then we can just make this function `(constantly? true)` to be on the safe side when people have added custom levels

borkdude: Also `(keyword nil)` should probably be prevented.

src/clj_kondo/core.clj
```diff
@@ -23,10 +23,12 @@
   subject to change."
   [{:keys [:config :findings :summary :analysis :report-level]}]
   (let [output-cfg (:output config)
-        report-level (some-> report-level keyword)
-        report-level? (or (set (drop-while #(not= report-level %)
-                                           [:info :warning :error]))
-                          (constantly true))
+        report-level? (let [report-level (some-> report-level keyword)]
```
borkdude, line 26: can you do an earlier check on report-level so we don't have to do all this work when report-level is nil

stig: Yes, but it adds some duplication. Let me try that.

stig:
Want me to mention somewhere that custom log levels will be hidden if you specify `--report-level $level`?

Or is it possible to do the right thing in that case?

PR #2440, opened 2024-11-22
Fix uneval on reader conditionals

### LINKED ISSUE #2439: Uneval nodes consuming reader conditional that doesn't have matching branch

**version**

clj-kondo v2024.11.14

**macro usage**

N/A

**platform**

Native (but applies to all)

**editor**

N/A

**problem**

(Both Clojure and Clojurescript share the same behavior, so I'll just refer to Clojure here.)

The uneval/discard reader macro `#_` requires a value to be returned (to discard). Reader conditionals produce a value from the matching branch or consume the whole form and return just the reader moved past the conditional. In combination, this means that `#_#?(:foo 1) 2 3` produces `3`, not `2`. However, clj-kondo's uneval consumes the reader conditional, returning `2 3`.

**repro**

```
$ clj
Clojure 1.11.1
user=> (+ 1 #_#?(:foo 2) "three" 4)
5

$ echo '(+ 1 #_#?(:foo 2) "three" 4)' | clj-kondo --lint -
<stdin>:1:19: warning: Expected: number, received: string.
linting took 10ms, errors: 0, warnings: 1
```

Using a vector to return all objects:

```clojure
[#_#?(:foo 1) 2 3]
; [3]

(:children (clj-kondo.impl.parser/parse-string "[#_#?(:foo 1) 2 3]"))
; ({:tag :vector,
;   :format-string "[%s]",
;   :wrap-length 2,
;   :seq-fn #function[clojure.core/vec],
;   :children
;   ({:value 2, :string-value "2"}, {:value 3, :string-value "3"})})
```

**config**

N/A

**expected behavior**

clj-kondo correctly consumes reader conditionals, only treating them as values if there's a matching branch, and discarding/uneval-ing the correct nodes.

NoahTheDuke:
Closes #2439

Situation:

Reader conditionals are consumed by the Clojure reader. If there's a matching branch, it's returned. If there's no matching branch, the reader moves on. Discard reader macros consume the next form. So when the discard macro attempts to get the next form and there's a reader conditional with no matching branch, the form _after_ the reader conditional is returned.

* `(+ 1 #_#?(:clj 2) :three 4)`
  * Read as `(+ 1 #_ 2 :three 4)`. The `:clj` branch of the `#?` form is used so the next form is `2`, `#_` ignores it, the result is `(+ 1 :three 4)`.
* `(+ 1 #_#?(:foo 2) :three 4)`
  * Read as `(+ 1 #_ :three 4)`. The `:clj` branch of the `#?` form doesn't exist so the next form is `:three`, `#_` ignores it, the result is `(+ 1 4)`.
* `(+ 1 #_#?(:clj 2 :default 100) :three 4)`
  * Read as `(+ 1 #_ 2 :three 4)`. The `:clj` branch of the `#?` form is used so the next form is `2`, `#_` ignores it, the result is `(+ 1 :three 4)`.
* `(+ 1 #_#?(:cljs 2 :default 100) :three 4)`
  * Read as `(+ 1 #_ 2 :three 4)`. The `:default` branch of the `#?` form is used so the next form is `100`, `#_` ignores it, the result is `(+ 1 :three 4)`.

Approach:

During parsing, if an uneval/discard reader macro is encountered, check if the next node is a reader conditional and if it contains all of the branches of the current reader features (tracked in a new dynamic variable `reader/*reader-features*`) or `:default`. If so, mark it with `:clj-kondo/uneval` and return. Otherwise, skip as normal.

During analysis, in `select-lang`, if a form with `:clj-kondo/uneval` metadata is encountered then:
* If there's a matching branch in the reader conditional, `#_` would discard it, so skip it.
* If there's no matching branch in the reader conditional, `#_` would discard the form _after_ the reader conditional, so skip 2 forms.

This is slightly awkward because parsing doesn't pass around a `ctx` object giving access to which lang we're in (and we don't want to parse twice).

---

### DIFF
--- a/parser/clj_kondo/impl/rewrite_clj/parser/core.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/parser/core.clj
@@ -6,7 +6,8 @@
    [clj-kondo.impl.rewrite-clj.parser.string :refer [parse-string parse-regex]]
    [clj-kondo.impl.rewrite-clj.parser.token :refer [parse-token]]
    [clj-kondo.impl.rewrite-clj.parser.utils :as u]
-   [clj-kondo.impl.rewrite-clj.reader :as reader]))
+   [clj-kondo.impl.rewrite-clj.reader :as reader]
+   [clojure.string :as str]))
 
 ;; ## Base Parser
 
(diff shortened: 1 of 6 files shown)


borkdude:
I decided to make a few tweaks and merged it to master now, thanks again

PR #2471, opened 2025-01-24
Fix #2277: prefer array class symbol in defprotocol

### LINKED ISSUE #2277: Class/forName considered as function name (proposal: say that it must be a class symbol)
- [X] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**version** 
2024.02.12

**platform**
linux - native binary

**problem**
I have the following in a file (for extending XTDB's handler):

```clojure
(extend-protocol pedestal.servlet-interceptor/WriteableBody
           (Class/forName "[B") ;; This errors out
           (write-body-to-stream [^InputStream input-stream ^OutputStream output-stream]
             (with-open [out output-stream] (.write out ^bytes input-stream)))
           juxt.clojars_mirrors.muuntaja.v0v6v8.muuntaja.protocols.StreamableResponse
           (write-body-to-stream [^InputStream input-stream ^OutputStream output-stream]
             (with-open [out output-stream] ((.f input-stream) ^OutputStream out))))
```

The highlighter line is reported as:

```shell
src/web2/server.clj:53:27: error: Function name must be simple symbol but got: "[B"
```
I don't think it's a configuration error and it only happens with the latest version of `clj-kondo`.

**expected behavior**

It should not error out as above.

borkdude:


nbogart-relay:
This assumes Clojure 1.12. Does clj-kondo do any checking for which version the code will use?

NoahTheDuke:
This assumes Clojure 1.12. Does clj-kondo do any checking of the code's version?

borkdude:
It assumes clojure 1.12, but clj-kondo doesn't know the clojure version of the code it analyzes. I'm willing to take this trade-off since `Class/forName` wasn't officially supported and worked only by accident.

PR #2495, opened 2025-03-05
Report toolsreader exceptions as new :reader-error linter

### LINKED ISSUE #2492: incorrect line number reported for unsupported escape seq errors

v2025.02.20

**platform**

MS-Windows native

**problem**

The linter incorrectly reports the correct row/column number for unsupported escape seq characters (.e.g. `\a`)

**repro**

1. Create a `bad.clj` file with an unsupported escape seq on a line other than 1:
```clj
123
456
"\a"
```
2. Run the linter, the error is incorrectly reported at 1:4, instead of 3:1
```sh
$ clj-kondo --lint bad.clj 
bad.clj:1:4: error: Unsupported escape character: \a.
linting took 6ms, errors: 1, warnings: 0
```
**expected behavior**

The error should be reported at the correct line (line 3).

[ Optional: if you are interested in doing a PR yourself, please leave a note. ]

I'm happy to do the PR, but I need guidance on the best approach.

- The error occurs in `clj-kondo.impl.utils/sexpr` when `expr->tag` calls `(sexpr expr)`, where `expr` is the `rewrite-clj` `token` `"\a"`.
- The exception originates from `clojure.tools.reader`, propagates through `rewrite-clj`, and reaches back to `clj-kondo`.
- `rewrite-clj` appears to lack the correct file line metadata, but `clj-kondo` retains it in the node's metadata suggetsting the fix belongs in `clj-kondo`.

As a solution I can think of, catching `:reader-error` in `sexpr` and reporting it as a new `reader-error` linter is possible but may not be ideal since `sexpr` is widely used in the codebase and would require a `ctx` that may not always be available from the caller.

Below is a FlowStorm stack trace showing where the exception with the wrong line coordinate is thrown, in case it helps.

Image

ikappaki:
Hi,

can you please review patch to correctly report reader errors encountered during the analysis phase. It fixes #2492.

The tools reader may throw `:reader-errors` during the analysis phase with misleading text coordinates because it does not have access to the full text. The fix tries to address this by catching the exception during the analysis phase and register it with the correct coordinates under a new `reader-error` linter.

Test added to test the same. Please feel free to suggest any other alternative.

Thanks  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 ## Unreleased
 
 - [#2493](https://github.com/clj-kondo/clj-kondo/issues/2493): reduce image size of native image
+- [#2492](https://github.com/clj-kondo/clj-kondo/issues/2492): New linter: `:reader-error`, reports reader errors encountered during analysis.
 
 ## 2025.02.20
 
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -53,6 +53,7 @@ configuration. For general configurations options, go [here](config.md).
     - [Inline def](#inline-def)
     - [Invalid arity](#invalid-arity)
     - [Conflicting arity](#conflicting-arity)
+    - [Reader error](#reader-error)
     - [Reduce without initial value](#reduce-without-initial-value)
     - [Loop without recur](#loop-without-recur)
     - [Line length](#line-length)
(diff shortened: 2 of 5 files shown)


borkdude:
@ikappaki Here is my suggestion in practice:

https://github.com/clj-kondo/clj-kondo/pull/2501

However I don't understand why your test only expects two cases instead of three and why the location metadata is slightly different in my case.

borkdude:
Fixed via other PR

ikappaki:
> @ikappaki Here is my suggestion in practice:
> 
> #2501

Thanks @borkdude, this is a much better and more focused solution. I initially tried the same approach using `reg-finding!` instead of `*reader-exceptions*` but ran into cyclic dependency issues.

> However I don't understand why your test only expects two cases instead of three and why the location metadata is slightly different in my case.

This happened because the error here was reported on the map form rather than the individual key-value strings, which wasn’t ideal.

PR #2515, opened 2025-04-06
Fix #2511: fix multiple parse errors caused by incomplete forms

### LINKED ISSUE #2511: More parsing errors caused by incomplete forms
already a sponsor, thank you! ]

**version**

clj-kondo v2025.02.21-SNAPSHOT

**macro usage**

NA

**platform**

native

**editor**

NA

**problem**

I ran into more issues similar to #2496 and #2500, where an incomplete call would cause linting of the entire file to fail with no location info.
For completeness I went through every public var in `clojure.core` and found the following failures:

**repro**
```
(ns repro)

(cond->)
(cond->>)
(defmethod)
(doto)
(ns)
```

```
$ echo '(cond->)' | clj-kondo --lint -
<stdin>:0:0: error: Can't parse <stdin>, java.lang.NullPointerException

$ echo '(cond->>)' | clj-kondo --lint -
<stdin>:0:0: error: Can't parse <stdin>, java.lang.NullPointerException

$ echo '(defmethod)' | clj-kondo --lint -
<stdin>:0:0: error: Can't parse <stdin>, No implementation of method: :sexpr of protocol: #'clj-kondo.impl.rewrite-clj.node.protocols/Node found for class: nil

$ echo '(doto)' | clj-kondo --lint -
<stdin>:0:0: error: Can't parse <stdin>, java.lang.NullPointerException

$ echo '(ns)' | clj-kondo --lint -
<stdin>:0:0: error: Can't parse <stdin>, No implementation of method: :sexpr of protocol: #'clj-kondo.impl.rewrite-clj.node.protocols/Node found for class: nil
```

**config**

NA

**expected behavior**

The `:invalid-arity` linter should be triggered, and not interfere with linting of the rest of the file.

**additional info**
Here's the code I used to generate the above, perhaps something along similar lines could be incorporated into regression tests:
```clj
(require '[babashka.process :refer [process]])

(for [f (sort (keys (ns-publics 'clojure.core)))
      :let [lines (->> (process {:in (format "(%s)" f)}
                                "clj-kondo --lint -")
                       :out slurp
                       (str/split-lines)
                       (filter (partial re-find #"Can't parse")))]
      :when (seq lines)]
  (into [f] lines))
```

borkdude:


NoahTheDuke:
oh this is a great change

PR #2561, opened 2025-06-28
Fix #2560: NEW :locking-suspicious-lock linter

### LINKED ISSUE #2560: Catch locking on non-symbol
A common mistake is to use `locking` on some Clojure expression, rather than on a unique object. This happened to me in `sci.impl.vars.cljc` recently and now I discovered this in `sci.lang.cljc` too:

https://github.com/babashka/sci/commit/5064780bc4c1b592c092355d63e2505b393b3eda

borkdude:


NoahTheDuke:
all three states is great to check, nice

PR #2567, opened 2025-07-12
Expand :condition-always-true checks

### LINKED ISSUE #2566: Expand :condition-always-true

`:condition-always-true` only checks for function vars and vars. It would be nice if it also checked for all truthy objects, such as maps, sets, keywords, strings, etc.

I thought of this from running into `(if (filter ...) 1 2)`, where they meant to write `(if (seq (filter ...)) 1 2)` (checking if the filter actually finds something).

Expand `:condition-always-true` to other truthy objects.

Add a different linter? Seems unnecessary.

**Additional context**
I'm willing to write this one (already have a small experimental branch).

NoahTheDuke:

Closes #2566

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -941,6 +941,35 @@
        ctx
        (node->line (:filename ctx) expr :syntax (format "%s binding vector requires exactly 2 forms" form-name))))))
 
+(defn condition-always-true-linter
+  [ctx expr]
+  (findings/reg-finding! ctx (assoc (meta expr)
+                                    :filename (:filename ctx)
+                                    :message "Condition always true"
+                                    :type :condition-always-true)))
+
+(defn analyze-condition
+  [ctx condition]
+  (let [ctx (update ctx :arg-types #(or % (atom [])))
+        condition (assoc condition :condition true)
+        analyzed (doall (analyze-expression** ctx condition))
+        t (tag condition)
+        loc (select-keys (if (identical? :quote t)
+                           (meta (first (:children condition)))
+                           (meta condition))
+                         [:row :col :end-row :end-col])]
+    (when (not (linter-disabled? ctx :condition-always-true))
+      (when-let [arg-type (some->> @(:arg-types ctx)
+                                   (filter #(= loc (select-keys % [:row :col :end-row :end-col])))
+                                   (first)
+                                   :tag
+                                   types/keyword)]
+        (when (not (or (types/nilable? arg-type)
+                       (types/match? arg-type :nil)
+                       (types/match? arg-type :boolean)))
+          (condition-always-true-linter ctx condition))))
+    analyzed))
+
 (defn analyze-conditional-let [ctx call expr]
   (let [children (next (:children expr))
         bv (first children)
@@ -955,8 +984,7 @@
                                                         :analyzed))]
         (lint-two-forms-binding-vector! ctx call bv)
         (concat (:analyzed bindings)
-                (analyze-expression** (update ctx :callstack conj [:vector])
-                                      (assoc condition :condition true))
+                (analyze-condition (update ctx :callstack conj [:vector]) condition)
                 (if if?
                   ;; in the case of if, the binding is only valid in the first expression
                   (concat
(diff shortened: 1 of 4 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -941,6 +941,35 @@
...
+(defn analyze-condition
+  [ctx condition]
+  (let [ctx (update ctx :arg-types #(or % (atom [])))
+        condition (assoc condition :condition true)
+        analyzed (doall (analyze-expression** ctx condition))
+        t (tag condition)
+        loc (select-keys (if (identical? :quote t)
+                           (meta (first (:children condition)))
+                           (meta condition))
+                         [:row :col :end-row :end-col])]
```
NoahTheDuke, line 960: This is a hack but i don't know how to get around it. We need some way of finding out which `:arg-types` entry is for our condition, and the the quoted list's metadata is attached to the child node.

borkdude:
This is indeed a problem:

``` clojure
(cond-> {}
  :always (assoc :cool :beans)
  (map? {}) (assoc :other :stuff)
  :always (assoc :another :key))
```

I know some codebases that do this...

borkdude:
Maybe we should just treat keywords as a generalization of booleans? Like `(def debug :very)` instead of `(def debug true)`? And therefore not warn on them?

NoahTheDuke:
I could see a config for `:allow-keyword`, defaulting to true, and then that would bypass any shenanigans with looking at the callstack. I'll write it up and see how it feels.

NoahTheDuke:
Okay, I added that config and think it works fine. I looked at a lot of code examples and never saw any `(if :foo 1 2)` constructions, but countless `(cond-> :always ...)`, so this fix that for the vast majority of the time.

borkdude:
I think this PR can be improved so it will also detect stuff like:

```
(defn dude [] :dude)

(if-let [x (dude)] x 2)
```

in the sense that `dude` always returns a keyword. But this will require moving the linting to a later phase where we have more information. Will take me some time to figure that out.

borkdude:
Ah yes, I think we can do that in `lint-arg-types!`

src/clj_kondo/impl/analyzer.clj
borkdude: Can you explain the special casing of the quoted value? Don't quite understand it

NoahTheDuke:
I experimented with that, actually! It required much more internal digging and led to some messy code I wasn't a fan of, so I left it out. One of the issues I ran into was `fn` being typed as `:fn`, and either overwriting that (leading to strange errors) or figuring it out how to encode fn types which is above my abilities lol

NoahTheDuke:
Do you want me to try that out in this PR, then?

borkdude:
@NoahTheDuke I think I'm going to take a look at this myself (already have done so, but not finished looking ...)

borkdude:
@NoahTheDuke Can you reply to my other inline comment?

src/clj_kondo/impl/analyzer.clj
NoahTheDuke: `:arg-types` is a vector that tracks the location and type of analyzed forms. Because it's a vector, we have to go from our condition form to the arg-type form in some way. We can't just go by the shape cuz there could be identical forms, so we gotta go by position. However, the location metadata on a quoted form (pre-analyzed) will be different than the location metadata on that form's arg-type because it uses the actual form's location:

```
'foo
^-- (meta condition)
 ^-- :arg-type
```

To solve this, I "unwrap" the quote so that the `loc` I'm looking for is the `foo` object itself.

borkdude: I think this may be actually a bug. It doesn't make sense to add a new arg-types atom when linting a quoted value. I'm going to fix this first.

borkdude:
@NoahTheDuke I simplified the code a little bit (feel free to review post-merge) and merged via another PR. Thanks so much! I decided to hold off on the keyword configuration for now since it's very context sensitive whether you may want to use keywords or not, so a coarse configuration didn't make sense to me. In some future work we may want to look at catching more cases by looking at types from user-defined functions, but also in another issue/PR perhaps.

NoahTheDuke:
Glad I could set it up for you!

PR #2608, opened 2025-08-16
Make inline-configs lock file vcs friendly

### LINKED ISSUE #2606: Make it easy for users to know how to treat `inline-config` file and its lock file

**version**
```
clj-kondo v2025.07.28
```

**platform**
native on linux

**problem**
When using clj-kondo inline config, after a lint run, I see some unexpected files after the run is complete.

**repro**
From a new empty dir:
```shell
mkdir .clj-kondo
mkdir -p src/foo
cat >> src/foo/bar.clj <<EOF
(ns foo.bar)

(defn- #_:clj-kondo/ignore booya [])
EOF
```

Run lint:
```shell
clj-kondo --lint src
```
Outputs then exits with 0:
```
linting took 11ms, errors: 0, warnings: 0
```

List files in `.clj-kondo`
```shell
ls -la .clj-kondo
```

**actual behaviour**
I see:
```
drwxrwxr-x 4 lee lee 4096 Aug 14 10:01 .
drwxrwxr-x 4 lee lee 4096 Aug 14 10:12 ..
drwxrwxr-x 3 lee lee 4096 Aug 14 10:01 .cache
drwxrwxr-x 3 lee lee 4096 Aug 14 10:01 inline-configs
-rw-rw-r-- 1 lee lee    0 Aug 14 10:01 .lock
```

**expected behavior**
I expected to only see the `.cache` directory.

Is the generated `inline-config` normal for this scenario?

The `.lock` file seems like it should not be there.
 
**next up**
Happy to help if this is an issue.

lread:
Clj-kondo users should not be committing the `inline-configs` lock file.

Move it from `./.clj-kondo/.lock` to `./.clj-kondo/.cache/.config-lock` where `./.clj-kondo` will be affected by `--config-dir` but `.cache` will not be affected by `--cache` nor `--cache-dir`.

This keeps the lock file in a constant spot relative to `inline-configs` dir, but now also a directory that clj-kondo users have been advised to git ignore.

Also: add advice on version control to docs.

Closes #2606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - Sort findings on filename, row, column and now additionally on message too
 - [#2602](https://github.com/clj-kondo/clj-kondo/issues/2602): Sort auto-imported configs to avoid differences based on OS or file system
 - [#2603](https://github.com/clj-kondo/clj-kondo/issues/2603): warn on `:inline-def` with nested `deftest`
+- [#2606](https://github.com/clj-kondo/clj-kondo/issues/2606): make it easy for users to know how inline-config files should be version controlled ([@lread](https://github.com/lread))
 
 ## 2025.07.28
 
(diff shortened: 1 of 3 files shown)


borkdude:
Thanks!

PR #2635, opened 2025-10-11
Add :interface? field to :java-class-definitions analysis output

hugoduncan:
Problem Statement

When analyzing Java class definitions, clj-kondo's :java-class-definitions
analysis output doesn't distinguish between regular classes and interfaces. This
makes it difficult for tools consuming the analysis data (like clojure-lsp) to
determine whether a Java type is a class or an interface, which is important
for:
- Providing accurate code intelligence and navigation
- Generating correct code completions
- Understanding the type hierarchy and contracts

Solution

Add an :interface? boolean field to the :java-class-definitions analysis output
that is set to true when the Java type is an interface. The field is only
present when true to maintain backward compatibility for regular classes.

Additionally, the :flags set now includes :interface for interface types, and
interface methods are correctly marked as :public (matching Java's implicit
public modifier for interface methods).

Implementation Details

The implementation handles both bytecode (.class) and source (.java) files:
- For .class files: Uses ASM's ACC_INTERFACE opcode flag
- For .java files: Uses JavaParser's ClassOrInterfaceDeclaration.isInterface()
method
- Interface methods from source files are automatically marked as :public to
match bytecode behavior

Changes

- Added :interface? field to :java-class-definitions for interface types
- Added :interface to the :flags set for interface types
- Ensured interface methods are marked as :public in source file analysis
- Added comprehensive tests covering both .class and .java files
- Updated analysis documentation

Test Coverage

The test suite includes:
- Interface detection from compiled .class files
- Interface detection from .java source files
- Verification that regular classes don't have :interface? field
- Verification that :interface flag is present in :flags set
- Verification that all interface methods are marked as :public

All tests pass with both clojure -M:test and lein test.

Example Output

Before:
{:class "foo.bar.SampleInterface"
:uri "file:///path/to/SampleInterface.java"
:filename "/path/to/SampleInterface.java"
:flags #{:public}}

After:
{:class "foo.bar.SampleInterface"
:uri "file:///path/to/SampleInterface.java"
:filename "/path/to/SampleInterface.java"
:flags #{:public :interface}
:interface? true}

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 ## Unreleased
 
 - [#2633](https://github.com/clj-kondo/clj-kondo/issues/2633): support new `defparkingop` macro in core.async alpha
-- [#2635](https://github.com/clj-kondo/clj-kondo/pull/2635): Add `:interface?` field to `:java-class-definitions` analysis output to distinguish Java interfaces from classes
+- [#2635](https://github.com/clj-kondo/clj-kondo/pull/2635): Add `:interface` flag to `:flags` set in `:java-class-definitions` analysis output to distinguish Java interfaces from classes
 
 ## 2025.09.22
 
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -202,7 +202,6 @@ The analysis output consists of a map with:
   - `:class`: full qualified class name.
   - `:uri`: file URI of this class.
   - `filename`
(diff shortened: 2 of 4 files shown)


borkdude:
@hugoduncan I think this is sufficient:

```
Additionally, the :flags set now includes :interface for interface types, and
interface methods are correctly marked as :public (matching Java's implicit
public modifier for interface methods).
```

So just `:flags` without the special `:interface?` field would be my preference.

borkdude:
Thanks!

PR #2650, opened 2025-10-26
Duplicate key args for associative fns

### LINKED ISSUE #2590: Lint assoc-ing the same key twice
`(assoc {} :foo 1 :bar 2 :foo 2)`

tomdl89:
Further fixes #2590

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -783,7 +783,7 @@ Explanation by Bozhidar Batsov:
 
 *Example trigger:* `(assoc {} :a 1 :a 2)`
 
-*Example message:* `Duplicate key args for assoc: :a`.
+*Example message:* `Duplicate key arg supplied to assoc: :a`.
 
 ### Dynamic vars
 
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -2119,7 +2119,7 @@
       (findings/reg-finding! ctx (assoc (meta k)
                                         :filename (:filename ctx)
                                         :type :duplicate-key-args
-                                        :message (str "Duplicate key args for " fn-name ": " k))))
+                                        :message (str "Duplicate key arg supplied to " fn-name ": " k))))
     (if (seq ks)
       (recur ks (update k-count k (fnil inc 0)))
       (analyze-children ctx children false))))
--- a/test-regression/clj_kondo/metabase/findings.edn
+++ b/test-regression/clj_kondo/metabase/findings.edn
@@ -923,7 +923,7 @@
   :col 62,
   :end-col 67,
   :langs (),
-  :message "Duplicate key args for assoc: :type",
+  :message "Duplicate key arg supplied to assoc: :type",
   :row 2857}
  {:end-row 821,
   :type :unresolved-symbol,
--- a/test/clj_kondo/duplicate_key_args_test.clj
+++ b/test/clj_kondo/duplicate_key_args_test.clj
@@ -9,7 +9,7 @@
      :row 1,
      :col 16,
      :level :warning,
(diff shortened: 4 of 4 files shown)


tomdl89:
Assuming you want these as separate linters? I couldn't think of a terse name that covers both, so for that reason I made a new one. Also went for the "Do Repeat Yourself" version of DRY 😏 for the logic, but I could pull out the common logic. Not sure it's really worthwhile, but I don't mind. Hopefully otherwise all straightforward.

borkdude:
Perhaps we can rename the linter to `duplicate-key-in-associative` as an umbrella for `dissoc`, `assoc` and `assoc!`

What comes to mind as well is `conj` but there duplicate elements aren't necessarily wrong and it's not associative. So limiting to associative makes sense?

tomdl89:
Do you mean like the following?
```clojure
(conj {} [:a 1] [:a 2])
;; => {:a 2}
```
I didn't address that in the latest commit, but did combine the linters for assoc and dissoc, and did abstract some common code. lmk if you want me to take a look at conj too. I think it's less common for people to use conj for maps, but it does happen.

borkdude:
LGTM but maybe one more thought about the name: duplicate-key-in-associative could be misunderstood as: ah, a duplicate key in a map literal (which kondo already lints) so perhaps `duplicate-key-in-associative-op` or something? But then again, we might lint `(conj #{} 1 1)` too and sets aren't associative... So let's think a bit more about the name.

tomdl89:
I don't know if you like "positive" warnings, but if so`unique-keys-expected` could fit the bill?

borkdude:
we already have a number of `:duplicate-*` linters so it would be nice the name could align with that. Perhaps just:

```
:duplicate-args
```

?

And this only warns for arguments that are expected to be different?

tomdl89:
Yeah I like that. Nice a simple too. One other thing that stuck me: the current implementation (using `frequencies`) means the line/column numbers correspond to the first such arg found. Might it be better to highlight the second?

borkdude:
Yeah, second is better

tomdl89:
OK, now linting on first duplicate rather than zeroth. Also, I went for
```
:duplicate-key-args
```
as I thought it was a bit clearer. But both commits are there, so very easy to discard if you disagree. I merged rather than rebased, for clarity. lmk if all looks ok.

tomdl89:
I realised lots more core fns should have non-duplicated key args, so added those too. Details in the commit message.

borkdude:
LGTM. Do you still think the duplicate-key name is fitting?

tomdl89:
I would say for these, the thing being checked for duplicates is definitely a key:
```
assoc
assoc!
dissoc
dissoc!
sorted-map-by
array-map
hash-map
sorted-map
```
For the set functions, it's less clear, but I think permissible. For each, I've noted the relevant param name in clojure.core:
```
hash-set       & keys
sorted-set     & keys
sorted-set-by  & keys
disj           key & ks
disj!          key & ks

```
and I'm no expert on these (they are effectively deprecated) but I'm pretty sure they are keys:
```
create-struct  & keys
defstruct      & keys
struct-map     & inits
```

Having said that, it's not a hill I would remotely die on 🙂

borkdude:
Yeah I think it makes sense. I'll sleep on it and merge tomorrow

borkdude:
One final nitpick. Could we change the messages to:

```
Duplicate key supplied to sorted-set-by: :a 
```

?

borkdude:
Or do you think it'd be harder to find the related linter when only looking at the message? In that case we could do:

```
Duplicate key arg supplied to sorted-set-by: :a 
```

borkdude:
Thanks!

PR #2655, opened 2025-11-08
Add redundant-let-binding linter

### LINKED ISSUE #2654: Lint redundant let bindings
e.g.
```clojure
(let [x x] x)
```
Just for simple symbols, and excluding type-hinted symbols, which may be used for perf.
Usually just left in mistakenly after refactoring. There are a couple of examples in clj-kondo itself, which is how I thought of it. A few crop up in the regression test too.

I'll make a PR. lmk what you think :)

tomdl89:
Fixes #2654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [clojure-lsp#2651](https://github.com/clojure-lsp/clojure-lsp/issues/2157): Include inner class java-class-definition analysis.
 - Bump `babashka/fs`
 - [#2532](https://github.com/clj-kondo/clj-kondo/issues/2532): Disable `:duplicate-require` in `require` + `:reload` / `:reload-all`
+- [#2432](https://github.com/clj-kondo/clj-kondo/issues/2432): Don't warn for `:redundant-fn-wrapper` in case of inlined function
 
 ## 2025.10.23
 
--- a/deps.edn
+++ b/deps.edn
@@ -4,7 +4,7 @@
         cheshire/cheshire {:mvn/version "5.13.0"}
         nrepl/bencode {:mvn/version "1.2.0"}
         org.babashka/sci {:mvn/version "0.10.47"}
-        babashka/fs {:mvn/version "0.5.28"}
+        babashka/fs {:mvn/version "0.5.30"}
         org.ow2.asm/asm {:mvn/version "9.7"}
         com.github.javaparser/javaparser-core {:mvn/version "3.26.1"}}
  :aliases {:cider {:extra-deps {org.clojure/clojure {:mvn/version "1.11.3"}}}
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1372,6 +1372,9 @@
   (or (contains? fixed-arities arg-count)
       (and varargs-min-arity (>= arg-count varargs-min-arity))))
 
+(def inlined-vars
+  '#{+' unchecked-remainder-int unchecked-subtract-int dec' short-array bit-shift-right aget = boolean bit-shift-left aclone dec < char unchecked-long unchecked-negate unchecked-inc-int floats pos? boolean-array alength bit-xor unsigned-bit-shift-right neg? unchecked-float num reduced? booleans int-array inc' <= -' * min get long double bit-and-not unchecked-add-int short quot unchecked-double longs unchecked-multiply-int int > unchecked-int unchecked-multiply unchecked-dec double-array float - byte-array zero? unchecked-dec-int rem nth nil? bit-and *' unchecked-add identical? unchecked-divide-int unchecked-subtract / bit-or >= long-array object-array doubles unchecked-byte unchecked-short float-array inc + chars ints bit-not byte max == count char-array compare shorts unchecked-negate-int unchecked-inc unchecked-char bytes})
+
 (defn redundant-fn-wrapper [ctx callstack children interop?]
   (when-let [fn-args (:fn-args ctx)]
     (when (and
@@ -1382,11 +1385,16 @@
            (= 1 (:fn-body-count ctx))
            (= 1 (:body-children-count ctx))
            (= (count children) (count fn-args))
-           (one-of (first callstack) [[clojure.core fn]
-                                      [clojure.core fn*]
-                                      [cljs.core fn]
-                                      [cljs.core fn*]])
-           (not= '[cljs.core .] (second callstack))
+           (one-of (second callstack) [[clojure.core fn]
+                                       [clojure.core fn*]
+                                       [cljs.core fn]
+                                       [cljs.core fn*]])
+           (let [[core-ns f] (first callstack)]
+             (or (not f)
+                 (and (or (= 'clojure.core core-ns)
+                          (= 'cljs.core core-ns))
+                      (not (contains? inlined-vars f)))))
+           (not= '[cljs.core .] (nth callstack 2 nil))
            (= (map #(str/replace % #"^%$" "%1") children)
               (map str fn-args)))
       (:fn-parent-loc ctx))))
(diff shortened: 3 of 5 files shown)


borkdude:
I wonder what happens if we would apply this to all "let-likes"? Any false positives to be expected?

tomdl89:
Yeah, it's quite common to see this pattern, which is totally legit:
```clojure
(loop [foo foo] ...)
```
`for` and `doseq` are also places where it's not redundant, just odd (and obviously shadowing):
```clojure
(for [foo foo] (inc foo))
```
But one place that did cross my mind is `:let` bindings. I'll have a dig there...

tomdl89:
OK that's added.

I also noticed, looking through the regression test output, that reader-conditionals can cause false positives: https://github.com/metabase/metabase/blob/master/src/metabase/util/formatting/internal/date_builder.cljc#L56 that is:
```clojure
(let [foo #?(:cljs foo :clj bar)] ...)
```
so will take a look at fixing that. Come to think of it, maybe it's still an ok warning? It could/should be changed to this:
```clojure
(let [#?@(:clj [foo bar], :default [])] ...)
```
thoughts?

borkdude:
There's also this case I discovered in some SCI code of mine:

```
(let [foo 1 bar foo])
```

Strictly speaking bar is unnecessary, you could just re-use foo. But maybe for clarity it's sometimes better to give things a different name?

borkdude:
I'd say with reader conditionals lean towards not warn. The issue of a redundant let binding isn't critical in any way (performance or correctness) so like redundant let en do it's probably better not to warn here?

tomdl89:
> But maybe for clarity it's sometimes better to give things a different name?

That's an interesting one. Probably safer not to warn. I think it's less likely someone would do this by mistake. I'll give some thought. 

> I'd say with reader conditionals lean towards not warn

Any tips on how to "see" the reader conditional at the analyze stage? It seems to be gone by then, cos all branches are analyzed separately..?

borkdude:
I do this usually in a post-processing step. If there aren't 2 warnings (for both clj and cljs) I don't warn at all. this is in a core or core.impl namespace

tomdl89:
Awesome, thanks. I'll take a look

borkdude:
@tomdl89 Dev tip, you can update the regression expected warnings with:

```
CLJ_KONDO_REGRESSION_UPDATE=true bb test:regression
```

tomdl89:
I added this linter to those in filter-findings, which did the trick. Also added conditional-let bindings (like `if-let` and `when-some`).

I considered adding the linting to `binding` and `with-redefs` too, but it did occur to me that `binding` is used as much for its restoring power as its binding power, so
```clojure
(binding [*ns* *ns*] ...)
```
could be ok. I found that here: https://github.com/nextjournal/clerk/blob/main/src/nextjournal/clerk/analyzer.clj#L332 which I think is legit. Any others you think we're missing? Otherwise, I could update the regression expected output. Bear in mind that some configs containing
```clojure
:lint-as {foo clojure.core/let}
```
(such as in metabase) will now warn. That could be handled by linting as something else instead (such as `clojure.core/binding`), but worth knowing about.

borkdude:
> :lint-as {foo clojure.core/let}
(such as in metabase) will now warn. That could be handled by linting as something else instead (such as clojure.core/binding), but worth knowing about.

That's a no-go. I think we should err on the side of not warning here, only warn when it's really clojure.core/let.

borkdude:
Well... maybe it's ok. I think we need to inspect it on a case by case basis. If it's really a let-like thing then the warning is justified I guess. You're right: linting as binding is more appropriate in such situations. Let's see.

tomdl89:
Yeah, the only things which look like false positives in metabase are these two:
```clojure
metabase.query-processor.setup/with-qp-setup
metabase.util.files/with-open-path-to-resource
```
which I think should be linted as `clojure.core/with-open` really.
Shall I push the updated findings.edn? Otherwise I could make a PR to metabase and once merged we could update the hash and then merge this?

borkdude:
Since `:redundant-let-binding` is a pretty risky thing to enable by default since it's new and may uncover more things AND since it doesn't really detect super critical stuff, let's just maybe set the default level to off so we can test it more in practice.  Most later non-critical linters are set to off by default. 

Having said this, we can enable it in the regression tests and update the findings, so we can still learn from detecting these things.

Also one thing: to save CPU cycles, let's check if the linter is enabled before checking for redundant let bindings?

tomdl89:
I also noticed sometimes you set linters to `:info` so they show up in output but don't affect the exit code? Could that be suitable here?

borkdude:
We have three :info linters and they warn about redundant calls which may be beneficial for performance (+ redundant ignore which is helpful for re-discovering previously ignored issues). redundant-let-binding is only beneficial for cosmetics

tomdl89:
> we can enable it in the regression tests and update the findings

the metabase and nextjournal tests both reference the `/.clj-kondo` config-dir, so I can't reference `:config` in the calls to `clj-kondo/run!` (it seems to ignore it). Any tips?

borkdude:
you can enable extra linters here:

https://github.com/clj-kondo/clj-kondo/blob/dd45054e3eb8d9aa76a7da0e42826144dbd957c6/test-regression/clj_kondo/metabase/metabase_test.clj#L34

tomdl89:
That's what I tried, and it didn't seem to work. I'm afk right now but will try again later. Maybe I missed something

borkdude:
OK, flying out to the conj this week, so limited time to look at new replies until next week

tomdl89:
Alright, I think it's good for final review / merge if all looks good to you.

tomdl89:
Is this ok to merge @borkdude ?

borkdude:
It's on my list to re-review, been getting a lot of PRs and issues I have to go through since I came back from the conj ;-)

tomdl89:
Ah of course, no worries :)

borkdude:
@tomdl89 The message contains some redundant info in my opinion. 

My suggestion, change the message from:

"Redundant binding of x to x" 

to

"Redundant let binding: x" (which is similar to other message where the non-unique part follows after a colon)

tomdl89:
I think maybe just:
"Redundant binding: x"
as it may not be `let`. E.g. it's called on `when-some` and `when-first`
wdyt?

borkdude:
Well, the linter name is called `redundant-let-binding`

I don't think the binding here is redundant for several reasons:

```
(when-let [inc inc] (inc 1))
```

First reasons is that dereferencing a var, like `inc` causes invocations to not go through the var anymore and prevents them from inlining.
Secondly, this binding isn't redundant. Removing it changes the semantics of the program (and in this case, makes it crash).
Redundant imo means: can be removed without changing anything.

tomdl89:
Sorry, just to be clear, are you saying that
```clojure
(when-let [inc inc] (inc 1))
```
can't be replaced with
```clojure
(when inc (inc 1))
```
because there are performance concerns?
surely if that's the case, that would mean the following shouldn't be a lint warning for the same reason?
```clojure
(when inc (let [inc inc] (inc 1)))
```
which would mena the whole linter is invalid for the same reason? (n.b. that's basically what `when-let` macroexpands to, minus an extra let block to avoid evaluating the test value twice).

> Redundant imo means: can be removed without changing anything.

Yeah, that's a fair point. For that reason, I can see why you might want to not call it on `when-let` etc. Maybe it's not obvious to everyone that `(when-let [foo foo]...` can be replaced with `(when foo ...` and `(when-some [foo foo]...` can be replaced with `(when (some? foo)...`.

I think in the interests of getting this merged, it'll be best for me to just rip this stuff out. It can always be added later in a later PR if we resolve the above.

borkdude:
My point is more that:

```
(when-let [inc inc] (inc 1))
```

can't be simply fixed by removing the redundant binding:

```
(when-let [] (inc 1))
```

so let's not warn about that right now

tomdl89:
Sorry for the back-and-forth and thanks for bearing with me 😄

borkdude:
No problem, I'm also learning more as we go back and forth ;)

PR #2669, opened 2025-12-11
Lint trailing varargs syntax errs, lint varargs errs in let bindings

### LINKED ISSUE #2668: Improve varargs syntax linting
I noticed kondo only lints varargs syntax errors in function args, not in let bindings.

E.g. this:

```clojure
(defn foo [x y & xs ys] [x y xs ys])
```

but not this:
```clojure
(let [[x y & xs ys] foo] [x y xs ys])
```

and it also misses nested varags, like this:
```clojure
(defn foo [a & [b & c d]] [a b c d])
```

and then I noticed it misses trailing `&`, which is a syntax error. e.g.
```clojure
(defn foo [a b &] (* a b))
```

so I fixed them all and this is its issue. PR incoming. Hopefully that's helpful :)

tomdl89:

fixes #2668

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -3170,7 +3170,17 @@
                                                             :message "Reader conditionals are only allowed in .cljc files")))
                         (analyze-reader-macro ctx expr))
         (:unquote :unquote-splicing)
-        (analyze-children ctx children)
+        (let [level (:syntax-quote-level ctx)]
+          (when-not (and level (pos? level))
+            (findings/reg-finding!
+             ctx
+             (node->line (:filename ctx) expr :unquote-not-syntax-quoted
+                         (if (= :unquote t)
+                           "Unquote (~) not syntax-quoted"
+                           "Unquote-splicing (~@) not syntax-quoted"))))
+          (let [new-level (if level (dec level) -1)
+                ctx (assoc ctx :syntax-quote-level new-level)]
+            (analyze-children ctx children)))
         :namespaced-map (do
                           (lint-unused-value ctx expr)
                           (usages/analyze-namespaced-map
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,9 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#2532](https://github.com/clj-kondo/clj-kondo/issues/2532): Disable `:duplicate-require` in `require` + `:reload` / `:reload-all`
 - [#2432](https://github.com/clj-kondo/clj-kondo/issues/2432): Don't warn for `:redundant-fn-wrapper` in case of inlined function
 - [#2599](https://github.com/clj-kondo/clj-kondo/issues/2599): detect invalid arity for invoking collection as higher order function
-
+- [#2661](https://github.com/clj-kondo/clj-kondo/issues/2661): Fix false positive `:unexpected-recur` when `recur` is used inside `clojure.core.match/match` ([@jramosg](https://github.com/jramosg))
+- [#2617](https://github.com/clj-kondo/clj-kondo/issues/2617): Add types for `repeatedly` ([@jramosg](https://github.com/jramosg))
+- [#2653](https://github.com/clj-kondo/clj-kondo/issues/2653): NEW linter: `:unquote-not-syntax-quoted` to warn on `~` and `~@` usage outside syntax-quote (`` ` ``) ([@jramosg](https://github.com/jramosg))
 ## 2025.10.23
 
 - [#2590](https://github.com/clj-kondo/clj-kondo/issues/2590): NEW linter: `duplicate-key-in-assoc`, defaults to `:warning`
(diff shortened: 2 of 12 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -264,7 +264,7 @@
                        exclude-as? (-> ctx :config :linters :unused-binding
                                        :exclude-destructured-as)
                        as-sym (when exclude-as?
-                                (let [[as as-sym] (drop (- (count children) 2) children)]
+                                (let [[as as-sym] (take-last 2 children)]
```
borkdude, line 305: Can you please revert this unnecessary change?

borkdude: Never mind, leave it in.

borkdude: No, I stand by my choice to drop this change. `drop` + `count` is much faster :).

```
user=> (let [r (range 10)] (time (dotimes [i 10000000] (take-last 2 r))))
"Elapsed time: 643.197 msecs"
nil
user=> (let [r (range 10)] (time (dotimes [i 10000000] (drop (- (count r) 2) r))))
"Elapsed time: 127.117583 msecs"
```

tomdl89: Ah sorry, I left that in by mistake. Yup, I'm aware of its shoddy performance. Reverting...

tomdl89: That's done. But if you're interested in the improvement in readability, while keeping perf. I could add this to utils.clj:
```clojure
(:refer-clojure :exclude [update-vals take-last])
...
(defn take-last [n coll]
  (if (counted? coll)
    (drop (- (count coll) n) coll)
    (clojure.core/take-last n coll)))
```
and use that? If not, feel free to ignore :)

borkdude: Even if the collection isn't counted, e.g. a realized lazy seq, it may be faster.

```
user=> (counted? (map inc [1 2 3]))
false
```

tomdl89: ~TIL how truly awful `take-last` is. I assumed it would be no worse for uncounted colls. But I just did some benchmarks and you're right. I think there are so few situations where you'd ever prefer it, that it shouldn't be in core as it is. Thanks for pointing it out!~

tomdl89: OK, I got into a rabbit-hole of benchmarking this evening.
TL;DR - most likely for this situation, `take-last` is quicker. Combined with its better readability, this refactor isn't  a bad one.

Your above examples are flawed, because you're comparing an eager function and a lazy function, and concluding the lazy function is faster, without realizing the outputs. When benchmarking, you need to use `dorun` or `doall` to compare them fairly.

I wrote up my findings on [clojuredocs](https://clojuredocs.org/clojure.core/take-last#example-693cae8fb7956e24e4cb4ec0), but the gist is that `take-last` can be faster when taking a small number of elements from lazy sequences. In pretty much all other situations, it's slower. But both approaches are fine for anything other than very hot paths.

```clojure
(defn my-take-last [n coll]
  (drop (- (count coll) n) coll))

(->> (c/benchmark-round-robin [(dorun (take-last    2 (map inc (range 100))))
                               (dorun (my-take-last 2 (map inc (range 100))))]
                              c/*default-benchmark-opts*)
     (map (comp first :mean)))
;; => (1.968E-6 2.613E-6)
;; take-last is ~25% faster

```
Realizing a lazy seq doesn't change that, although maybe narrows the gap:
```clojure
(->> (c/benchmark-round-robin [(dorun (take-last    2 (doall (map inc (range 100)))))
                               (dorun (my-take-last 2 (doall (map inc (range 100)))))]
                              c/*default-benchmark-opts*)
     (map (comp first :mean)))
;; => (2.691E-6 3.097E-6)
;; take-last is ~13% faster 
```
In this specific situation, `children` is a realised lazy sequence, so benefits from take-last.

I put a `time` in-situ around the surrounding when block in both scenarios, and `take-last` was 2.17% faster than the current solution.

If you've read all of this, sorry that you won't get these minutes of your life back 😂 . To me, it's interesting, and being right all along (although by luck, not judgement!) is a nice bonus.

borkdude: > OK, I got into a rabbit-hole of benchmarking this evening 

😂

> But both approaches are fine for anything other than very hot paths.

Yes, I expected as much ;-)

My stance is that in a PR that is not related to this change, this code shouldn't have changed as it just creates confusion (and thus more time to review) for the reviewer. 

But now that we spent the time on this, just leave it in.

tomdl89: > My stance is that in a PR that is not related to this change, this code shouldn't have changed as it just creates confusion (and thus more time to review) for the reviewer.

Very good point. I was too badly nerd sniped to see that 😬 

> But now that we spent the time on this, just leave it in.

Cheers!

tomdl89:
I realised this PR could include `:as` binding problems too, so pulled the binding vector analysis into its own function. I think this makes everything more readable too. lmk what you think :)

borkdude:
Thanks, I'll hopefully review tomorrow

PR #2670, opened 2025-12-12
Fix: redundant-call's level check didn't work

tomdl89:
Just a small thing - looks like an oversight when it was originally written.
Should have a small perf improvement.


borkdude:
Good catch!

PR #2679, opened 2025-12-17
Fix Detect unused :excluded vars (`:unused-excluded-var`)

### LINKED ISSUE #2600: Detect unused :excluded vars (`:unused-excluded-var`)
E.g.:

```
(ns sci.impl.read
  {:no-doc true}
  (:refer-clojure :exclude [read read-string read+string])
  (:require [clojure.tools.reader.reader-types :as r]
            [sci.ctx-store :as store]
            [sci.impl.io :as io]
            [sci.impl.parser :as parser]
            [sci.impl.utils :as utils]))
```

If `read` etc aren't overrriden, we can report that it's an unused-excluded-var

jramosg:
Fixes #2600 - This commit introduces checks for unused excluded variables in the Clojure codebase. It updates the findings for various files to include warnings for unused excluded vars. Additionally, a new test suite is added to ensure the linter correctly identifies unused excluded vars and handles cases where they are used or excluded. This enhances code quality by helping developers identify and remove unnecessary variables.

### DIFF
--- a/test/clj_kondo/main_test.clj
+++ b/test/clj_kondo/main_test.clj
@@ -221,7 +221,7 @@
      '({:file "corpus/exclude_clojure.clj"
         :row 6
         :col 29
-        :level :warning
+        :level :info
         :message "Unused excluded var: get"}
        {:file "corpus/exclude_clojure.clj",
         :row 12,
@@ -348,7 +348,7 @@ foo/foo ;; this does use the private var
       (assert-submaps
        '({:row 2,
           :col 29,
-          :level :warning
+          :level :info
           :message "Unused excluded var: cond"}
          {:row 9,
           :col 3,
@@ -584,7 +584,7 @@ foo/foo ;; this does use the private var
    '({:file "corpus/rename.cljc"
       :row 2
       :col 58
-      :level :warning
+      :level :info
       :message "Unused excluded var: conj"}
      {:file "corpus/rename.cljc",
       :row 4,


borkdude:
@jramosg Can you resolve merge conflicts?

jramosg:
> @jramosg Can you resolve merge conflicts?
> 
> 

I'm out without computer until Wednesday, I will do it as soon as I can, sorry!

borkdude:
No worries. I'll make a release of clj-kondo and we can integrate this and your other work in the next version. It's big enough as it is already :)

borkdude:
I found a false positive:

```
$ clj -M:clj-kondo/dev --lint - <<< "(ns foo (:refer-clojure :exclude [comp]) (:require [other-ns :refer [comp]])) comp"
<stdin>:1:35: warning: Unused excluded var: comp
<stdin>:1:70: warning: Unresolved var: comp
linting took 34ms, errors: 0, warnings: 2
```

jramosg:
@borkdude I merged and fixed the false positive. Thank you!

test-regression/clj_kondo/clj_kondo/findings.edn
```diff
@@ -3550,6 +3991,69 @@
...
+  :message "Unused excluded var: if-some",
+  :row 39}
+ {:end-row 39,
+  :type :unused-excluded-var,
+  :level :warning,
+  :filename "cljs/core.cljc",
+  :col 37,
+  :end-col 46,
+  :langs (),
+  :message "Unused excluded var: when-some",
```
borkdude, line 4055: when-some (and others) seem to be redefined here:

https://github.com/clojure/clojurescript/blob/8a596a8bdf8188112332207e5351e143740168f5/src/main/clojure/cljs/core.cljc#L537

so it's a false positive

jramosg: > when-some (and others) seem to be redefined here:
> 
> https://github.com/clojure/clojurescript/blob/8a596a8bdf8188112332207e5351e143740168f5/src/main/clojure/cljs/core.cljc#L537
> 
> so it's a false positive

updated by looking at `import-macros`

jramosg: looking at every `used-vars` was not enough because namespaced calls like the next `replace` didn't throw a warning. I think they should.

```clojure
(ns foo
  (:require [lib.util.match])
  (:refer-clojure :exclude [replace]))

(defn desugar-does-not-contain
  [m]
  (lib.util.match/replace m
                          [:does-not-contain & args]
                          [:not (into [:contains] args)]))
```

borkdude: > updated by looking at import-macros

Please remove this, it's too ad-hoc. If the false positive came from clj-kondo not understanding the import-macros macro, then it's fine. We don't change stuff just so that one particular namespace will work.

borkdude: > looking at every used-vars was not enough because namespaced calls like the next replace didn't throw a warning. I think they should.

Do you mean, that it should warn that replace was an unused excluded var? That's right. You shouldn't be looking at used-vars at all probably, since `:exclude` is only about defined vars.

jramosg: Yes, I think it should warn and it actually does

jramosg: > > updated by looking at import-macros
> 
> Please remove this, it's too ad-hoc. If the false positive came from clj-kondo not understanding the import-macros macro, then it's fine. We don't change stuff just so that one particular namespace will work.

@borkdude  But if I remove it, this test will fail and `when-some` is really used in `clj` by `import-macros` and in `cljs` by `defmacro`. The `defmacro` was detected good previous to the change but the `import-macros` not.

```clojure
(testing "excluded var used via conditional import and macro definition"
    (is (empty? (lint! "(ns foo
      (:refer-clojure :exclude [when-some])
    (:require clojure.walk
              clojure.set
              #?(:cljs [cljs.core :as core])))
    #?(:clj
       (import-macros clojure.core
                      [when-some]))
    
    #?(:cljs
       (core/defmacro when-some
         \"bindings => binding-form test
    
          When test is not nil, evaluates body with binding-form bound to the
          value of test\"
         [bindings & body]
         (assert-args when-some
                      (vector? bindings) \"a vector for its binding\"
                      (= 2 (count bindings)) \"exactly 2 forms in binding vector\")
         (core/let [form (bindings 0) tst (bindings 1)]
           `(let [temp# ~tst]
              (if (nil? temp#)
                nil
                (let [~form temp#]
                  ~@body))))))"
                       "--lang" "cljc"))))
```

borkdude: Just ignore this, this is highly specific to how CLJS macros are defined and not interesting to clj-kondo users in general.

jramosg: done!

src/clj_kondo/impl/linters.clj
```diff
@@ -704,6 +705,26 @@
...
+    (doseq [{:keys [clojure-excluded] :as ns} (namespace/list-namespaces ctx)
+            :when (and (seq clojure-excluded)
+                       (not (linter-disabled? ns :unused-excluded-var)))
+            :let [{:keys [lang referred-vars vars bindings]} ns
+                  core-original-name (fn [[_ {var-name :name}]]
+                                       (when (var-info/core-sym? lang var-name)
+                                         var-name))
+                  used (set (concat (keys vars)
+                                    (keep core-original-name referred-vars)
+                                    (map :name bindings)))]
```
borkdude, line 716: Why are bindings referred here? Binding names do not count as var usages?

jramosg: I'm not sure if this should  represent local shadowing rather than actual usage of the excluded core var.

In regression tests I found that a lot of vars were excluded because of bindings, that's the reason I refer them.
This is a simple test i did to test that. In this example, read is not count as var usage (at least in `vars` or  `referred-vars`).

```clojure
 (ns foo 
   (:refer-clojure :exclude [read]))

 (let [read 1] read)
```
 
I could just delete the `(map :name bindings)` and this example will throw the `:unused-excluded-var` warning

borkdude: That's a good point. In this case people could add an ignore hint:

```
 (ns foo 
   (:refer-clojure :exclude [#_:clj-kondo/ignore read]))
```

Were there any other places outside of clj-kondo where you found this?

jramosg: Yes, I found quite a lot in metabase

borkdude: interesting, can you let me know an example?

jramosg: There are 3 in metabase:

I suppose that it is to avoid `shadowed-var` warning. Maybe be can look at bindings only when `shadowed-var` is not disabled

```clojure
{:end-row 2,
  :type :unused-excluded-var,
  :level :warning,
  :filename
  "test-regression/checkouts/metabase/test/metabase/formatter/impl_test.clj",
  :col 29,
  :end-col 35,
  :langs (),
  :message "Unused excluded var: format",
  :row 2}
```

```clojure
(deftest ambiguous-column-types-test
  (testing "Ambiguous column types (eg. `:type/SnowflakeVariant` pass through the formatter without error (#46981)"
    (let [format (fn [value viz]
                   (str ((formatter/number-formatter {:id 1
                                                      :base_type :type/SnowflakeVariant}
                                                     {::mb.viz/column-settings
                                                      {{::mb.viz/field-id 1} viz}})
                         value)))]
      (is (= "variant works"
             (format "variant works" {}))))))
```
---

```clojure
{:end-row 2,
  :type :unused-excluded-var,
  :level :warning,
  :filename
  "test-regression/checkouts/metabase/src/metabase/lib/schema/util.cljc",
  :col 29,
  :end-col 32,
  :langs (),
  :message "Unused excluded var: ref",
  :row 2}
```

```clojure
(defn ref-distinct-key
  "For deduplicating refs: keep just the keys that are essential to distinguishing one ref from another."
  [ref]
  (let [options (lib.options/options ref)]
    (lib.options/with-options ref
      ;; Using reduce-kv to remove namespaced keys and some other keys to perform the comparison.
      (reduce-kv (fn [acc k _]
                   (if (or (qualified-keyword? k)
                           (#{:base-type :effective-type} k))
                     (dissoc acc k)
                     acc))
                 options options))))
```
---

```clojure
{:end-row 4,
  :type :unused-excluded-var,
  :level :warning,
  :filename
  "test-regression/checkouts/metabase/src/metabase/query_processor/middleware/resolve_joins.clj",
  :col 29,
  :end-col 34,
  :langs (),
  :message "Unused excluded var: alias",
  :row 4}
```

```clojure
(defn- source-metadata->fields [{:keys [alias], :as join} source-metadata]
  (when-not (seq source-metadata)
    (throw (ex-info (tru "Cannot use :fields :all in join against source query unless it has :source-metadata.")
                    {:join join})))
  (let [duplicate-ids (into #{}
                            (keep (fn [[item freq]]
                                    (when (> freq 1)
                                      item)))
                            (frequencies (map :id source-metadata)))]
    (for [{field-name :name
           base-type :base_type
           field-id :id
           field-ref :field_ref} source-metadata]
      ;; If `field-ref` is an id-based reference, only use it if the source query uses it.
      (or (when (and field-id (not (contains? duplicate-ids field-id)))
            (lib.util.match/match-one field-ref
              [:field (id :guard pos-int?) _opts]
              [:field field-id {:join-alias alias}]))
          [:field field-name {:base-type base-type, :join-alias alias}]))))
```

borkdude: @jramosg I'm looking at the first example.

```
{:end-row 2,
  :type :unused-excluded-var,
  :level :warning,
  :filename
  "test-regression/checkouts/metabase/src/metabase/lib/schema/util.cljc",
  :col 29,
  :end-col 32,
  :langs (),
  :message "Unused excluded var: ref",
  :row 2}
  ```
  
The code you list seems to be coming from a different file, since it's a test...?
  
In this file: https://github.com/metabase/metabase/blob/master/src/metabase/lib/schema/util.cljc I don't see a local (binding) named `ref`, nor do I see a var named ref in there.

The second map you mention is the same as the first? The code fragment is not something I can find in metabase. Is this made up by an LLM perhaps?

The third map: I looked at the code but the fragment I also can't find. `alias` doesn't seem to be a false positive here. What's going on?
https://github.com/metabase/metabase/blob/master/src/metabase/query_processor/middleware/resolve_joins.clj

Comment (jramosg) [reply src/clj_kondo/impl/linters.clj:716]: Sorry, I pasted wrong the first one (already corrected). It is in https://github.com/metabase/metabase/blob/2c01e5063b9998ab64c9ce1292c07aef778392fb/test/metabase/formatter/impl_test.clj#L241

These ones are the findings that appear when I comment `(map :name bindings)`.
I don't know why they don't appear in the repo, I see it  when I do `CLJ_KONDO_REGRESSION_UPDATE=true bb test:regression` and the repo is downloaded locally in `test-regression/checkouts/metabase`:

- This is the second one: 

Screenshot from 2025-12-31 07-23-15

- And this the third one

Screenshot from 2025-12-31 07-27-36

So why my local `test-regression/checkouts/metabase/src/metabase/lib/schema/util.cljc` is not the same as [the one I see remotely](https://github.com/metabase/metabase/blob/master/src/metabase/lib/schema/util.cljc)?

We also have some in clj-kondo
```clojure
 {:end-row 4,
  :type :unused-excluded-var,
  :level :warning,
  :filename "src/clj_kondo/impl/analysis.clj",
  :col 29,
  :end-col 36,
  :langs (),
  :message "Unused excluded var: ns-name",
  :row 4}
 {:end-row 3,
  :type :unused-excluded-var,
  :level :warning,
  :filename "src/clj_kondo/impl/analyzer.clj",
  :col 29,
  :end-col 36,
  :langs (),
  :message "Unused excluded var: ns-name",
  :row 3}
 {:end-row 3,
  :type :unused-excluded-var,
  :level :warning,
  :filename "src/clj_kondo/impl/analyzer/namespace.clj",
  :col 29,
  :end-col 36,
  :langs (),
  :message "Unused excluded var: ns-name",
  :row 3}
 {:end-row 3,
  :type :unused-excluded-var,
  :level :warning,
  :filename "src/clj_kondo/impl/analyzer/potemkin.clj",
  :col 29,
  :end-col 36,
  :langs (),
  :message "Unused excluded var: ns-name",
  :row 3}
 {:end-row 3,
  :type :unused-excluded-var,
  :level :warning,
  :filename "src/clj_kondo/impl/analyzer/usages.clj",
  :col 29,
  :end-col 36,
  :langs (),
  :message "Unused excluded var: ns-name",
  :row 3}
 {:end-row 713,
  :type :unused-binding,
  :level :warning,
  :filename "src/clj_kondo/impl/linters.clj",
  :col 51,
  :end-col 59,
  :langs (),
  :message "unused binding bindings",
  :row 713}
 {:end-row 3,
  :type :unused-excluded-var,
  :level :warning,
  :filename "src/clj_kondo/impl/namespace.clj",
  :col 29,
  :end-col 36,
  :langs (),
  :message "Unused excluded var: ns-name",
  :row 3}
```

For example. if I take the first one
```clojure
(ns clj-kondo.impl.analysis
  "Helpers for analysis output"
  {:no-doc true}
  (:refer-clojure :exclude [ns-name])
  (:require
   [clj-kondo.impl.overrides :as overrides]
   [clj-kondo.impl.utils :as utils :refer [assoc-some export-ns-sym select-some]]))
```
I see it is used as  a var in 

```clojure
(defn reg-namespace! [{:keys [analysis-ns-meta analysis base-lang lang] :as _ctx}
                      filename row col ns-name in-ns? metadata]
  (when analysis
    (swap! analysis update :namespace-definitions conj
           (assoc-some
            (cond-> (merge {:filename filename
                            :row row
                            :col col
                            :name ns-name}
                           metadata)
              analysis-ns-meta (-> (assoc :meta
                                          (cond-> (apply merge (:user-meta metadata))
                                            (not (true? analysis-ns-meta)) (select-keys analysis-ns-meta)))
                                   (dissoc :user-meta)))
            :in-ns (when in-ns? in-ns?) ;; don't include when false
            :lang (when (= :cljc base-lang) lang)))))
```

borkdude: OK, let's keep bindings in then, it's probably a good default. The difference could be explained because in the regression tests a specific SHA is used.

src/clj_kondo/impl/linters.clj
```diff
@@ -704,6 +705,26 @@
...
                        :unused-alias (str "Unused alias: " alias))))))))
 
+(defn lint-unused-excluded-vars! [ctx]
+  (when-not (linter-disabled? ctx :unused-excluded-var)
+    (doseq [{:keys [clojure-excluded] :as ns} (namespace/list-namespaces ctx)
+            :when (and (seq clojure-excluded)
+                       (not (linter-disabled? ns :unused-excluded-var)))
+            :let [{:keys [lang referred-vars vars bindings]} ns
+                  core-original-name (fn [[_ {var-name :name}]]
+                                       (when (var-info/core-sym? lang var-name)
```
borkdude: Can you explain this bit of code?

jramosg: It takes only the core symbols from `referred-vars`

```clojure
(keep core-original-name referred-vars)
```
but I realized that we already check if it is a core symbol in `:when` 
```clojure
:when (and (not (contains? used excluded))
           (var-info/core-sym? lang excluded))
```

so I simplified to:
```clojure
(map :name (vals referred-vars))
```

doc/linters.md
```diff
@@ -1617,6 +1618,18 @@ Example warning: `require with :refer`.
 {:linters {:refer {:exclude [clojure.set]}}}
 ```
 
+### Unused excluded var
+
+*Keyword:* `:unused-excluded-var`.
+
+*Description:* warns when `:refer-clojure :exclude` contains vars that are not redefined in the current namespace.
```
borkdude: The docs do not mention that locals also influence this linter.

src/clj_kondo/impl/config.clj
```diff
@@ -113,6 +113,7 @@
                                  ;; different from str
                                  :aliases {#_clojure.string #_str}}
               :unused-import {:level :warning}
+              :unused-excluded-var {:level :warning}
```
borkdude: Let's put the default level to `:info`

jramosg:
@borkdude I believe all the requested changes are done. Please let me know if I missed anything.

borkdude:
Thanks!

PR #2680, opened 2025-12-18
Add :cond-as-case linter

### LINKED ISSUE #2563: Suggest using `case` instead of `cond` and `condp` where it's an obvious possibility

It's a rather frequent occurrence for me when working on someone else's projects that I see code like this
```clj
(cond
  (= x :a) ...
  (= x :b) ...)
```
or like this
```clj
(condp = x
  :a ...
  :b ...)
```
or even this
```clj
(condp contains? x
  #{:a} ...
  #{:b :c} ...)
```

Sometimes the constants are keywords, sometimes not.

All of the examples above should be replaced with `case` since it's less verbose and turns linear look-ups into constant-times ones.

The only case where it's not that obvious that the replacement can be made is when the constants have hash collisions and the user sets `*warn-on-reflection*`:
```clj
(set! *warn-on-reflection* true)
=> true
(case nil 
  :fqzn 1
  :h3zn 2
  3)
Performance warning, [...] - hash collision of some case test constants; if selected, those entries will be tested sequentially.
=> 3
```

Same can happen with different types:
```clj
(case nil 
  0 0 
  nil 1 
  2)
Performance warning, [...]
```

I suppose there are two reasonable options:
1. Make clj-kondo check the hashes and warn about a possibility of using `case` only when there are no collisions
2. Make the check opt-in and let users manually mark particular usages of `cond`/`condp` as ignored by clj-kondo

jramosg:
# Fixes #2563

Add new linter that suggests replacing cond/condp with case when all test expressions are equality checks against the same variable with compile-time constants.

Detects three patterns:
- `(cond (= x :a) 1 (= x :b) 2)`
- `(condp = x :a 1 :b 2)`
- `(condp contains? x #{:a} 1 #{:b} 2)`

The linter avoids false positives by:
- Requiring at least 2 test constants
- Checking for hash collisions (0, 0.0, nil share the same hash)
- Only suggesting when all constants are case-compatible types

Default level is :off since this is a style suggestion rather than a bug detection, and to avoid noise in regression tests against real codebases.

## Notes
The default level is set to `:off` rather than `:warning` for regression test compatibility:

-  The codebase includes regression tests against real-world projects like Athens (re-frame-athens-lint-test). Setting the default to :warning would cause these tests to fail since real codebases commonly use cond/condp patterns that could technically be replaced with case.

## Future considerations
If there's interest, the linter could be extended to support more granular configuration, such as:

- Separate linters for cond and condp:

 ```clojure
{:linters {:cond-as-case {:level :off}
            :condp-as-case {:level :warning}}}
```

- Or configuration within the linter:

```clojure
{:linters {:cond-as-case {:level :warning
                          :include #{:cond :condp} ;; or  :exclude #{:condp}
                          }}}
```

Happy to implement either approach if maintainers think it would be useful.

### DIFF
--- a/doc/hooks.md
+++ b/doc/hooks.md
@@ -37,7 +37,8 @@ Other API functions:
 - `sexpr`: turns a node into a Clojure s-expression. Useful for analyzing concrete values.
 - `reg-keyword!`: indicates that a keyword's analysis should be marked as a
   definition. Expects the keyword node and either `true` or the fully-qualified
-  function that registered it. This can be used to implement keyword navigation for `clojure-lsp`.
+  function that registered it. Returns a new keyword node that must be in the transformed body.
+  This can be used to implement keyword navigation for `clojure-lsp`.
 - `tag`: returns node's tag, can be used as a dispatch function for multimethods, if you want your hook to be opened for extension.
 - `reg-finding!`: registers a finding. Expects a map with:
   - `:message`: the lint message


tomdl89:
I think there are more case-compatible types than you have accounted for. For example:
```clojure
user> (condp = [:a]
        [:b] :bar
        [:a] :foo
        :qux)
:foo
user> (case [:a]
        [:b] :bar
        [:a] :foo
        :qux)
:foo
```
There are others too, like vectors, sets, and maps of constants. Obvs, best to err on the side of false negatives than false positives, but maybe worth a look.

jramosg:
@tomdl89 you are right, I tried to add more case-compatible types including nested structures.
Thank you!

src/clj_kondo/impl/linters.clj
```diff
@@ -35,33 +35,130 @@
...
+  (or (some #(% v) [keyword? number? string? char? nil? true? false? symbol?])
+      (and (some #(% v) [vector? set? list?])
+           (every? case-testable? v))
+      (and (map? v)
+           (every? case-testable? (keys v))
+           (every? case-testable? (vals v)))))
+
+(defn- has-hash-collisions? [constants]
+  (let [hashes (map hash constants)]
+    (not= (count hashes) (count (set hashes)))))
```
borkdude: you can replace this with `distinct?`

src/clj_kondo/impl/linters.clj
```diff
@@ -35,33 +35,130 @@
...
+
+(defn- parse-equality-condition [cond-node]
+  (when (= :list (tag cond-node))
+    (let [children (:children cond-node)]
+      (when (>= (count children) 3)
+        (let [[op arg1 arg2] children]
+          (when (= '= (sexpr op))
```
borkdude: This may not be clojure.core's `=` but could also be an overridden `=`

borkdude: Also: prefer to not call `sexpr` when possible since `sexpr` may throw or be an expensive call. To get the symbol from a node you can call `:value` on it.

src/clj_kondo/impl/linters.clj
```diff
@@ -35,33 +35,130 @@
...
+            (cond
+              (and (constant? arg2) (not (constant? arg1)))
```
borkdude, line 78: Why do we need the sexpr of a non-constant?

src/clj_kondo/impl/linters.clj
```diff
@@ -45,7 +45,7 @@
 
 (defn- has-hash-collisions? [constants]
   (let [hashes (map hash constants)]
-    (not= (count hashes) (count (set hashes)))))
+    (not= (count hashes) (count (distinct hashes)))))
```
borkdude: I mean: `(distinct? hashes)`

doc/linters.md
```diff
@@ -250,6 +251,52 @@ enabling this linter, you can prepend the `case` expression with
...
 
+### Cond-as-case
+
+*Keyword:* `:cond-as-case`.
+
+*Description:* warn on `cond` or `condp` expressions that can be replaced with `case`.
+This improves both readability and performance since `case` uses constant-time lookup
+instead of linear search.
+
+*Default level:* `:off`.
```
jramosg, line 262: default `off` is okay @borkdude ? maybe `:info`?

src/clj_kondo/impl/linters.clj
```diff
@@ -35,33 +35,140 @@
...
+                  (let [test-val (sexpr test-node)]
+                    (when (case-testable? test-val)
+                      (recur (rest rest-clauses) (conj constants test-val)))))))))))))
+
+(defn- extract-condp-contains-pattern [expr]
+  (let [children (:children expr)]
+    (when (>= (count children) 4)
+      (let [[_ pred-node _var-node & clauses] children
+            pred-sexpr (sexpr pred-node)]
+        (when (= 'contains? pred-sexpr)
```
borkdude: The same remark I made about `=` also goes for this one.

Also: people can write:

```
(ns foo (:require [clojure.core :as core]))

(core/contains? #{1 2 3} 1)
```

So just checking the `contains?` symbol isn't robust enough.

This also goes for `=`.

borkdude: I'm fine with the symbol check while ignoring the `core/contains?` case though.

But I think we do need to check additionally if `=` and `contains?` weren't redefined in the current ns by the user.

jramosg:
@borkdude I think that now every check is handled:
```clojure
(defn- core-symbol? [ctx sym-value core-sym-name]
  (let [lang (:lang ctx)
        caller-ns (:caller-ns ctx)
        core-ns (if (= :cljs lang)
                  'cljs.core 'clojure.core)]
    (or
     ;; Unqualified symbol that's not excluded
     (and (symbol? sym-value)
          (= core-sym-name (name sym-value))
          (nil? (namespace sym-value))
          (not (get-in caller-ns [:clojure-excluded (symbol core-sym-name)])))
     ;; Fully qualified with clojure.core or cljs.core
     (and (symbol? sym-value)
          (= (symbol (str core-ns "/" core-sym-name)) sym-value))
     ;; Qualified with an alias for clojure.core
     (when (and (symbol? sym-value) (namespace sym-value))
       (let [ns-alias (symbol (namespace sym-value))
             sym-name (name sym-value)]
         (and (= core-sym-name sym-name)
              (= core-ns (get-in caller-ns [:qualify-ns ns-alias]))))))))
```

src/clj_kondo/impl/linters.clj
```diff
@@ -35,33 +35,154 @@
...
+      (and (some #(% v) [vector? set? list?])
+           (every? case-testable? v))
+      (and (map? v)
+           (every? case-testable? (keys v))
+           (every? case-testable? (vals v)))))
+
+(defn- all-hashes-distinct? [constants]
+  (apply distinct? (map hash constants)))
+
+(defn- core-symbol? [ctx sym-value core-sym-name]
```
borkdude, line 49: `=` could also be a local like:

```
(let [= my-equals])
```

but at this point we're probably re-implementing `resolve-name` from `analyzer/namespace.clj`

doc/linters.md
borkdude: I think defaulting to `:off` would be safest for now given the complexity and chance of bugs. But maybe you can report in a comment which regression findings come up when you enable it.

jramosg:
This are some findings. There are more, but every finding I looked at looks like a true positive for me

```clojure
{:end-row 109,
  :type :cond-as-case,
  :level :info,
  :filename
  "test-regression/checkouts/metabase/src/metabase/activity_feed/models/recent_views.clj",
  :col 45,
  :end-col 92,
  :langs (),
  :message "cond can be replaced with case",
  :row 107}
```
```clojure
(cond (= model :card) (h2x/literal "question")
      ;; TODO add metric
      (= model :dataset) (h2x/literal "model"))
```

---

```clojure
{:end-row 46,
  :type :cond-as-case,
  :level :info,
  :filename
  "test-regression/checkouts/metabase/src/metabase/api/common/internal.clj",
  :col 6,
  :end-col 12,
  :langs (),
  :message "condp can be replaced with case",
  :row 37}
```
```clojure
(condp = schema-type
  ;; can use any regex directly
  :re       (first (mc/children schema))
  :keyword  #"[\S]+"
  'pos-int? #"[0-9]+"
  :int      #"-?[0-9]+"
  'int?     #"-?[0-9]+"
  :uuid     u/uuid-regex
  'uuid?    u/uuid-regex
  nil)
```

---

```clojure
{:end-row 35,
  :type :cond-as-case,
  :level :info,
  :filename
  "test-regression/checkouts/metabase/src/metabase/cloud_migration/api.clj",
  :col 9,
  :end-col 25,
  :langs (),
  :message "condp can be replaced with case",
  :row 32}
```
```clojure
(condp = (-> e ex-data :status)
  404 {:status 404 :body "Could not establish a connection to Metabase Cloud."}
  400 {:status 400 :body "Cannot migrate this Metabase version."}
  {:status 500})
```

---
```clojure
{:end-row 42,
  :type :cond-as-case,
  :level :info,
  :filename "src/aaaa_this_has_to_be_first/pprint.clj",
  :col 27,
  :end-col 36,
  :langs (),
  :message "condp can be replaced with case",
  :row 39}
```
```clojure
(condp = optval
  nil (java.io.StringWriter.)
  true *out*
  optval)
```

---

```clojure
 {:end-row 37,
  :type :cond-as-case,
  :level :info,
  :lang :clj,
  :filename "sci/impl/doseq_macro.cljc",
  :col 24,
  :end-col 59,
  :cljc true,
  :langs (:clj :cljs),
  :message "cond can be replaced with case",
  :row 28}
```
```clojure
(cond
  (= k :let) [needrec `(let ~v ~subform)]
  (= k :while) [false `(~'when ~v
                               ~subform
                               ~@(when needrec [recform]))]
  (= k :when) [false `(if ~v
                        (do
                          ~subform
                          ~@(when needrec [recform]))
                        ~recform)])
```

borkdude:
I notice that when mixing keywords, symbols and/or numbers, case acts differently.

```
user=> (macroexpand '(case x 1 2 foo 3 :bar 4))
(let* [G__139 x] (case* G__139 0 3 (throw (java.lang.IllegalArgumentException. (clojure.core/str "No matching clause: " G__139))) {0 [foo 3], 1 [1 2], 3 [:bar 4]} :compact :hash-equiv nil))
user=> (macroexpand '(case x :bar 4))
(let* [G__142 x] (case* G__142 0 0 (throw (java.lang.IllegalArgumentException. (clojure.core/str "No matching clause: " G__142))) {1020205863 [:bar 4]} :compact :hash-identity nil))
```

Does this mean that when mixing different types, it falls back to `=`?

There might be cases where `case` doesn't even end. up faster than `condp =`. So maybe we should do more research on this before merging this.

PR #2684, opened 2025-12-20
Add ratio type support for numerator and denominator

jramosg:
# Add :ratio type support for numerator and denominator functions
## Summary
This PR adds comprehensive :ratio type support to clj-kondo's type system, enabling type checking for the numerator and denominator functions.

## Changes

- Added :ratio as a new type in the type system
- Added to known-types set
- Defined type relations: `:ratio` is-a `:number` 
- Added to could-be-relations so `:number` could-be `:ratio`
- Added user-friendly label: "ratio"
- Added Java type mapping for clojure.lang.Ratio
- Updated number->tag function to recognize ratio literals (1/2, 3/5, etc.)
- Updated type signatures for numerator and denominator:  
- - Both functions now correctly require `:ratio` argument type
- - Both return `:number` (matching Clojure's behavior)
- Added comprehensive test coverage

## Why (numerator (/ 1 1)) doesn't warn
This is intentional and static analysis behavior:

- The / operator returns type` :number`, which can be an integer, ratio, or double depending on the operands
- Static analysis cannot evaluate division at compile time - we can't know `if (/ x y)` will produce a ratio or integer without running the code
- `:number `could be `:ratio`, since `:ratio` is a subtype of `:number`, the type checker correctly allows it
- Avoiding false positives: warning here would incorrectly flag legitimate code like `(numerator (/ 3 2))` which does produce a ratio at runtime
- While `(denominator (/ 1 1))` will throw a ClassCastException at runtime (because `(/ 1 1)` evaluates to 1, not a ratio), this is a runtime error that static analysis cannot prevent without actually evaluating the division. The trade-off favors avoiding false positives in real code.

--- 

### DIFF
--- a/src/clj_kondo/impl/types/clojure/core.clj
+++ b/src/clj_kondo/impl/types/clojure/core.clj
@@ -789,10 +789,10 @@
    ;; 4745 'prn-str
    ;; 4754 'print-str
    ;; 4763 'println-str
-   ;; 4794 'ex-info
-   'ex-info {:arities {2 {:args [:nilable/string :map]
+   ;; 4794
+   'ex-info {:arities {2 {:args [:nilable/string :nilable/map]
                           :ret :throwable}
-                       3 {:args [:nilable/string :map :any]
+                       3 {:args [:nilable/string :nilable/map :any]
                           :ret :throwable}}}
    ;; 4803 'ex-data
    ;; 4800 'ex-message
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,9 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
-- [#2668](https://github.com/clj-kondo/clj-kondo/issues/2668): Lint `&` syntax errors in let bindings and lint for trailing `&` ([@tomdl89](https://github.com/tomdl89)) 
 - [#2654](https://github.com/clj-kondo/clj-kondo/issues/2654): NEW linter: `redundant-let-binding`, defaults to `:off` ([@tomdl89](https://github.com/tomdl89))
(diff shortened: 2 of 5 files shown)


src/clj_kondo/impl/types.clj
```diff
@@ -205,6 +208,7 @@
                      ;; can be auto-cast to
                      :int}
     (Float Double java.lang.Float java.lang.Double) #{:double :float :int :nil}
+    (ratio?) :ratio
```
borkdude, line 211: This appears non-sensical to me. `ratio?` is a predicate, not a type hint. AFAIK there usually isn't a type hint for ratios in Clojure.

src/clj_kondo/impl/types/clojure/core.clj
```diff
@@ -23,6 +23,9 @@
 
 (def any->boolean {:arities {1 {:ret :boolean}}})
 
+(def ratio->number {:arities {1 {:args [:ratio]
+                                 :ret :number}}})
```
borkdude, line 27: The result of numerator and denominator is always a `java.math.BigInteger`. Is that the reason you went with `:number` here? Which is fine, just double checking.

jramosg: you are right, nat-int would be better

borkdude: nope:

```
user=> (nat-int? (numerator 1/2))
false
user=> (nat-int? (denominator 1/2))
false
```

borkdude: Let's keep the `:number`

jramosg: Oh, I didn't expect that, so perfect, I leave number

borkdude:
Can you fix merge conflicts? Then I'll merge

PR #2693, opened 2025-12-24
Lint quoted forms in call position

### LINKED ISSUE #2692: Lint quoted forms in call position
Currently, booleans, strings, characters and numbers are registered as "not a function" when in call position. E.g.

```clojure
(lint! "(false 42)")
;; => ({:file "<stdin>", :row 1, :col 1, :level :error, :message "a boolean is not a function"})
```

but quoted forms, such as any of these plus quoted lists, are not registered:

```clojure
(lint! "('(foo) 1)")
;; => ()
```

We should check for this. And give an error, like so:

```clojure
(lint! "('(foo) 1)")
;; => ({:file "<stdin>", :row 1, :col 3, :level :error, :message "a list is not a function"})

(lint! "('42 42)")
;; => ({:file "<stdin>", :row 1, :col 3, :level :error, :message "a number is not a function"})
```

PR incoming :)

tomdl89:
Fixes #2692

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ For a list of breaking changes, check [here](#breaking-changes).
 ## Unreleased
 - [#2692](https://github.com/clj-kondo/clj-kondo/issues/2692): Lint quoted forms which are not functions ([@tomdl89](https://github.com/tomdl89))
 
+- Add type checking support for `sorted-map-by`, `sorted-set`, and `sorted-set-by` functions
+([@jramosg](https://github.com/jramosg))
+
 ## 2025.12.23
 
 - [#2654](https://github.com/clj-kondo/clj-kondo/issues/2654): NEW linter: `redundant-let-binding`, defaults to `:off` ([@tomdl89](https://github.com/tomdl89))
--- a/src/clj_kondo/impl/types.clj
+++ b/src/clj_kondo/impl/types.clj
@@ -36,6 +36,7 @@
     :map
     :nil
     :set
+    :sorted-set
     :fn
     :keyword
     :symbol
@@ -75,6 +76,7 @@
    :nil #{:seqable}
    :coll #{:seqable}
    :set #{:seqable :coll :ifn :ilookup}
+   :sorted-set #{:set :seqable :coll :ifn :ilookup}
    :fn #{:ifn}
    :keyword #{:ifn}
    :symbol #{:ifn}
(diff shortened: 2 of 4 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -3339,27 +3364,9 @@
...
-                                            (rest children)))
-                      (utils/number-token? function)
-                      (do (reg-not-a-function! ctx expr "number")
-                          (analyze-children (update ctx :callstack conj [nil t])
-                                            (rest children)))
-                      :else
-                      (do
-                        (types/add-arg-type-from-expr (update ctx :callstack conj [nil t]) expr)
-                        (analyze-children ctx children)))))
+                    (when (false? (analyze-non-symbol-function ctx children t function))
```
borkdude, line 3367: Why is this when false? here? Doesn't this prevent linting children? Can't tell without reading the code in my editor, the layout in Github is wrapped.

tomdl89: `analyze-non-symbol-function` calls `reg-not-a-function` and `analyze-children` for all cases, except for the `:else` case, where it returns `false`. The two callsites for `analyze-non-symbol-function` handle the `false` case slightly differently: the one in the `:quote` branch just calls `analyze-children` but the one in the `:token` branch calls `types/add-arg-type-from-expr` _and_ `analyze-children`. But in all cases, `analyze-children` is called. I needed to use `false` (rather than the implicit `nil` of `cond` not matching any conditions) because `analyze-children` (the last form of most of the `cond` branches) can return `nil`. Tbh I could have returned anything as a sentinel in the `:else` branch, but `false` felt right.

borkdude:
When I print the expression being linted in `analyze-expression**` I see that some nodes are now analyzed multiple times.

```
$ clj -M:clj-kondo/dev --lint - <<< "('1 2)"
:expr <list: ('1 2)>
:expr <token: 2>
:expr <quote: '1>
:expr <token: 1>
:expr <token: 2>
<stdin>:1:3: error: a number is not a function
linting took 28ms, errors: 1, warnings: 0
```

Why does `(analyze-non-symbol-function ctx children t quoted-child)` analyze its children?

In general: can you keep your code changes in PRs as small as possible? Any change that doesn't relate to the problem at hand, should be avoided in order to make reviewing the PR less time consuming.

tomdl89:
I'll take a look. In general, I take your point, but this isn't an example of me doing more than necessary (on purpose at least!). I pulled the logic out into a function because it would have been duplicated otherwise. Hopefully I'll get some time soon. Cheers

borkdude:
OK, got it. One other piece of feedback: the `analyze-non-symbol-function` contains a fair amount of duplication, perhaps it can be shorted. Just the `tag` seems to be different for each branch.

tomdl89:
Upon further reflection, I decided a little bit of duplication is preferable to this confusing abstraction. Closing this in favour of #2718

PR #2696, opened 2025-12-25
Fix config-in-call to disable unquote-not-syntax-quoted

### LINKED ISSUE #2695: `:unquote-not-syntax-quoted` triggers false positives in `project.clj` when using defproject with shared version vars

**version**

clj-kondo v2025.12.23

**macro usage**

Yes. This seems related to `leiningen`’s `defproject` macro and unquote usage in `project.clj`.

**platform**

Both native version and on the JVM.

**editor**

Editor: IntelliJ
Plugin: Cursive

But also plain cli.

**problem**

After upgrading to a `clj-kondo` version that includes the new `:unquote-not-syntax-quoted` linter, I’m getting warnings in `project.clj` for a common Leiningen pattern: defining a shared version var and reusing it in dependencies via `~`.  

Example pattern:
```clojure
(def proto-version "1.2.3")

(defproject my-project "0.1.0"
  :dependencies [[some/lib ~proto-version]
                 [other/lib ~proto-version]])
```

**repro**

Minimal repo repro:
https://github.com/evg-tso/cursive-shrink-issue/blob/master/project.clj

**expected behavior**

- `:unquote-not-syntax-quoted` should not warn for valid `~` usage inside `defproject` / `project.clj`

jramosg:
Update the linter configuration to correctly disable the `:unquote-not-syntax-quoted` linter in specific calls, such as `defproject` in `project.clj`. This change ensures that users can safely use unquote syntax without triggering unnecessary warnings when configured appropriately.

Fixes #2695

--- 

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -2461,19 +2461,8 @@
             resolved-var-sym (symbol (str resolved-namespace) (str resolved-name))
             cfg (when-let [in-call-cfg (:config-in-call config)]
                   (or (get in-call-cfg resolved-var-sym)
-                      (when (and (or (not resolved-namespace)
-                                     (= :clj-kondo/unknown-namespace resolved-namespace))
-                                 (symbol? full-fn-name)
-                                 (not (namespace full-fn-name)))
-                        (or (get in-call-cfg full-fn-name)
-                            (let [matches (filter (fn [[k _]]
-                                                    (and (symbol? k)
-                                                         (namespace k)
-                                                         (= (name k)
-                                                            (name full-fn-name))))
-                                                  in-call-cfg)]
-                              (when (= 1 (count matches))
-                                (second (first matches))))))))
+                      (and (= :clj-kondo/unknown-namespace resolved-namespace)
+                           (in-call-cfg (symbol "user" (str resolved-name))))))
             cfg (when cfg
                   (config/expand-ignore cfg))
             ctx (if cfg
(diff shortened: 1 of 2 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -2460,7 +2460,20 @@
             expr-meta (meta expr)
             resolved-var-sym (symbol (str resolved-namespace) (str resolved-name))
             cfg (when-let [in-call-cfg (:config-in-call config)]
-                  (get in-call-cfg resolved-var-sym))
+                  (or (get in-call-cfg resolved-var-sym)
```
borkdude, line 2463: Can you explain what these changes do? I don't think `:config-in-call` with an unqualified symbol should work.

jramosg: what about using `user` as default namespace when `resolved-namespace` is `:clj-kondo/unknown-namespace`? or the problem comes from `resolved-namespace` beeing `:clj-kondo/unknown-namespace`?

borkdude: Yeah, this is a bit messy. 

Perhaps we should change the existing config in to `leiningen.core.project/defproject` and automatically set this namespace in `project.clj` so people can then use `leiningen.core.project/defproject` in configurations everywhere. `user/defproject` is basically nonsense.

borkdude: I'll change this in a different PR.

borkdude:
Fixing this in another PR. Thanks

PR #2697, opened 2025-12-29
Add type checking for Clojure array functions

jramosg:
This PR adds type checking support for Clojure's array manipulation functions: `to-array`, `alength`, `aget`, `aset`, and `aclone`. These functions are now properly typed in the clj-kondo type system, allowing for better static analysis and error detection when using arrays in Clojure code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 - Add type checking support for `sorted-map-by`, `sorted-set`, and `sorted-set-by` functions
 ([@jramosg](https://github.com/jramosg))
+- Add new type `array` and type checking support for the next functions: `to-array`, `alength`, 
+`aget`, `aset` and `aclone` ([@jramosg](https://github.com/jramosg))
 
 ## 2025.12.23
 


test-regression/clj_kondo/clj_kondo/findings.edn
```diff
@@ -17545,6 +17545,15 @@
   :langs (),
   :message "unused binding name",
   :row 170}
+ {:end-row 183,
```
borkdude, line 17548: Have you checked what code this warning related to and if it was a true positive or a false positive?

jramosg: Yes, it comes from here
[https://github.com/clojure/clojurescript/blob/master/src/main/cljs/clojure/browser/repl.cljs](https://github.com/clojure/clojurescript/blob/master/src/main/cljs/clojure/browser/repl.cljs)
```clojure
(def load-queue nil)

(defn bootstrap
  "Reusable browser REPL bootstrapping. Patches the essential functions
  in goog.base to support re-loading of namespaces after page load."
  []
  ;; Monkey-patch goog.provide if running under optimizations :none - David
  (when-not js/COMPILED
    (set! (.-require__ js/goog) js/goog.require)
    ;; suppress useless Google Closure error about duplicate provides
    (set! (.-isProvided_ js/goog) (fn [name] false))
    ;; provide cljs.user
    (goog/constructNamespace_ "cljs.user")
    (set! (.-writeScriptTag__ js/goog)
      (fn [src opt_sourceText]
        ;; the page is already loaded, we can no longer leverage document.write
        ;; instead construct script tag elements and append them to the body
        ;; of the page, to avoid parallel script loading enforce sequential
        ;; load with a simple load queue
        (let [loaded (atom false)
              onload (fn []
                       (when (and load-queue (false? @loaded))
                         (swap! loaded not)
                         (if (zero? (alength load-queue))
                           (set! load-queue nil)
                           (.apply js/goog.writeScriptTag__ nil (.shift load-queue)))))]
```
We cannot do `(alength nil)`

```clojure
(alength nil) ;=>
Execution error (TypeError) at (<cljs repl>:1).
Cannot read properties of null (reading 'length')
```

PR #2702, opened 2025-12-30
Add new linter for eager :or expressions in mapping in destructuring

### LINKED ISSUE #2459: Lint "improper" use of  :or mapping in map destructuring forms

- [ ] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

I have encountered in multiple project issues caused by the developer's expectation that the `:or` in destructuring behaves lazily like an `or`, meaning it stops evaluating at the first truthy expression. It actually evals the default value regardless:

```clojure
user> (defn f1 [] (println "WOW YOU DON'T WANT THIS TO RUN"))
#'user/f1
user> (let [{:keys [x] :or {x (f1)}} {:x 1}] x)
WOW YOU DON'T WANT THIS TO RUN
1
```

I think clj-kondo should mark the use of s-expressions as default value in the :or mapping of map destructuring forms as an error.

jramosg:
This commit introduces a new linter `:destructured-or-always-evaluates` that warns when s-expressions are used as defaults in map destructuring. This change aims to prevent unexpected side effects due to eager evaluation of defaults, enhancing code safety and clarity.

Fixes #2459

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -1016,7 +1016,7 @@ expression that always evaluates, e.g. a function call.
 (let [{:keys [x] :or {x (f1)}} {:x 1}] x)
 ```
 
-*Example message:* `Default :or value is always evaluated`
+*Example message:* `Default :or value is always evaluated.`
 
 ### Invalid arity
 
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -2473,7 +2473,7 @@
            (assoc expr-meta
                   :type :destructured-or-always-evaluates
                   :filename (:filename ctx)
-                  :message "Default :or value is always evaluated")))
+                  :message "Default :or value is always evaluated.")))
         (cond unresolved-ns
               (let [fn-name (-> full-fn-name name symbol)]
                 (namespace/reg-unresolved-namespace! ctx ns-name
--- a/test/clj_kondo/destructured_or_always_evaluates_test.clj
+++ b/test/clj_kondo/destructured_or_always_evaluates_test.clj
@@ -12,24 +12,24 @@
         :row 1
         :col 25
         :level :warning
-        :message "Default :or value is always evaluated"})
+        :message "Default :or value is always evaluated."})
      (lint! "(let [{:keys [x] :or {x (f1)}} {:x 1}] x)"
             config)))
   (testing "multiple calls in :or mappings"
     (assert-submaps
      '({:row 1
         :col 27
-        :message "Default :or value is always evaluated"}
+        :message "Default :or value is always evaluated."}
        {:row 1
         :col 34
-        :message "Default :or value is always evaluated"})
+        :message "Default :or value is always evaluated."})
      (lint! "(let [{:keys [x y] :or {x (f1) y (f2)}} {}] [x y])"
             config)))
   (testing "nested map destructuring"
     (assert-submaps
      '({:row 1
         :col 26
-        :message "Default :or value is always evaluated"})
+        :message "Default :or value is always evaluated."})
      (lint! "(let [{{:keys [b] :or {b (f1)}} :a} {}] b)"
             config)))
   (testing "vector destructuring (should not trigger)"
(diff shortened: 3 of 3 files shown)


borkdude:
Thanks. Some thoughts.

- I think the name could be improved: "improper" doesn't tell you much. The issue is that some work or side effect is happening in the right hand side of the or clause that may go unnoticed. Proposals?
- A list doesn't always mean something is executed at runtime: one notable exception is the empty list `()`. Another example could be a macro which just expands in an expression (we can ignore this for now I guess). 
- There may be collection literals that contain side effects. E.g. `{:or {a {x (launch-rocket)}}}`. This is probably uncommon but they may exist.
- Given the last two: it may be better to trigger the warning inside `analyze-call`, now we know for sure that a call or side effect is happening, despite the or expression being a list or not. To do this, we could add some flag to the context to tell we're inside an or default.

False positives?

- The following case may be considered a false positive - maybe. When is a calculation cheap enough to be "proper" for an or default? https://github.com/babashka/sci/blob/47e48bbc74facf0c649b4245c616eb69fbf7f57e/src/sci/impl/opts.cljc#L212 

Have you looked at the other findings and found similar things? Linking them here may be valuable so we can triage them.

seancorfield:
Maybe `:eager-or-expression`?

jramosg:
@borkdude about false positives I don't know if we can have a rule to know if a calculation is cheap enough. Sure that `(str (gensym "inline"))` or `{:or {load-fn (:load-fn env)}}` are cheap but how can we put this as a rule?
Given this, I think it's reasonable to emit a warning, leaving it up to the user to decide whether to ignore or suppress it. 
Let me know if you think we could stablish a rule to set if it is cheap enough or not.

About the findings,  I found two in this repo which are clearly cheap, and I supressed them in this PR: 

- [https://github.com/clj-kondo/clj-kondo/pull/2702/files#diff-19afea30102cec0da2c4d0516d5e834219f6ccaadc2019a9b86e5af604050425](https://github.com/clj-kondo/clj-kondo/pull/2702/files#diff-19afea30102cec0da2c4d0516d5e834219f6ccaadc2019a9b86e5af604050425)
- https://github.com/clj-kondo/clj-kondo/pull/2702/files#diff-1939d4ee2643086d4da4bd5c3d1c1941bd77e5a141e6c64dd8b03521e86a5755


Other ones I found in Metabase for example are:

- https://github.com/metabase/metabase/blob/2c01e5063b9998ab64c9ce1292c07aef778392fb/src/metabase/channel/render/preview.clj#L103 
- https://github.com/metabase/metabase/blob/2c01e5063b9998ab64c9ce1292c07aef778392fb/src/metabase/embedding_rest/api/common.clj#L406 and https://github.com/metabase/metabase/blob/2c01e5063b9998ab64c9ce1292c07aef778392fb/src/metabase/embedding_rest/api/embed.clj#L88 (if they are cheap or not depends on `(qp.constraints/default-query-constraints)`, which I suppose it actually is)
- https://github.com/metabase/metabase/blob/2c01e5063b9998ab64c9ce1292c07aef778392fb/src/metabase/legacy_mbql/util.cljc#L144 
- https://github.com/metabase/metabase/blob/2c01e5063b9998ab64c9ce1292c07aef778392fb/src/metabase/parameters/shared.cljc#L161

About the side effects I have the same doubt, how can we know that the called function has side effects? Should not be warned and ignored manually in that case?

borkdude:
Yeah, I don't think we can check if a computation is cheap so let's keep it simple.
I do think we should either turn this linter off by default since not all people would be interested in this?

About the message:

> Default value in :or is an s-expression and will be eagerly evaluated.

I think this can be improved. Everything is an s-expression: numbers, data literals, and lists.
Proposal:

> Default :or value is eagerly evaluated

borkdude:
We already have one other linter around `:or`:

```
:destructured-or-binding-of-same-map
```

Perhaps we can start the name of this linter also with `:destructured-or` and then add `-default-always-evaluates`?

```
:destructed-or-default-always-evaluates
```

PR #2703, opened 2025-12-30
Add linter for imported but not required namespaces

### LINKED ISSUE #2538: Warn about imported-but-not-required Clojure-generated classes (`deftype`, etc)
It's possible for namespace `A` to refer to a var from namespace `B` without requiring it, but this will only work if Clojure happens to have already loaded `B`. This is brittle, which is why kondo has the lint rule `:unresolved-namespace`.

A similar footgun exists for Clojure-defined Java classes (e.g. `deftype`), when imported but not required:

```clojure
;;; bar.clj
(ns bar)
(deftype Bar [])

;;; foo.clj
(ns foo (:import (bar Bar)))
(Bar.)
```

Loading `foo` will succeed only if `bar` has already been loaded. I would like kondo to warn me about this, either via `:unresolved-namespace` or a new lint rule, to avoid this brittleness.

jramosg:
This change introduces a new linter that warns when a Clojure-defined Java class is imported but the corresponding namespace is not required. The linter is added to the configuration and documented in the linters overview. Tests are included to verify its functionality and behavior in various scenarios.

Fixes #2538

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#2779](https://github.com/clj-kondo/clj-kondo/issues/2779), [#2780](https://github.com/clj-kondo/clj-kondo/issues/2780), [#2781](https://github.com/clj-kondo/clj-kondo/issues/2781), [#2782](https://github.com/clj-kondo/clj-kondo/issues/2782), [#2783](https://github.com/clj-kondo/clj-kondo/issues/2783), [#2785](https://github.com/clj-kondo/clj-kondo/issues/2785), [#2786](https://github.com/clj-kondo/clj-kondo/issues/2786), [#2794](https://github.com/clj-kondo/clj-kondo/issues/2794): Performance optimizations.
 - Performance: cache hook-fn lookups to avoid repeated SCI evaluation
 - [#2621](https://github.com/clj-kondo/clj-kondo/issues/2621): load imports from symlinked config dir ([@walterl](https://github.com/walterl))
+- [#2798](https://github.com/clj-kondo/clj-kondo/issues/2798): report correct filename and error details when `StackOverflowError` occurs during analysis
 
 
 ## 2026.01.19
--- a/corpus/stackoverflow_hook/.clj-kondo/config.edn
+++ b/corpus/stackoverflow_hook/.clj-kondo/config.edn
@@ -1,0 +1,1 @@
+{:hooks {:macroexpand {foo/my-macro foo/my-macro}}}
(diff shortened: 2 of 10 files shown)


borkdude:
- It seems this linter warns at the place of usage, not at the place where the class is imported. Is that intentional? 
- And should this be related to import at all since people can also use the class without importing it, using it by global name?
- Are hyphens in namespaces accounted for?

test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -1918,6 +1918,18 @@
   :langs (),
   :message "Format string contains no format specifiers",
   :row 123}
+ {:end-row 30,
```
jramosg, line 1921: true positive, last line import `metabase.server.streaming_response` is not required 
```clj
(ns metabase.test.http-client
  "HTTP client for making API calls against the Metabase API. For test/REPL purposes."
  (:require
   [clj-http.client :as http]
   [clojure.core.async :as a]
   [clojure.edn :as edn]
   [clojure.java.io :as io]
   [clojure.string :as str]
   [clojure.test :refer :all]
   [java-time.api :as t]
   [malli.core :as mc]
   [medley.core :as m]
   [metabase.config.core :as config]
   [metabase.server.middleware.session :as mw.session]
   [metabase.server.test-handler :as server.test-handler]
   [metabase.test-runner.assert-exprs :as test-runner.assert-exprs]
   [metabase.test.initialize :as initialize]
   [metabase.util :as u]
   [metabase.util.date-2 :as u.date]
   [metabase.util.json :as json]
   [metabase.util.log :as log]
   [metabase.util.malli :as mu]
   [metabase.util.malli.humanize :as mu.humanize]
   [metabase.util.malli.registry :as mr]
   [metabase.util.malli.schema :as ms]
   [peridot.multipart]
   [ring.util.codec :as codec])
  (:import
   (java.io ByteArrayInputStream InputStream)
   (metabase.server.streaming_response StreamingResponse)))
```

jramosg:
I moved the location to the import, I think it's better there.
Also I managed to find the ones that are fully-qualified without imports. Maybe we should change the linter name with this change?

borkdude:
Yeah, let's change the linter name. Any suggestions?

jramosg:
`unrequired-namespace`? `class-used-without-require`?

NoahTheDuke:
:unnecessary-qualified-class?

borkdude:
@NoahTheDuke I think that's a different issue/PR than this one

jramosg:
other option could be to expand `unresolved-namespace`

borkdude:
Maybe `:missing-type-require` ?

PR #2716, opened 2026-01-06
Add redundant format linter and tests

### LINKED ISSUE #2712: :format linter doesn't handle `logf` correctly
The `:format` linter correctly handles `errorf`, `infof` and `debugf` (in that it knows that the format string can be the 0th or 1st arg, due to the optional `Throwable` arg). But it also handles `logf` using the exact same logic, even though `logf` takes one additional argument. This means that:

```clojure
(errorf e "id: %s") ;; "Format string expects 1 argument instead of 0"
(logf :error e "id: %s") ;; no warning
```

The second line above should issue the same warning as the first line.

jramosg:
Introduce a new linter `:redundant-format` to warn when format strings contain no format specifiers. This change includes updates to the analyzer, configuration, and documentation. Additionally, tests have been added to ensure the linter functions correctly for various cases, including formats with and without placeholders.

Fixes #2712


src/clj_kondo/impl/analyzer.clj
```diff
@@ -2026,7 +2026,13 @@
       (findings/reg-finding! ctx
                              (node->line (:filename ctx) format-str-node :format
                                          (format "Format string expects %s arguments instead of %s."
-                                                 percent-count arg-count))))))
+                                                 percent-count arg-count))))
+    (when (and (zero? percent-count)
```
tomdl89, line 2030: If `(not= percent-count arg-count)` and `(zero? percent-count)` won't we get double findings? Maybe that's a bit overkill. E.g. for this:

```clojure
(format "foo" 42)
```

we'd get a `Format string expects 0 arguments instead of 1` and a `Format string contains no format specifiers` finding. Which is really saying the same thing.

I haven't run the code, so I may be wrong. But it does seem that way.

jramosg: You are right, I will change it and leave only the "expects x arguments"
Thanks!

jramosg: done!

borkdude:
Thanks @jramosg . Thanks for reviewing @tomdl89 !

PR #2718, opened 2026-01-07
Register list in call position as not a function

### LINKED ISSUE #2692: Lint quoted forms in call position

tomdl89:
Fixes #2692

Replaces #2693 

@borkdude I decided to go with the "DO repeat yourself" approach, which 1, isn't that much repetition anyway and 2, keeps things much simpler, without any sentinel `false`. I ran all the relevant tests with the `(prn :expr expr)` form uncommented in `analyze-expression**` and now nothing is analyzed twice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,11 @@ For a list of breaking changes, check [here](#breaking-changes).
 <!-- - [ ] update lein-clj-kondo -->
 <!-- - [ ] update carve -->
 
+
 ## Unreleased
 
 - [#2692](https://github.com/clj-kondo/clj-kondo/issues/2692): Lint quoted forms which are not functions ([@tomdl89](https://github.com/tomdl89))
+- [#2691](https://github.com/clj-kondo/clj-kondo/issues/2691): Fix: `:refer-clojure :exclude` now properly ignores elements with `#_` (e.g. `#_:clj-kondo/ignore comp2`) [@jramosg](https://github.com/jramosg).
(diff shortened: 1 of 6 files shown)


test-regression/clj_kondo/clj_kondo/findings.edn
```diff
@@ -16132,8 +16132,8 @@
   :type :not-a-function,
   :level :error,
   :filename "cljs/vendor/clojure/data/json.clj",
-  :col 7,
-  :end-col 17,
+  :col 8,
```
borkdude, line 16135: What's the explanation behind these changes?

tomdl89: I updated the existing calls to `reg-not-a-function!` to supply the function as the expr, rather than the whole expression. I think it makes more sense to highlight the thing that is "not a function" rather than the whole call. Happy to revert if you disagree.

borkdude: Makes sense.

PR #2720, opened 2026-01-08
Fix arities for comp so type mismatches are found

### LINKED ISSUE #2719: Args for comp are not properly type checked
I think it's just a mistake, but args for comp are not currently properly type checked. So this does not return a type mismatch error:

```clojure
(comp 7 8)
```

PR incoming to fix.

tomdl89:
Fixes #2719 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+- [#2719](https://github.com/clj-kondo/clj-kondo/issues/2719): Fix: `comp` args are now properly type checked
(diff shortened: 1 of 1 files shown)


tomdl89:
Technically, `comp` with one arg returns the arg itself, so `:ifn` as the return type isn't always strictly correct. So this code is legal:
```clojure
(+ (comp 1) 2)
```
even though kondo would now flag it as a type mismatch. Personally I think this is so hideous that it deserves a warning 😂 . Safe to assume that anyone doing this did it by mistake, or is writing intentionally unreadable code, in which case they probably don't care what a linter says. So, happy to leave as is, but can add the correct `1` arity if you prefer.

borkdude:
> So this code is legal:

> (+ (comp 1) 2)

I don't think this _should_ be legal and this is why we have linters. Thanks :)

PR #2730, opened 2026-01-12
Lint map,vector,set calls when bound, not just literal

### LINKED ISSUE #2729: Lint when calling maps, vectors & sets as fns when bound
Currently, literal maps, vectors & sets are linted for invalid arity:

```clojure
({:a 1} 1 2 3)
;; map is called with 3 args but expects 1 or 2
```

but if they are bound in a let binding, they are not:

```clojure
(let [foo {:a 1}]
  (foo 1 2 3))
```

but I think they should be. PR incoming :)

tomdl89:
Fixes #2729 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


borkdude:
Thanks!

PR #2740, opened 2026-01-15
Extend :equals-expected-position linter to also warn for not=

### LINKED ISSUE #2739: Feature: not= should trigger 'Write expected value first' warning like =

## Is your feature request related to a problem? Please describe.

The `not=` operator does not trigger the "Write expected value first" warning, while `=` does.

**Example:**

```clojure
;; This triggers a warning (as expected):
(defn foo [x] (if (= x 1) :one :other))
;; <stdin>:1:24: warning: Write expected value first

;; This does NOT trigger a warning (but should):
(defn foo [x] (if (not= x 1) :other :one))
;; linting took 58ms, errors: 0, warnings: 0
```

## Describe the solution you'd like

`not=` should trigger the same "Write expected value first" warning as `=` when the variable appears before the literal/constant value. The expected code would be:

```clojure
(defn foo [x] (if (not= 1 x) :other :one))
```

## Describe alternatives you've considered

- Leaving it as-is, but this creates inconsistency in the linting rules

## Additional context

This appears to be a gap in the linter's rule implementation. The same principle that applies to `=` (constants/literals should come first for readability and to avoid accidental assignment-like patterns from other languages) should also apply to `not=`.

I will submit a PR

jramosg:
This update enhances the `:equals-expected-position` linter to also warn when using `not=`. This aims to improve code quality by ensuring consistent equality checks across different forms.

Fixes #2739 

### DIFF
--- a/test-regression/clj_kondo/nextjournal/ductile-findings.edn
+++ b/test-regression/clj_kondo/nextjournal/ductile-findings.edn
@@ -1,4 +1,100 @@
-[{:end-row 5,
+[{:filename
+  "test-regression/checkouts/ductile/dev/ductile/insights/billing/as400_billing_exploration.clj",
+  :row 182,
+  :col 1,
+  :type :aliased-referred-var,
+  :message "Var query! is referred but used via alias: clerk-elog",
+  :level :info,
+  :langs ()}
+ {:filename
+  "test-regression/checkouts/ductile/dev/ductile/insights/billing/as400_billing_exploration.clj",
+  :row 192,
+  :col 1,
+  :type :aliased-referred-var,
+  :message "Var query! is referred but used via alias: clerk-elog",
+  :level :info,
+  :langs ()}
+ {:filename
+  "test-regression/checkouts/ductile/dev/ductile/insights/billing/as400_billing_exploration.clj",
+  :row 198,
+  :col 1,
+  :type :aliased-referred-var,
+  :message "Var query! is referred but used via alias: clerk-elog",
+  :level :info,
+  :langs ()}
+ {:filename
+  "test-regression/checkouts/ductile/dev/ductile/insights/billing/as400_billing_exploration.clj",
+  :row 203,
+  :col 19,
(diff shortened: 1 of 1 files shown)


test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -1299,6 +1299,16 @@
...
   :row 93}
+ {:end-row 99,
+  :type :equals-expected-position,
+  :level :warning,
+  :filename
+  "test-regression/checkouts/metabase/test/metabase/driver/sql_jdbc/connection_test.clj",
+  :col 25,
+  :end-col 31,
+  :langs (),
+  :message "Write expected value first",
```
jramosg, line 1419: true positive
```clojure
(is (not= db-nm "null"))
```

test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -1309,6 +1319,26 @@
   :langs (),
   :message "Condition always true",
   :row 132}
+ {:end-row 153,
```
jramosg, line 1431: true positive
```clojure
(is (not= db-hash-1 0))
```

test/clj_kondo/equals_expected_position_test.clj
```diff
@@ -0,0 +1,36 @@
+(ns clj-kondo.equals-expected-position-test 
+  (:require
+   [clj-kondo.test-utils :refer [assert-submaps2 lint!]]
+   [clojure.test :refer [deftest is testing]]))
+
+(deftest equals-expected-order-test
```
borkdude, line 6: Why was this function moved?

jramosg: I thoguht it was cleaner because other linters are tested outside main-test too. I can rollback if you prefer

borkdude: Ah right, this is good

PR #2742, opened 2026-01-15
Add duplicate-exclude linter: catches new duplicated excludes and consolidates the previously separate :duplicate-refer

### LINKED ISSUE #2741: Feature: Add linter for duplicate excludes in ns forms

clj-kondo currently doesn't warn about duplicate symbols in `:exclude` clauses within `ns` forms. This can lead to redundant and potentially confusing namespace declarations, similar to how duplicate refers are already detected.

Add a new linter (e.g., `:duplicate-exclude`) that detects and warns about duplicate symbols in:
1. `(:refer-clojure :exclude [sym1 sym2 sym1])` - duplicate in refer-clojure exclude
2. `(:require [some.ns :exclude [sym1 sym2 sym1]])` - duplicate in require exclude

The warning should be similar to the existing `:duplicate-refer` linter.

- Keep relying on manual code review to catch duplicates

**Additional context**
Currently, clj-kondo successfully detects duplicate refers:
```clojure
;; This produces: "Duplicate refer: split"
(ns foo (:require [clojure.string :refer [split split]]))
```

But does not detect duplicate excludes:
```clojure
;; This produces no duplicate warning, only "Unused excluded var: map"
(ns foo (:refer-clojure :exclude [map map]))
```

The new linter would provide consistency with the existing duplicate detection capabilities and help catch redundant namespace declarations early.

jramosg:
Introduce a new linter `:duplicate-refer` that warns on duplicate entries in `:exclude` vectors within `:refer-clojure` and `:require` clauses. 

Also consolidates the previously separate `:duplicate-refer`  linter into this one.

Fixes Feature: Add linter for duplicate excludes in ns forms Fixes #2741

### DIFF
--- a/test-regression/clj_kondo/clj_kondo/findings.edn
+++ b/test-regression/clj_kondo/clj_kondo/findings.edn
@@ -29542,7 +29542,7 @@
   :message "Unresolved excluded var: set!",
   :row 5}
  {:end-row 6,
-  :type :duplicate-require-option,
+  :type :duplicate-require-item,
   :level :warning,
   :lang :clj,
   :filename "sci/core.cljc",
--- a/test-regression/clj_kondo/metabase/findings.edn
+++ b/test-regression/clj_kondo/metabase/findings.edn
@@ -520,7 +520,7 @@
   :message "Unused excluded var: ->>",
   :row 4}
  {:end-row 5,
-  :type :duplicate-require-option,
+  :type :duplicate-require-item,
   :level :warning,
   :lang :clj,
   :filename
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -38,7 +38,7 @@ configuration. For general configurations options, go [here](config.md).
     - [Docstring leading trailing whitespace](#docstring-leading-trailing-whitespace)
     - [Duplicate map key](#duplicate-map-key)
     - [Duplicate require](#duplicate-require)
-    - [Duplicate require option](#duplicate-require-option)
+    - [Duplicate require item](#duplicate-require-item)
     - [Duplicate set key](#duplicate-set-key)
     - [Duplicate field name](#duplicate-field-name)
     - [Duplicate key args](#duplicate-key-args)
@@ -780,9 +780,9 @@ Explanation by Bozhidar Batsov:
 
 *Example message:* `duplicate require of clojure.string`
 
-### Duplicate require option
+### Duplicate require item
 
-*Keyword:* `:duplicate-require-option`.
+*Keyword:* `:duplicate-require-item`.
 
 *Description:* warns on var that has been specified more than once in a `:refer`, `:refer-macros`, or `:exclude` vector.
 
--- a/src/clj_kondo/impl/analyzer/namespace.clj
+++ b/src/clj_kondo/impl/analyzer/namespace.clj
@@ -99,22 +99,22 @@
   (let [message-prefix (case option-type
                          :refer "Duplicate refer: "
                          :exclude "Duplicate exclude: ")]
-   (when-not (linter-disabled? ctx :duplicate-require-option)
-     (reduce (fn [seen node]
-               (if (utils/ignored? node)
-                 seen
-                 (let [v (:value node)]
-                   (if (contains? seen v)
-                     (do
-                       (findings/reg-finding!
-                        ctx
-                        (node->line (:filename ctx)
-                                    node
-                                    :duplicate-require-option
-                                    (str message-prefix v)))
-                       seen)
-                     (conj seen v)))))
-             #{} nodes))))
+    (when-not (linter-disabled? ctx :duplicate-require-item)
+      (reduce (fn [seen node]
+                (if (utils/ignored? node)
+                  seen
+                  (let [v (:value node)]
+                    (if (contains? seen v)
+                      (do
+                        (findings/reg-finding!
+                         ctx
+                         (node->line (:filename ctx)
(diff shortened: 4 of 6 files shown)


test-regression/clj_kondo/clj_kondo/findings.edn
```diff
@@ -29535,6 +29535,17 @@
   :langs (:clj :cljs),
   :message "Unresolved excluded var: set!",
   :row 5}
+ {:end-row 6,
```
jramosg, line 29544: https://github.com/babashka/sci/blob/9edc11ae15f2f099348a1fbfc7c3e0607a20909b/src/sci/core.cljc#L3C1-L8C1

```clojure
  (:refer-clojure :exclude [with-bindings with-in-str with-out-str
                            binding future pmap alter-var-root
                            intern ns create-ns *1 *2 *3 *e
                            ns-name assert print-dup find-ns all-ns ns-name
                            resolve])
```

test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -427,6 +427,18 @@
   :langs (:clj :cljs),
   :message "Unused excluded var: ->>",
   :row 4}
+ {:end-row 5,
```
jramosg, line 522:
```clojure
(ns metabase.lib.core
  "Currently this is mostly a convenience namespace for REPL and test usage. We'll probably have a slightly different
  version of this for namespace for QB and QP usage in the future -- TBD."
  (:refer-clojure :exclude [filter remove replace and or not = < <= > ->> >= not-empty case count distinct max min
                            + - * / time abs concat replace ref var float])
```

tomdl89:
Just a thought - couldn't this just be part of the `:duplicate-refer` linter? I know it's technically not a refer, but it's similar enough that I doubt anyone would be confused. And I can't imagine a situation where someone would have a different `:level` for each. With a little bit of mind bending, you can kind of see that an exclude is like a "negative" refer anyway..? Just mindful that an opportunity not to grow the already large list of possible linters could be a good thing..?

borkdude:
I second @tomdl89's thoughts. I think all of these can be lumped together (with a better name maybe?)

jramosg:
what do you think about calling both `duplicate-require-option`?

tomdl89:
That makes me think more of something like this: `(ns foo (:require [bar :exclude [baz] :exclude [qux]]))` where there are 2 `:exclude`s. What about `duplicate-require-symbol` or `duplicate-require-var`?

eta: I think we're safe calling it `duplicate-require-***` even though these filters can be used outside of `require` (e.g. `refer`) because you rarely see the latter in reality.

tomdl89:
This has blown my mind:
```clojure
foo> (ns goo (:require [clojure.set :refer [union] :refer [difference]]))
nil
goo> (union #{:a} #{:b})
Syntax error compiling at (*cider-repl repos/clj-kondo:localhost:61759(clj)*:1116:6).
Unable to resolve symbol: union in this context
goo> (difference #{:a} #{:b})
#{:a}
```

I would have expected either it throw a compile error, take the first :refer, or merge both. Taking the last feels unintuitive 🤷 . I feel a new linter coming...

jramosg:
I've joined both linters  `duplicate-require` and `duplicate-refer` into `duplicate-require-option` (name to decide).
Thanks!

About `(ns goo (:require [clojure.set :refer [union] :refer [difference]]))` I didn't know that. We can create `duplicate-require-key` for that

jramosg:
> This has blown my mind:
> 
> ```clojure
> foo> (ns goo (:require [clojure.set :refer [union] :refer [difference]]))
> nil
> goo> (union #{:a} #{:b})
> Syntax error compiling at (*cider-repl repos/clj-kondo:localhost:61759(clj)*:1116:6).
> Unable to resolve symbol: union in this context
> goo> (difference #{:a} #{:b})
> #{:a}
> ```
> 
> I would have expected either it throw a compile error, take the first :refer, or merge both. Taking the last feels unintuitive 🤷 . I feel a new linter coming...

I'm thinking about implementing this, what do you think is better? A new linter (for example `:duplicate-require-key`) or using already existing `unreachable-code`? Since here, in row 1 `union` is unreachable

```clojure
foo> (ns goo (:require [clojure.set :refer [union] :refer [difference]]))
nil
goo> (union #{:a} #{:b})
```

tomdl89:
It doesn't feel like unreachable code to me. It is reached, it just doesn't get referred:
```clojure
hoo> (require ['clojure.string :refer [(do (println "hello!") 'trim)] :refer ['triml]])
hello!
nil
hoo> (trim " bar")
Syntax error compiling at (*cider-repl repos/clj-kondo:localhost:61759(clj)*:1159:6).
Unable to resolve symbol: trim in this context
hoo> (triml " bar")
"bar"
```

note the printing of `hello!`

NoahTheDuke:
`require` is a function so it eagerly evaluates the body (executing `(println "hello")` as shown), and then it [calls](https://github.com/clojure/clojure/blob/a3fa897590f70207eea3573759739810f2b6ab6c/src/clj/clojure/core.clj#L5988) `(apply hash opts)` which creates a map with the last defined kv pair for each unique key.

tomdl89:
And furthermore, in the `ns` macro, the first refer in my example is silently ignored, and the println doesn't happen. So it is effectively unreachable there. Just felt it wasn't strictly accurate because of the `require` function, so maybe worth choosing a better name.

PR #2748, opened 2026-01-16
Fix #2746: Recognize primitive array class syntax in type checking

### LINKED ISSUE #2746: Regression in v2026.01.12: `error: Expected: class, received: function` when using `primitive/N` array class syntax

**version**

v2026.01.12

**macro usage**

no

**platform**

several

**problem**

When using the [array class syntax ](https://clojure.org/reference/java_interop#_class_access) supported since Clojure 1.12 in combination with primitives the symbol gets misinterpreted as a function.

**repro**

 This code

```clojure
(instance? byte/1 (byte-array 0))
```

yields

```
error: Expected: class, received: function.
```

I think even syntactically `byte/1` (and the other primitive array classes) can't ever be a function, because if `byte` is interpreted as a namespace alias, `1` would be symbol starting with a digit.

**expected behavior**

Symbols of the form `primitive/N` should be interpreted as class literals not as functions therefore the above warning shouldn't occur.

jramosg:
Primitive array class literals (byte/1, int/2, etc.) introduced in Clojure 1.12 are now correctly recognized as :class types rather than :fn types when used with functions like instance?, cast, and make-array. The fix adds a helper function primitive-array-class? that detects symbols matching the pattern primitive/N where primitive is one of: boolean, byte, short, int, long, float, double, or char, and N is a digit representing the array dimension.
Fixes #2746

### DIFF
--- a/src/clj_kondo/impl/types.clj
+++ b/src/clj_kondo/impl/types.clj
@@ -386,7 +386,7 @@
         ;; Check if this is an array class literal
         ;; (Clojure 1.12+) would emit those as :class tags (type String/1) => java.lang.Class
         (and (identical? :clj (:lang ctx))
-             (array-class-literal? (sexpr expr)))
+             (array-class-literal? (:value expr)))
         {:tag :class}
 
         :else {:usage (or tag


src/clj_kondo/impl/types.clj
```diff
@@ -368,18 +368,33 @@
                     (when-let [v (:val t)]
                       {:tag (get v nm)}))))))))))
 
+(defn- primitive-array-class? [sym]
+  (when (symbol? sym)
+    (when-let [ns-part (namespace sym)]
+      (when (contains? #{"boolean" "byte" "short" "int" "long" "float" "double"
```
borkdude: This is too limiting since any class may be referenced this way, e.g. `String/1`

src/clj_kondo/impl/types.clj
```diff
@@ -387,10 +383,10 @@
       (cond
         tag {:tag tag}
 
-        ;; Check if this is a primitive array class literal
-        ;; (Clojure 1.12+) would emit those as :class tags (type byte/1) => java.lang.Class
+        ;; Check if this is an array class literal
+        ;; (Clojure 1.12+) would emit those as :class tags (type String/1) => java.lang.Class
         (and (identical? :clj (:lang ctx))
-             (primitive-array-class? (sexpr expr)))
+             (array-class-literal? (sexpr expr)))
```
borkdude: This sexpr call went unnoticed last time. Please avoid sexpr when you can, it can crash.

Change this to `(:v expr)` to get the symbol. Or use `symbol-from-token` from utils.

jramosg: sorry, I forgot about it!

PR #2751, opened 2026-01-16
Fix gensym bindings in nested syntax quotes

### LINKED ISSUE #2747: Regression in v2026.01.12: Bindings of unquoted gensyms not detected anymore

**version**

v2026.01.12

**macro usage**

Not in macro usage, but in macro definition.

**platform**

Several

**problem**

The bindings of unquoted [gensyms](https://clojure.org/guides/weird_characters#gensym), as they can appear in nested syntax quotes, are not detected anymore.

**repro**

Linting

```clojure
(defmacro def-some-macro
  []
  `(defmacro ~'some-macro
     [x#]
     `(list ~x#)))
```

Yields: `error: Unresolved symbol: x#` 

**expected behavior**

No error should be generated if there is a binding.

jramosg:
Gensym bindings are now correctly recognized in nested syntax quotes, preventing unresolved symbol errors. This change improves linting accuracy for macros that utilize gensym in their definitions.

Fixes #2747 

### DIFF
--- a/src/clj_kondo/impl/analyzer/usages.clj
+++ b/src/clj_kondo/impl/analyzer/usages.clj
@@ -139,13 +139,12 @@
          unquote-tag? (one-of t [:unquote :unquote-splicing])
          new-syntax-quote-level (cond syntax-quote-tag? (inc syntax-quote-level)
                                       :else syntax-quote-level)
-         new-syntax-quote-level-pos? (pos? new-syntax-quote-level)
          syntax-quote? (or syntax-quote? syntax-quote-tag?)
          ctx (assoc ctx :syntax-quote-level new-syntax-quote-level)
          ctx (if syntax-quote-tag?
                (update ctx :callstack #(cons [:syntax-quote] %))
                ctx)]
-     (if (and new-syntax-quote-level-pos? unquote-tag?)
+     (if (and (pos? syntax-quote-level) unquote-tag?)
        (common/analyze-expression** ctx expr)
        (if quote?
          (do
@@ -168,7 +167,8 @@
                      symbol-val (if simple?
                                   (namespace/normalize-sym-name ctx symbol-val)
                                   symbol-val)
-                     expr-meta (meta expr)]
+                     expr-meta (meta expr)
+                     new-syntax-quote-level-pos? (pos? new-syntax-quote-level)]
                  (if-let [b (when (and simple? (not new-syntax-quote-level-pos?))
                               (or (get (:bindings ctx) symbol-val)
                                   (get (:bindings ctx)


jramosg:
I think it is fixed but I found this possible bug or behaviour that I don't know if it is already known

This test passed before the fix too.

```clojure
(deftest issue-2747-test
  (testing "Gensym bindings in nested syntax quotes should not cause unresolved symbol errors"
    (is (empty? (lint! "(defmacro def-some-macro [] `(defmacro ~'some-macro [x#] `(list ~x#)))")))
    (is (empty? (lint! "(defmacro outer [] `(defmacro ~'inner [a# b#] `(+ ~a# ~b#)))")))))
```

However this was throwing an error before the fix. Now it is not
```bash
clojure -M:clj-kondo/dev --lint - <<< "(defmacro def-some-macro [] \`(defmacro ~'some-macro [x#] \`(list ~x#)))"
```


```

**Comment by borkdude** [comment]
@jramosg The issue is probably that `:unresolved-symbol` is off in the base config for the tests. Maybe it's better to change this but then maybe we would have to change lots of tests. But passing the explicit config would make it work I think.

**Comment by jramosg** [comment]
thanks!!!

File: src/clj_kondo/impl/analyzer/usages.clj [hunk 1]
```diff
@@ -167,8 +167,9 @@
                      symbol-val (if simple?
                                   (namespace/normalize-sym-name ctx symbol-val)
                                   symbol-val)
-                     expr-meta (meta expr)]
-                 (if-let [b (when (and simple? (not syntax-quote?))
+                     expr-meta (meta expr)
+                     in-syntax-quote? (or syntax-quote? (pos? (or (:syntax-quote-level ctx) 0)))]
```
borkdude: - We already check (pos? syntax-quote-level) above, we can bind that to a variable above and re-use it here
- Probably we only have to check `(pos? syntax-quote-level)` since it will be positive when syntax-quote? is also true

src/clj_kondo/impl/analyzer/usages.clj
```diff
@@ -130,6 +130,7 @@
    (let [ns (:ns ctx)
          dependencies (:dependencies ctx)
          syntax-quote-level (or (:syntax-quote-level ctx) 0)
+         sintax-quote-level-pos? (pos? syntax-quote-level)
```
borkdude: let's call it syntax instead of sintax? :)

borkdude: btw, I think we need to use new-syntax-level here

borkdude: which already exists

borkdude:
Merged via other PR with some minor additional changes. Thanks!

PR #2760, opened 2026-01-21
Add :duplicate-require-key linter

### LINKED ISSUE #2744: Feature Request: Linter for Duplicate Require Keys
## Description

Detect when the same key (like `:refer`, `:exclude`, etc.) appears multiple times in a single require clause.

## Context

As discovered by [@tomdl89](https://github.com/tomdl89) in [this comment](https://github.com/clj-kondo/clj-kondo/pull/2742#issuecomment-3755076210), Clojure allows duplicate keys in require clauses but behaves unexpectedly - it takes the **last** occurrence:

```clojure
(ns goo (:require [clojure.set :refer [union] :refer [difference]]))

(union #{:a} #{:b})
;; Syntax error compiling - Unable to resolve symbol: union in this context

(difference #{:a} #{:b})
;; => #{:a}
```

This behavior is unintuitive. Most developers would expect either:
- A compile error
- The first `:refer` to be used
- Both `:refer` vectors to be merged

Instead, the last `:refer` silently overrides the first, making `union` unreachable.

## Proposed Solution

- Create a new linter (e.g., `:duplicate-require-key`) that warns when the same key appears multiple times in a require clause.
- Use the existing `:unreachable-code` linter since the symbols in the first `:refer` become unreachable. 

Example warning:
```
Duplicate :refer key in require clause. Only the last occurrence will be used.
```

## Attribution

Suggested by [@tomdl89](https://github.com/tomdl89) in https://github.com/clj-kondo/clj-kondo/pull/2742#issuecomment-3755076210

---

I can do a PR myself (when we decide what proposed solution is better)

jramosg:
Detects when the same key (like :refer, :as, :exclude, etc.) appears multiple times in a single require clause. Clojure uses only the last occurrence, making earlier values unreachable.

Closes #2744

- Add lint-duplicate-require-keys! function to detect duplicate keys
- Add :duplicate-require-key linter to default config
- Add comprehensive tests including inline ignore support
- Update main_test.clj to expect duplicate-require-key warning
- Update linters.md documentation
- Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## 2026.01.19
 
+- [#2744](https://github.com/clj-kondo/clj-kondo/issues/2744): NEW linter: `:duplicate-require-key` which warns when the same key (like `:refer`, `:as`, `:exclude`, etc.) appears multiple times in a single require clause. ([@jramosg](https://github.com/jramosg))
 - [#2735](https://github.com/clj-kondo/clj-kondo/issues/2735): NEW linter: `:duplicate-refer` which warns on duplicate entries in `:refer` of `:require`. ([@jramosg](https://github.com/jramosg))
 - [#2734](https://github.com/clj-kondo/clj-kondo/issues/2734): NEW linter: `:aliased-referred-var`, which warns when a var is both referred and accessed via an alias in the same namespace. ([@jramosg](https://github.com/jramosg))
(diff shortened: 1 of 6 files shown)


borkdude:
Is this a follow-up to #2742 or supersedes it?

jramosg:
These are two different validations that detect different kinds of duplication.

---

### 1) Duplicate symbols in `:refer` / `:exclude`

This warning checks for duplicate symbols inside the refer/exclude vector.

It means the same var name is listed more than once

```clojure
(ns my.app.core
  (:require
    [clojure.string :refer [split split]]
    [my.lib :exclude [foo foo]]))

```

### 2) Duplicate keys inside a single `:require` libspec

This warning detects duplicate keys inside the same libspec vector, e.g. multiple `:refer`, `:as`, `:rename`, etc.

Only the last occurrence of a key is used, so earlier ones are silently ignored.

---

```clojure
(ns my.app.core
  (:require
    [clojure.string
     :refer [trim]
     :refer [trimr]]))

```

PR #2761, opened 2026-01-21
Add new type ipending and type checking for realized?

jramosg:
- Add new type `:ipending` and type checking for `realized?`
- Add return type for `repeat` with two args

---


src/clj_kondo/impl/types.clj
```diff
@@ -240,7 +248,6 @@
     (float) :float
     (Double java.lang.Double) :nilable/double
     (Float java.lang.Float) :nilable/float
-    (ratio?) :ratio
```
NoahTheDuke: why this change?

borkdude: I guess this is removed since `ratio?` can't really be a tag and this was a typo. Can you confirm @jramosg ?

borkdude:
Distinguishing between seq and lazy-seq: does it have any benefits for the type system, I wonder?
Also: many things marked as lazy-seq now aren't really lazy seqs. E.g.:

```
user=> (instance? clojure.lang.LazySeq (range 1 10))
false
```

So I propose to roll the lazy-seq change back and make it just seq again, which makes the diff for this PR much smaller.

jramosg:
Range with 1 and 2 args is typed as seq, not lazy-seq [https://github.com/jramosg/clj-kondo/blob/feature/lazy-seq-ipending-types/src/clj_kondo/impl/types/clojure/core.clj#L637](https://github.com/jramosg/clj-kondo/blob/feature/lazy-seq-ipending-types/src/clj_kondo/impl/types/clojure/core.clj#L637)

I introduced lazy-seq because some functions, for example, realized?, accept lazy-seq but not seq. ipending could be a lazy-seq but not a seq

borkdude:
I think we're entering overly specific territory here.

```
user=> (realized? (range 1))
Execution error (ClassCastException) at user/eval156 (REPL:1).
class clojure.lang.LongRange cannot be cast to class clojure.lang.IPending (clojure.lang.LongRange and clojure.lang.IPending are in unnamed module of loader 'app')
user=> (realized? (range))
true
```

Passing a seq to `realized?` is mostly useless anyway.

This is an interesting thread on Clojurians Slack about it:

https://clojurians.slack.com/archives/C03S1KBA2/p1764878187471899

jramosg:
I deleted lazy-seq type checking

src/clj_kondo/impl/types.clj
```diff
@@ -160,6 +162,7 @@
    :ratio "ratio"
    :seqable "seqable collection"
    :seq "seq"
+   :ipending "pending (unrealized lazy seq, delay, future, or promise)"
```
borkdude: ipending doesn't mean unrealized: once realized it doesn't turn into a non-ipending

PR #2769, opened 2026-02-09
Add new linter for redundant declare usage

### LINKED ISSUE #2768: declare after var impl should warn
Example:

``` clojure
(defn foo [])

(declare foo) ;; warning
```

jramosg:
Introduce a new linter `:redundant-declare` that warns when `declare` is used after a var is already defined in the same namespace. Update documentation and configuration to include this new linter, and add tests to ensure correct functionality and behavior.

Fixes #2768

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -1537,11 +1537,11 @@
                             (current-namespace-var-name ctx var-name-node var-sym)))
                         var-name-nodes)
         vars-ns-path [(:base-lang ctx) (:lang ctx) ns-name :vars]
-        vars (get-in @(:namespaces ctx) vars-ns-path {})]
+        vars (get-in @(:namespaces ctx) vars-ns-path)]
     (doseq [var-name var-names
             :let [var-name-meta (meta var-name)]]
       (when-not (linter-disabled? ctx :redundant-declare)
-        (let [existing-var (vars var-name)]
+        (let [existing-var (get vars var-name)]
           (when (and existing-var
                      (not (utils/ignored? existing-var)))
             (findings/reg-finding!


test-regression/clj_kondo/clj_kondo/findings.edn
```diff
@@ -1469,6 +1469,17 @@
   :langs (:clj :cljs),
   :message "unused binding op",
   :row 1702}
+ {:end-row 1793,
```
jramosg, line 1472: true positive:
declare on line 1783 and `def` in line 1503 https://github.com/clojure/clojurescript/blob/091df76c65478f9150e83ebbb13b4cf1097a2b2a/src/main/clojure/cljs/analyzer.cljc#L1503

test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -613,6 +613,16 @@
...
+ {:end-row 528,
+  :type :redundant-declare,
+  :level :warning,
+  :filename
+  "test-regression/checkouts/metabase/src/metabase/models/serialization.clj",
+  :col 1,
+  :end-col 24,
+  :langs (),
+  :message "redundant declare of extract-query",
+  :row 528}
```
jramosg, line 625: true positive, definied in line 444

Image

test-regression/clj_kondo/nextjournal/clerk-findings.edn
```diff
@@ -58,6 +58,16 @@
   :langs (:cljs),
   :message "Var ordered-map is referred but used via alias: omap",
   :row 105}
+ {:end-row 219,
```
jramosg, line 61: true positive, definied in line 148

Image

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1536,19 +1536,29 @@
...
+      (when-not (linter-disabled? ctx :redundant-declare)
+        (let [var-path [(:base-lang ctx) (:lang ctx) ns-name :vars var-name]
+              existing-var (get-in @(:namespaces ctx) var-path)]
+          (when (and existing-var
+                     (not (:declared existing-var))
+                     (not (utils/ignored? existing-var)))
+            (findings/reg-finding!
+             ctx
+             (node->line (:filename ctx) expr :redundant-declare
+                         (str "redundant declare of " var-name))))))
```
borkdude: Change the message to:

Redundant declare: <var-name>

test/clj_kondo/redundant_declare_test.clj
```diff
@@ -0,0 +1,43 @@
...
+    (assert-submaps2
+     '({:file "<stdin>", :row 1, :col 13, :level :warning,
+        :message "redundant declare of foo"})
+     (lint! "(def foo 1) (declare foo)")))
+
+  (testing "no warning for normal forward declare pattern"
+    (is (empty? (lint! "(declare foo) (defn foo [])"))))
+
+  (testing "no warning for idempotent declare"
+    (is (empty? (lint! "(declare foo) (declare foo)"))))
```
borkdude: Should we also regard the second one as redundant, since it's already declared?

test/clj_kondo/redundant_declare_test.clj
```diff
@@ -0,0 +1,43 @@
...
+
+  (testing "no warning when linter is disabled"
+    (is (empty? (lint! "(defn foo []) (declare foo)"
+                       {:linters {:redundant-declare {:level :off}}}))))
+
+  (testing "no warning when ignored inline"
+    (is (empty? (lint! "(defn foo []) ^{:clj-kondo/ignore [:redundant-declare]} (declare foo)"))))
```
borkdude: Can you also test the `#_:clj-kondo/ignore` syntax

test/clj_kondo/redundant_declare_test.clj
jramosg: yes okey, I agree

test/clj_kondo/redundant_declare_test.clj
jramosg: done

src/clj_kondo/impl/analyzer.clj
```diff
@@ -1536,19 +1536,28 @@
...
-                                   :name-col (:col var-name-meta)
-                                   :name-end-row (:end-row var-name-meta)
-                                   :name-end-col (:end-col var-name-meta)
-                                   :declared true
-                                   :defined-by defined-by
-                                   :defined-by->lint-as defined-by->lint-as))))))
+    (doseq [var-name var-names
+            :let [var-name-meta (meta var-name)]]
+      (when-not (linter-disabled? ctx :redundant-declare)
+        (let [var-path [(:base-lang ctx) (:lang ctx) ns-name :vars var-name]
```
borkdude: I think we can get the right namespace map before the doseq prevent extra map lookups
so lift the ns-path (ending with `:vars` above the doseq and then te var-path becomes ns-path conjed with var-name?

jramosg: sure, changed!

tomdl89:
Does this catch redundant declaration using `def` rather than `declare`?
```clojure
(def foo 42)

foo
;;=> 42

(def foo) ;; <-- redundant declaration, using def rather than declare

foo
;;=> 42
```

From a quick glance, I can't see that it does, so maybe worth adding for completeness?

borkdude:
It's already caught as a re-initialized var, different linter. Also another thing clj-kondo reports is that `foo` doesn't have an init value.

tomdl89:
Ah cool. Sorry for the noise!

PR #2777, opened 2026-02-27
Set from-var in var-usages analysis for defmethod bodies

### LINKED ISSUE #2776: var-usages inside defmethod bodies have null from-var
Hey, I found this quirky behavior while using the analyzer to build an internal tool and I had Claude give it a shot at fixing it. Let me know if you think this is useful, otherwise happy to close both the issue and the PR. Here's Claude's description:

## Problem

When analyzing `defmethod` bodies, `var-usages` in the analysis output have `from-var` set to `null`. Calls inside `defn` bodies correctly get `from-var` set to the enclosing function name, but `defmethod` doesn't establish this context.

## Reproduction

```clojure
(ns test.defmethod
  (:require [clojure.string :as str]))

(defn helper [x]
  (str/upper-case x))

(defmulti greet :type)

(defmethod greet :formal
  [{:keys [name]}]
  (str/join " " ["Dear" (helper name)]))

(defmethod greet :casual
  [{:keys [name]}]
  (str/lower-case (str "hey " name)))
```

```bash
clj-kondo --lint /tmp/test_defmethod.clj \
  --config '{:output {:format :json} :analysis {:arglists true}}' \
  2>/dev/null | jq '.analysis["var-usages"][] | {row, name, from, "from-var", to}'
```

### Expected

Calls inside defmethod bodies should have `from-var` set to the multimethod name:

```json
{"row":11,"name":"join","from":"test.defmethod","from-var":"greet","to":"clojure.string"}
{"row":11,"name":"helper","from":"test.defmethod","from-var":"greet","to":"test.defmethod"}
```

### Actual

`from-var` is `null` for all calls inside defmethod bodies:

```json
{"row":11,"name":"join","from":"test.defmethod","from-var":null,"to":"clojure.string"}
{"row":11,"name":"helper","from":"test.defmethod","from-var":null,"to":"test.defmethod"}
```

Meanwhile, calls inside `defn` bodies correctly have `from-var` set:

```json
{"row":5,"name":"upper-case","from":"test.defmethod","from-var":"helper","to":"clojure.string"}
```

## Root Cause

`analyze-defmethod` in `analyzer.clj` calls `analyze-fn` without setting `:in-def` on the context. In contrast, `analyze-def` (for `defn`) does `(assoc ctx :in-def var-name)` before analyzing the function body. Since `:in-def` is what gets mapped to `:from-var` in the analysis output, defmethod bodies produce `null`.

## Use Case

This matters for tools that build call graphs from clj-kondo analysis. Without `from-var`, calls inside defmethod bodies can't be attributed to any function, leaving gaps in dependency graphs and blast-radius analysis.

## PR

Fix submitted in #2777.

volrath:
Fixes #2776

## Summary

- `analyze-defmethod` now sets `:in-def` to the multimethod name on the context before calling `analyze-fn`, so `from-var` in `var-usages` analysis output is correctly populated for calls inside defmethod bodies
- Also improves `:inline-def` linter coverage: deeply nested `def` forms inside defmethod bodies are now caught (previously only direct children were detected via a callstack check)
- Updated metabase regression expected findings for 1 new true-positive inline-def

## Test plan

- [x] Added `defmethod-from-var-test` in `analysis_test.clj`
- [x] Full test suite passes (604 tests, 4117 assertions, 0 failures)
- [x] Regression tests pass
- [x] Manual smoke test confirms `from-var` is set for all calls inside defmethod bodies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,11 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#2762](https://github.com/clj-kondo/clj-kondo/issues/2762): Fix false positive: `throw` with string in CLJS no longer warns about type mismatch ([@jramosg](https://github.com/jramosg))
 - Add type support for `pmap` ([@jramosg](https://github.com/jramosg))
 - Type system: Add type support for future-related functions (`future`, `future-call`, `future-done?`, `future-cancel`, `future-cancelled?`) ([@jramosg](https://github.com/jramosg))
+- [#2770](https://github.com/clj-kondo/clj-kondo/issues/2770): Fix: linter-specific ignores now correctly respect the specified linters instead of suppressing all linters for `:unused-excluded-var` and `:unresolved-excluded-var`
 - [#2773](https://github.com/clj-kondo/clj-kondo/issues/2773): Align executable path for images to be `/bin/clj-kondo` ([@harryzcy](https://github.com/harryzcy))
+- [#2779](https://github.com/clj-kondo/clj-kondo/issues/2779), [#2780](https://github.com/clj-kondo/clj-kondo/issues/2780), [#2781](https://github.com/clj-kondo/clj-kondo/issues/2781), [#2782](https://github.com/clj-kondo/clj-kondo/issues/2782), [#2783](https://github.com/clj-kondo/clj-kondo/issues/2783), [#2785](https://github.com/clj-kondo/clj-kondo/issues/2785), [#2786](https://github.com/clj-kondo/clj-kondo/issues/2786): Performance optimizations.
+- Performance: cache hook-fn lookups to avoid repeated SCI evaluation
+
 
 ## 2026.01.19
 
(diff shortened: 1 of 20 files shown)


borkdude:
thanks, I'll take a look soon!

test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -844,6 +844,16 @@
   :langs (),
   :message "Redundant let binding: query",
   :row 211}
+ {:end-row 674,
+  :type :inline-def,
+  :level :warning,
+  :filename
+  "test-regression/checkouts/metabase/src/metabase/query_processor/streaming/xlsx.clj",
```
borkdude, line 851: Just a note: this inline def comes from here:

https://github.com/metabase/metabase/blob/aa0cdb546d7c9e4ef5c52ad23c656272b7599e23/src/metabase/query_processor/streaming/xlsx.clj#L674

PR #2790, opened 2026-03-17
format-output optimization

jramosg:
Replaced 9 chained str/replace calls with a single str/replace using a compiled regex + case dispatch. The pattern string is
scanned once instead of 9 times per finding.


borkdude:
I don't think `format-output` is on a hot path? How much time is spent there compared to other parts? (My expectation: only a tiny fraction). "Improves performance" doesn't say much unless you provide numbers. I merged the previous PR since it was only a small change, but I don't think that this change will help maintainability of readability of the code.

PR #2793, opened 2026-03-19
Extract var definitions from AOT-only jars

### LINKED ISSUE #2792: Support linting of AOT-only libraries (jars with only .class files)

**Problem**

Some libraries ship jars containing only AOT-compiled `.class` files and no `.clj` source files. When a project uses `:use` or `:require :refer :all` (or `:require :refer [specific-var]`) on a namespace from such a jar, clj-kondo has no way of knowing what vars are defined in that namespace. This produces false-positive `unresolved-symbol` and `unresolved-var` warnings.

**Proposed solution**

Extract Clojure var definitions from AOT-compiled `__init.class` bytecode during jar processing and write them to the clj-kondo cache. The `__init.class` files contain `RT.var` calls with var names, arglists, and metadata (`:private`, `:macro`, etc.) that can be parsed using ASM (already a dependency).

This should:
- Be always-on with no config flag
- Provide equivalent information to source analysis: var names, arities, private/public, macro flags
- Give precedence to source analysis when both `.clj` and `.class` exist for the same namespace
- Work within the existing `--dependencies` / cache workflow — AOT extraction happens during `sources-from-jar`, and the skip-mark mechanism prevents redundant work on subsequent runs

**Alternatives considered**

- *Routing extracted data through the analyzer pipeline*: The analyzer exists to transform source text into structured var data. For `.class` files the structured data is already in the bytecode — there is no source to parse. Routing through the analyzer would require synthesizing fake source or adding a parallel code path through `analyze-input`, adding complexity for no benefit.
- *Config-based opt-in*: Rejected because this should work transparently. The false positives affect any project depending on an AOT-only library.
- *Stub files / manual configuration*: Users can work around this today with `:config {:linters {:unresolved-symbol {:exclude [...]}}}`, but this loses arity checking and requires per-library maintenance.

hugoduncan:

## Summary

Fixes #2792

When a jar contains only AOT-compiled `.class` files and no `.clj` source, clj-kondo now extracts Clojure var definitions from `__init.class` bytecode using ASM. This eliminates false-positive `unresolved-symbol` and `unresolved-var` warnings for `:use`, `:require :refer :all`, and `:require :refer [specific]`.

Extracted metadata includes var names, arities, varargs, private and macro flags. Source analysis takes precedence when both `.clj` and `.class` exist for the same namespace.

## Test plan

- [x] Unit: bytecode extraction correctness (arities, private, macro, varargs, plain defs)
- [x] Integration: `:refer :all` resolves without false positives
- [x] Integration: `:refer [specific]` resolves without false positives
- [x] Arity checking works against AOT-only functions
- [x] Source precedence: jar with both `.clj` and `.class` uses source-analyzed data
- [x] Full test suite passes (609 tests, 0 failures)
- [x] Native binary compiles and works


borkdude:
The normal flow of analysis in clj-kondo:

1. process-jar returns source entries (data only, no side effects)
2. Sources get analyzed, building up idacs (intermediate analysis data)
3. sync-cache writes everything to cache at the end

This PR mixes analysis with I/O in process-jar. Is there anything we could do about that?

hugoduncan:
I was using the precendent of analysis writing directly during `process-jar` to avoid have to explicitly code the .clj over .class preference.
If you prefer, I can have process-jar return an idac and add code for the  explicit .clj over .class.

borkdude:
@hugoduncan I messed a little bit with this PR and made the AOT analysis return data so the writing is done as part of the same pipeline as the other analysis. Can you review?

PR #2796, opened 2026-03-20
Fix cache lock contention with concurrent clj-kondo processes

### LINKED ISSUE #2795: Cache lock contention with concurrent clj-kondo processes (editor + pre-commit hook) causes unhandled exception

**version**

v2026.01.19

**macro usage**

N/A

**platform**

JVM (also reproduced with native binary)

**editor**

Calva (VS Code): clj-kondo runs in the background via the editor integration

**problem**

When committing with a pre-commit hook that runs `clj-kondo --lint`, the lint fails with an "Unexpected error" if the editor's clj-kondo integration is linting the same files concurrently. The two processes compete for the `.clj-kondo` cache file lock; after 6 retries (~1.5s total backoff), clj-kondo throws a hard exception that propagates as an unhandled error rather than a recoverable warning.

```
Unexpected error. Please report an issue.
java.lang.Exception: Clj-kondo cache is locked by other thread or process.
    at clj_kondo.impl.cache$sync_cache.invokeStatic(cache.clj:202)
    at clj_kondo.core$run_BANG_.invokeStatic(core.clj:242)
    at clj_kondo.main$main.invokeStatic(main.clj:150)
    ...
```

**repro**

1. Open a Clojure project in an editor that runs clj-kondo on file changes (e.g. Calva, clojure-lsp)
2. Add a pre-commit hook that runs `clj-kondo --lint` on staged files
3. Save a file (triggering background lint) and immediately `git commit`
4. The pre-commit hook fails with the cache lock error above

The same scenario occurs with two pre-commit hooks both invoking clj-kondo (e.g. a local `clj-kondo --lint` hook alongside a Docker-based hook that mounts the project directory).

**config**

N/A. bug occurs with default configuration.

**expected behavior**

When the cache is locked by another clj-kondo process, linting should continue without the cache (degraded mode) and emit a warning to stderr rather than failing the entire lint run with an unhandled exception. Cache contention is a normal operational condition when an editor integration and a pre-commit hook run concurrently; it should not prevent commits.

jramosg:
When an editor integration (e.g. Calva) lints files in the background concurrently with a pre-commit hook running clj-kondo, both processes compete for the .clj-kondo cache file lock. After 6 retries (~1.5s), clj-kondo would throw an unhandled exception causing the commit to fail.

This change makes cache contention a recoverable condition: when the lock cannot be obtained, clj-kondo now warns to stderr and continues linting without cache rather than aborting with an exception. This allows concurrent invocations (editor + hook, or multiple hooks) to coexist gracefully with degraded but functional behavior.

Fixes #2795

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 - [#2773](https://github.com/clj-kondo/clj-kondo/issues/2773): Align executable path for images to be `/bin/clj-kondo` ([@harryzcy](https://github.com/harryzcy))
 - [#2779](https://github.com/clj-kondo/clj-kondo/issues/2779), [#2780](https://github.com/clj-kondo/clj-kondo/issues/2780), [#2781](https://github.com/clj-kondo/clj-kondo/issues/2781), [#2782](https://github.com/clj-kondo/clj-kondo/issues/2782), [#2783](https://github.com/clj-kondo/clj-kondo/issues/2783), [#2785](https://github.com/clj-kondo/clj-kondo/issues/2785), [#2786](https://github.com/clj-kondo/clj-kondo/issues/2786): Performance optimizations.
 - Performance: cache hook-fn lookups to avoid repeated SCI evaluation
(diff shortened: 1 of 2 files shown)


borkdude:
Another option would be to increase the retries? It should be pretty uncommon that two clj-kondo processes hold the lock for very long.

jramosg:
For some reason I don't know in my VS code  it was locked permanently (until I restarted) so pre-commit hook was failing. Increasing retries wouldn't fix it in this case

borkdude:
I think I'd rather crash than wipe the bug under the carpet and I'd rathe learn why your vscode was constantly locking.

Issue #2778, opened 2026-03-01
Improve the performance of certain code paths


alexander-yakushev:
This is part of an effort to improve Clojure LSP which transitively heavily relies on clj-kondo. I'll use this issue as an umbrella to link all the PRs to it.

ericdallo:
FYI I bumped clojure-lsp after all those PRs were merged!

ericdallo:
I did an extensive benchmark with the clj-kondo bump including those fixes and the results are promissing for medium/big projects!

I asked LLM to run `clojure-lsp diagnostics` both cold and warm, where cold means no lsp or clj-kondo cache, also I run in small, medium/big and huge projects. Artemisia and stormshield are similar in size but codebase quite different.

Again amazing work @alexander-yakushev ! 🚀 

Here's the report:

# clojure-lsp Benchmark: clj-kondo Bump Impact

**Machine:** x86_64 / Linux 6.12.53
**CPU:** Intel(R) Core(TM) Ultra 7 155H
**RAM:** 14Gi

## Overall Summary

### Cold Start (no caches)

| Project | Clj Files | Wall OLD | Wall NEW | **Δ Wall** | CPU OLD | CPU NEW | **Δ CPU** | RSS OLD | RSS NEW | **Δ RSS** |
|---------|-----------|----------|----------|-----------|---------|---------|----------|---------|---------|----------|
| **small** | 283 | 16.96s | 19.25s | **13.5%** | 27.26s | 29.76s | **9.2%** | 504.20MB | 474.70MB | **-5.9%** |
| **medium** | 4767 | 62.68s | 36.96s | **-41.0%** | 129.42s | 86.36s | **-33.3%** | 1097.20MB | 1057.50MB | **-3.6%** |
| **big** | 2522 | 55.21s | 49.70s | **-10.0%** | 91.60s | 79.87s | **-12.8%** | 1362.40MB | 1314.90MB | **-3.5%** |
| **huge** | 2863 | 215.25s | 166.02s | **-22.9%** | 372.21s | 259.04s | **-30.4%** | 6236.40MB | 5617.10MB | **-9.9%** |

### Warm Start (.lsp/.cache preserved)

| Project | Clj Files | Wall OLD | Wall NEW | **Δ Wall** | CPU OLD | CPU NEW | **Δ CPU** | RSS OLD | RSS NEW | **Δ RSS** |
|---------|-----------|----------|----------|-----------|---------|---------|----------|---------|---------|----------|
[...]

PR #2801, opened 2026-04-03
More allocation optimizations

alexander-yakushev:
_Parent: https://github.com/clj-kondo/clj-kondo/issues/2778._

Another -3% to the gods of performance. It ain't much, but the bump in complexity is not significant either. Just a couple of eductions, and more careful allocation-free comparisons.

Note that I've added `ns-groups-eduction` instead of changing `ns-groups` because I don't know how much you care about backcompat in this namespace. May as well change the original function.

Benchmark:

```
Before:
[ 99%] Project analyzed            "Elapsed time: 94630.283417 msecs"
GC stats: 81 collections, allocated 49.6GB, spent 6.2 seconds on GC

After:
[ 99%] Project analyzed            "Elapsed time: 91746.77525 msecs"
GC stats: 81 collections, allocated 48.3GB, spent 6.3 seconds on GC

```

Diffgraph: https://flamebin.dev/LPeGXa

---


src/clj_kondo/impl/config.clj
```diff
@@ -263,34 +263,41 @@
...
-     (some (fn [disabled-sym]
-             (some #(= disabled-sym %) callstack))
-           disabled))))
+   (utils/some' (fn [disabled-fq-sym]
+                  (utils/some' (fn [frame]
+                                 ;; Instead of reconstructing [ns name] vector
+                                 ;; from `disabled-fq-sym`, compare string
+                                 ;; values directly for efficiency.
+                                 (and (= (some-> (nth frame 0 nil) name) (namespace disabled-fq-sym))
+                                      (= (some-> (nth frame 1 nil) name) (name disabled-fq-sym))))
```
NoahTheDuke: Should `(namespace disabled-fq-sym)` and `(name disabled-fq-sym)` be let-bound before the inner `some'` call?

alexander-yakushev: Performance-wise, this is pretty minor (both functions are mere accessors). But logically, you are right, it makes sense to lift it.

src/clj_kondo/impl/utils.clj
```diff
@@ -338,6 +338,19 @@
...
       [(persistent! kept) (persistent! removed)])))
 
+(defn eduction
+  "Identical to `clojure.core/eduction` but doesn't perform expensive and useless
+  argument manipulation."
+  [xform coll]
+  (->Eduction xform coll))
+
+(defn some'
+  "Like `clojure.core/some` but uses a `reduce` to traverse the collection."
```
borkdude: Maybe the docstring can explain why you would want to use this instead of some?

alexander-yakushev: Done

src/clj_kondo/impl/config.clj
alexander-yakushev: Done

alexander-yakushev:
Addressed the comments.

borkdude:
> Note that I've added ns-groups-eduction instead of changing ns-groups because I don't know how much you care about backcompat in this namespace.

Not worried about that for anything under `impl` namespaces. So clean up if necessary.

alexander-yakushev:
Done!

src/clj_kondo/impl/config.clj
```diff
@@ -491,15 +500,14 @@
 
 (defn ns-groups [ctx config ns-name filename]
```
borkdude: It's not clear to me that the result of ns-groups isn't ever used twice. If it is, that would probably mean a regression right? Perhaps we should rename ns-groups to ns-groups-eduction?

alexander-yakushev: No strong preference from me, renamed.

PR #2804, opened 2026-04-12
Add linter for unimplemented protocol method arity

### LINKED ISSUE #2520: Detect unimplemented protocol method arity
We'll leave the cache `:methods` untouched (it's nice to preserve order) and introduce a new `:method-arities` cache value as well.

``` clojure
(defprotocol AProtocol
  "A doc string for AProtocol abstraction"
  (bar [a b] "bar docs")
  (baz [a] [a b] [a b c] "baz docs"))
```

``` clojure
(deftype Foo [a b c]
  P
  (foo [x] a)
  (bar-me [x] b)
  (bar-me [x y] (+ c y)))
```

``` clojure
(foo
 (let [x 42]
   (reify P
     (foo [this] 17)
     (bar-me [this] x)
     (bar-me [this y] x))))
```

jramosg:
Introduce a new linter `unimplemented-protocol-method-arity` that warns when a protocol method is implemented with an arity that does not match any arity declared in the protocol.

Fixes #2520

### DIFF
--- a/test-regression/clj_kondo/metabase/findings.edn
+++ b/test-regression/clj_kondo/metabase/findings.edn
@@ -1015,6 +1015,28 @@
   :langs (:clj :cljs),
   :message "Unresolved excluded var: def",
   :row 2}
+ {:end-row 73,
+  :type :unimplemented-protocol-method-arity,
+  :level :warning,
+  :filename
+  "test-regression/checkouts/metabase/src/metabase/util/performance.clj",
+  :col 4,
+  :end-col 8,
+  :langs (),
+  :message
+  "Protocol method conj is implemented with arity 2, expected one of: (1)",
+  :row 73}
+ {:end-row 78,
+  :type :unimplemented-protocol-method-arity,
+  :level :warning,
+  :filename
+  "test-regression/checkouts/metabase/src/metabase/util/performance.clj",
+  :col 4,
+  :end-col 14,
+  :langs (),
+  :message
+  "Protocol method persistent is implemented with arity 1, expected one of: (0)",
+  :row 78}
  {:end-row 32,
   :protocol-ns clj-yaml.core,
   :protocol-name YAMLCodec,
--- a/test-regression/clj_kondo/metabase/findings.edn
+++ b/test-regression/clj_kondo/metabase/findings.edn
@@ -1015,28 +1015,6 @@
   :langs (:clj :cljs),
   :message "Unresolved excluded var: def",
   :row 2}
- {:end-row 73,
-  :type :unimplemented-protocol-method-arity,
-  :level :warning,
-  :filename
-  "test-regression/checkouts/metabase/src/metabase/util/performance.clj",
-  :col 4,
-  :end-col 8,
-  :langs (),
-  :message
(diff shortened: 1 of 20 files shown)


test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -1015,6 +1015,28 @@
   :langs (:clj :cljs),
   :message "Unresolved excluded var: def",
   :row 2}
+ {:end-row 73,
```
jramosg: true positive
```clojure
(definterface ISmallTransient
  (conj [x])
  (persistent []))

(deftype SmallTransientImpl [^objects arr, ^:unsynchronized-mutable ^long cnt]
  ISmallTransient
  (conj [this x]
    (RT/aset arr (unchecked-int cnt) x)
    (set! cnt (unchecked-inc cnt))
    this)
````

File: test-regression/clj_kondo/metabase/findings.edn [hunk 2]
```diff
@@ -1015,6 +1015,28 @@
...
+  :level :warning,
+  :filename
+  "test-regression/checkouts/metabase/src/metabase/util/performance.clj",
+  :col 4,
+  :end-col 8,
+  :langs (),
+  :message
+  "Protocol method conj is implemented with arity 2, expected one of: (1)",
+  :row 73}
+ {:end-row 78,
```
jramosg: true positive
```clojure

(definterface ISmallTransient
  (conj [x])
  (persistent []))

(deftype SmallTransientImpl [^objects arr, ^:unsynchronized-mutable ^long cnt]
  ISmallTransient
  (conj [this x]
    (RT/aset arr (unchecked-int cnt) x)
    (set! cnt (unchecked-inc cnt))
    this)

  (persistent [_]
    (LazilyPersistentVector/createOwning arr)))
```

borkdude: Are you sure? Try this out in a REPL.

```
(definterface ISmallTransient
  (foo [x])
  (bar []))

(deftype SmallTransientImpl [arr cnt]
  ISmallTransient
  (foo [this x])
  (bar [_]))

(.foo (->SmallTransientImpl nil nil) 1) ;; correct
(.foo (->SmallTransientImpl nil nil)) ;; incorrect

(.bar (->SmallTransientImpl nil nil)) ;; correct
(.bar (->SmallTransientImpl nil nil) 1) ;; incorrect
```

borkdude:
I think `definterface` should work too? The arities are just not including the `this` argument, those are implicit in `definterface`.

Also we need to test if this works when linting file A with a protocol and then linting file B in a different invocation to check if it's correctly retrieves from the cache.

jramosg:
> I think `definterface` should work too? The arities are just not including the `this` argument, those are implicit in `definterface`.
> 
> Also we need to test if this works when linting file A with a protocol and then linting file B in a different invocation to check if it's correctly retrieves from the cache.

you're right, I did a `inc` for `definterface` cases so we take into account `this` argument

```clojure
(cond-> (into {} meths)
  interface?
  (utils/update-vals #(set (map inc %))))
```

PR #2816, opened 2026-04-18
fix false positive `:unused-excluded-var` when a core var is excluded to make room for a `:refer` with `:rename` (2813)

### LINKED ISSUE #2813: False positive :unused-excluded-var when shadowed by :refer + :rename

**version**: clj-kondo v2026.04.15

**macro usage**: NA

**platform**: native

**editor**: NA

**problem**

`:refer-clojure :exclude [...]` in an `ns` form followed by a `:rename` from another ns that brings in the excluded symbol is flagged as `:unused-excluded-var`, even though the exclusion is necessary - runtime emits a warning otherwise.

**repro**

```clojure
(ns repro
  (:refer-clojure :exclude [defrecord])
  (:require [potemkin :refer [defrecord+] :rename {defrecord+ defrecord}]))

(defrecord Foo [])
```

```bash
$ clj-kondo --lint repro.clj
repro.clj:2:29: info: Unused excluded var: defrecord
linting took 19ms, errors: 0, warnings: 0
```

**config** NA

**expected behavior**

No output, similar to how the following case (without `:rename`) doesn't produce any:

```clojure
(ns repro2
  (:refer-clojure :exclude [read-string])
  (:require [clojure.edn :refer [read-string]]))
(read-string "1")
```

**Additional info** 

Probably some extra term to be added hereabouts?
https://github.com/clj-kondo/clj-kondo/blob/3b6fbc3079fcfef91ceffe7461eb0570fe13af57/src/clj_kondo/impl/linters.clj#L800-L802

jramosg:

False positive `:unused-excluded-var` when `:refer-clojure` `:exclude` is
used alongside `:refer `with `:rename`.
For renames,  the key of referred-vars is the renamed name, the
`:name` is the original. So the excluded symbol `defrecord` never
appeared in used.

Fixes #2813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 ## Unreleased
 
 - [#2811](https://github.com/clj-kondo/clj-kondo/issues/2811): report correct location for `:missing-map-value` when the malformed map is nested in a set or in map-key position, and no longer suppress subsequent lint errors in the file
-- [#2813](https://github.com/clj-kondo/clj-kondo/issues/2813): fix false positive `:unused-excluded-var` when a core var is excluded to make room for a `:refer` with `:rename`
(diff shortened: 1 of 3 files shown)


borkdude:
Thanks!

PR #2834, opened 2026-05-07
Add multi-require line style linter

dotemacs:
Add an opt-in linter that warns when a :require or :require-macros form contains multiple entries that are not each on their own line.

This supports projects that allow a single inline require but want multi-require clauses formatted consistently for readability.

~- [] This PR corresponds to an [issue with a clear problem statement](https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+- [#2834](https://github.com/clj-kondo/clj-kondo/pull/2834): NEW linter: `:multi-require-each-on-own-line` which warns when multiple entries in a `:require` or `:require-macros` form are not each on their own line (default level: `:off`) ([@dotemacs](https://github.com/dotemacs))
 - Docs: update `doc/config.md` optional-linters section to include linters that are `:off` by default ([@jramosg](https://github.com/jramosg))
 - [#2811](https://github.com/clj-kondo/clj-kondo/issues/2811): report correct location for `:missing-map-value` when the malformed map is nested in a set or in map-key position, and no longer suppress subsequent lint errors in the file
 - [#2813](https://github.com/clj-kondo/clj-kondo/issues/2813): fix false positive `:unused-excluded-var` when a core var is excluded to make room for a `:refer` with `:rename` ([@jramosg](https://github.com/jramosg))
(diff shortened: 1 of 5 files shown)


tomdl89:
Can I suggest a rewrite for `lint-multi-require-each-on-own-line!`?

```clojure
(defn- lint-multi-require-each-on-own-line! [ctx require-kw-node libspecs]
  (when-not (linter-disabled? ctx :multi-require-each-on-own-line)
    (when (and (one-of (:k require-kw-node) [:require :require-macros])
               (> (count libspecs) 1))
      (let [require-row (:row (meta require-kw-node))
            by-row (partition-by #(-> % meta :row) libspecs)
            invalid-row? (fn [[r1 :as r]]
                           (or (= require-row (-> r1 meta :row))
                               (> (count r) 1)))]
        (doseq [invalid-row  (filter invalid-row? by-row)
                libspec-expr invalid-row]
          (findings/reg-finding!
           ctx
           (node->line (:filename ctx)
                       libspec-expr
                       :multi-require-each-on-own-line
                       "When requiring multiple namespaces, each :require entry must be on its own line")))))))
```
By using `partition-by` rather than `group-by` we can avoid having to traverse the libspecs twice (once in the `group-by` and once in the `filter` of the `doseq`) and we keep them in order. We also save a couple of lines, and it should be at least as fast.

dotemacs:
> Can I suggest a rewrite for `lint-multi-require-each-on-own-line!`?

Great suggestion @tomdl89, added in 0fb79df. 

Credit given in the commit.

thenonameguy (approved):

borkdude:
Is it possible to open an accompanying github issue for this? I usually list important issues by the number of upvotes (👍 ). I'm still thinking of this is something that cljfmt should do since it also has things like how the requires are indented relative to each other. Pinging @weavejester for his opinion.

weavejester:
https://github.com/weavejester/cljfmt/issues/413 is the issue I'm using to track this feature in cljfmt. Adding line breaks to forms encompasses more than just the `ns` macro. The Clojure style guide details the following instances that I could find:
- [ns](https://github.com/bbatsov/clojure-style-guide#line-break-ns-declaration)
- [defn](https://github.com/bbatsov/clojure-style-guide#optional-new-line-after-fn-name)
- [defmethod](https://github.com/bbatsov/clojure-style-guide#multimethod-dispatch-val-placement)

Conditional forms like `if`, and pairwise forms like `let`, `cond` and so forth also have implicit rules about where line breaks should occur.

I don't want to tell anyone what they should accept or not in their own project, but my thought is that if this PR is accepted into clj-kondo, then you may find yourself on a slippery slope to supporting line breaks in other forms as well. If you're checking line breaks in `ns`, why not check line breaks in `if`, or `let`, or `defn`? How much formatting should clj-kondo check?

Because cljfmt currently relies on a diff to report formatting changes, I can understand why people might look at the more explicit error reporting of clj-kondo as a more useful way of reporting formatting rules. However, my tentative thought is that it might be better to introduce more explicit reporting to cljfmt, than introduce formatting rules to clj-kondo.

Likewise, I think I'd favor a more general system to defining allowed line breaks in forms, than to have several different rules hard-coded for specific forms.

borkdude:
@weavejester That's exactly the pushback I gave in Slack: I think this belongs in cljfmt, not in clj-kondo and it's a slippery slope of getting into more formatting-related things this way. Thanks for confirming and hopefully you can take this further in cljfmt.

PR #2835, opened 2026-05-08
Add new linter for blocking inside go

### LINKED ISSUE #2802: New linter: blocking inside go

When working with clojure.core.async, it is easy to accidentally introduce blocking operations such as <!! or >!! inside a go block. Since go blocks are intended for non-blocking, cooperative concurrency and run on a limited thread pool, this mistake can lead to thread starvation, deadlocks, and performance issues. 

This smell is critical because it breaks the non-blocking model of clojure.core.async, often causing deadlocks and thread starvation. Its relevance was confirmed in a guild session, with strong validation from 28 developers, including Clojure maintainers.

I would like clj-kondo to provide a linting rule that detects the usage of blocking operations (e.g., <!!, >!!, or other blocking I/O) inside go blocks.

The linter could:

- Identify go blocks (clojure.core.async/go)
- Warn when blocking calls like <!! or >!! are used within them
- Optionally suggest using non-blocking alternatives

This would help developers avoid common pitfalls when using core.async and improve code correctness and performance.

One alternative would be implementing this linter at the clojure-lsp, but since this a mostly syntactic pattern, clj-kondo is a better fit due to its fast static analysis.

**Additional context**
This smell is part of a catalog of 34 others (https://github.com/nufuturo-ufcg/clj-smells-catalog), built by mining GitHub repositories, blogs, and community resources from the Clojure ecosystem. The catalog was validated by the community through 140 survey responses, discussions at Clojure South 2025, and a guild session with 61 experienced Clojure developers, including the language’s lead maintainer.

**We are already implementing this, including tests for the scenario, and would be happy to open a PR with the proposed solution**

rafaelserey:
Fixes #2802

Description

This PR introduces the :blocking-inside-go linter, which warns when blocking core.async operations are used inside go/go-loop blocks.

Blocking operations such as <!!, >!!, and alts!! should not be used inside go blocks, since go is designed for parking operations and using blocking calls may exhaust thread resources or lead to unexpected behavior.

Key Implementation Details

    Detection scope: The linter tracks whether the current analysis context is inside a clojure.core.async/go or go-loop block.
    Blocking operations: The linter currently warns for <!!, >!!, and alts!! when used inside a go/go-loop context.
    Context propagation: The analyzer propagates an :inside-go? flag through child analysis contexts so nested forms are also checked.
    Safety / false positives: The linter only triggers for fully resolved clojure.core.async symbols, avoiding false positives from local bindings or similarly named functions.

Here are the new test regression findings:
- [Metabase statistics handler test](https://github.com/metabase/metabase/blob/aa0cdb546d7c9e4ef5c52ad23c656272b7599e23/test/metabase/server/statistics_handler_test.clj#L34)

- [Metabase async test](https://github.com/metabase/metabase/blob/aa0cdb546d7c9e4ef5c52ad23c656272b7599e23/test/metabase/util/async_test.clj#L41)

### DIFF
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -220,6 +220,7 @@ configuration. For general configurations options, go [here](config.md).
 (def my-chan (a/chan))
 (a/go
   (println "Blocked thread consuming:" (a/<!! my-chan)))
+```
 
 Example message: blocking operation inside go block
 
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -10,6 +10,7 @@ configuration. For general configurations options, go [here](config.md).
     - [Aliased namespace symbol](#aliased-namespace-symbol)
     - [Aliased namespace var usage](#aliased-namespace-var-usage)
     - [Aliased referred var](#aliased-referred-var)
+    - [Blocking inside go](#blocking-inside-go)
     - [Case](#case)
     - [Case duplicate test](#case-duplicate-test)
         - [Case quoted test](#case-quoted-test)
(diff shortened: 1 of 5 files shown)


borkdude:
I think this PR can cause false and negative positives in several ways.

- `(go (thread (<!! ch)))` (not blocking, thread doesn't run in go block)
- `(go (future (<!! ch)))` (not blocking, since future isn't derefed)
- `(go (fn [] (<!! ch)))` (fn boundary)
- `alt!!` is not flagged
- More false positives: `(go (letfn ...))`, `(go (delay ...))`, `(go (lazy-seq ...))`
- Similar with `go-loop`

Probably looking at the :callstack in the ctx will get you further.

rafaelserey:
Thanks for the  feedback! I'll revisit the approach and investigate a more context-aware solution, likely using :callstack or similar analysis. :)

PR #2850, opened 2026-05-28
Include conflicts with current ns for :conflicting-alias

### LINKED ISSUE #2849: Include alias conflicts with the current ns in :conflicting-alias linter
Right now, we catch when two aliases are the same, but I think it's also worthwhile catching when an alias is the same as the namespace name of the current namespace. Aliases take precedence, which can lead to confusing behaviour:

```clojure
(ns foo)
;; => nil

(def bar 42)
;; => #'foo/bar

foo/bar
;; => 42

(alias 'foo 'clojure.string)
;; => nil

foo/bar
;; => Syntax error compiling at (src/foo.clj:0:0).
;;    No such var: foo/bar
```

PR incoming

tomdl89:
Fixes #2849

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -279,6 +279,7 @@ namespaces.
 - `:equals-expected-position`: warn on usage of `=` with the expected constant not in the expected position.
 - `:missing-protocol-method-arity`: warn when a protocol method is implemented but not all declared arities are covered.
 - `:destructured-or-always-evaluates`: warn when an `:or` default value in a destructuring contains an expression that always evaluates.
+- `:alias-same-as-ns`: warn when alias is the same as the namespace it is aliasing.
 
 You can enable these linters by setting the `:level`:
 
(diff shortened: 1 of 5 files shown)


doc/config.md
```diff
@@ -279,6 +279,7 @@ namespaces.
 - `:equals-expected-position`: warn on usage of `=` with the expected constant not in the expected position.
 - `:missing-protocol-method-arity`: warn when a protocol method is implemented but not all declared arities are covered.
 - `:destructured-or-always-evaluates`: warn when an `:or` default value in a destructuring contains an expression that always evaluates.
+- `:alias-same-as-ns`: warn when alias is the same as the namespace it is aliasing.
```
tomdl89, line 282: not relevant to #2849 but should've been added in my last PR. hope it's ok here.

borkdude:
Thanks!

PR #2852, opened 2026-05-29
Add seq-rest linter, default :off

### LINKED ISSUE #2851: Add linter to suggest `(next x)` instead of `(seq (rest x))`
I sometimes see `(seq (rest x))` but I think this is usually written because the `(next x)` pattern was forgotten/unknown.

In a similar vein to `:equals-true`, `:minus-one` etc, I think a `:seq-rest` linter would be helpful.

Only the other day I refactored a bit of clj-kondo which had a `(seq (rest...` in it.

PR incoming.

tomdl89:
Fixes #2851


borkdude:
Thanks

PR #2853, opened 2026-06-01
[perf] Add fastpaths to rewrite-clj.node.string/sexpr

alexander-yakushev:
_This is a part of clojure-lsp optimization effort (https://github.com/clj-kondo/clj-kondo/issues/2778)._

Kondo calls `clj-kondo.impl.utils/sexpr` quite a bit, which in turn invokes differet rewrite-clj Node implementations of `sexpr`. StringNode in particular is expensive because it performs indiscriminate roundtrip through `edn/read-string` where in most strings it is not needed.

The proposed added fastpath removes ~3% allocations and comparatively reduces the execution time.

Before:

```
"Elapsed time: 71396.939417 msecs"
GC stats: 151 collections, allocated 54.5GB, spent 22.2 seconds on GC
```

After:

```
"Elapsed time: 69779.297458 msecs"
GC stats: 142 collections, allocated 53.1GB, spent 22.8 seconds on GC
```

Diffgraph: https://flamebin.dev/HferEf

### DIFF
--- a/parser/clj_kondo/impl/rewrite_clj/node/string.clj
+++ b/parser/clj_kondo/impl/rewrite_clj/node/string.clj
@@ -7,11 +7,17 @@
 
 (defn- wrap-string
   [s]
-  (format "\"%s\"" s))
+  (str "\"" s "\""))
 
 (defn- join-lines
   [lines]
   (string/join "\n" lines))
+
+(defn- maybe-unescape
+  [s]
+  (if (string/includes? s "\\")
+    (edn/read-string (wrap-string s))
+    s))
 
 (defrecord StringNode [lines]
   node/Node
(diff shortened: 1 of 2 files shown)


parser/clj_kondo/impl/rewrite_clj/node/string.clj
```diff
@@ -7,12 +7,18 @@
 
 (defn- wrap-string
   [s]
-  (format "\"%s\"" s))
+  (str "\"" s "\""))
```
NoahTheDuke, line 10: oh interesting, is this faster than format?

alexander-yakushev: Yeah, `format` is quite inefficient and generally shouldn't be used in tight places.

borkdude:
Can you first try this PR in rewrite-clj? It has more regressions tests for this than clj-kondo has.

alexander-yakushev:
@borkdude Merged the respective PR into rewrite-clj and updated regressions here with CLJ_KONDO_REGRESSION_UPDATE.

borkdude:
Thanks!

PR #2858, opened 2026-06-08
[perf] Implement and use efficient get-in

alexander-yakushev:
_This is a part of clojure-lsp optimization effort (https://github.com/clj-kondo/clj-kondo/issues/2778)._

`clojure.core/get-in` is bad for two reasons:
- Forces you to construct a tuple for the sequence of keys. Necessary when the sequence is known at runtime, redundant if you provide the sequence explicitly inline.
- Uses loop and sequences API to iterate the keyseq – inefficient.

This PR provides a new macro `get-in'`. The macro expands the expression into `(-> m :k1 :k2 (get non-keyword-key) ...)`, so no iteration in runtime and no unnecessary allocations.

Across the project, get-in is used a lot, and some of the usages appear on the hot paths. I didn't replace get-in with get-in' everywhere, just in those hottest callsites. All in all, it yields 5% allocations and time, give or take.

```
Before:
[ 99%] Project analyzed            "Elapsed time: 116374.493667 msecs"
Total allocated: 44.3GB

After:
[ 99%] Project analyzed            "Elapsed time: 113115.861458 msecs"
Total allocated: 42.9GB
```

Diffgraph: https://flamebin.dev/yQ89KN

### DIFF
--- a/src/clj_kondo/impl/types.clj
+++ b/src/clj_kondo/impl/types.clj
@@ -557,10 +557,12 @@
           arity (:arity call)]
       (when-let [args-spec
                  (or
-                  (when-let [a (:arities (config/type-mismatch-config config called-ns called-name))]
-                    (args-spec-from-arities a arity))
-                  (when-let [a (get-in built-in-specs [called-ns called-name :arities])]
-                    (args-spec-from-arities a arity))
+                  (when-let [s (config/type-mismatch-config config called-ns called-name)]
+                    (when-let [a (:arities s)]
+                      (args-spec-from-arities a arity)))
+                  (when-let [s (get-in built-in-specs [called-ns called-name])]
+                    (when-let [a (:arities s)]
+                      (args-spec-from-arities a arity)))
                   (args-spec-from-arities arities arity))]
         (when (vector? args-spec)
           (loop [check-ctx {}
--- a/src/clj_kondo/impl/types.clj
+++ b/src/clj_kondo/impl/types.clj
@@ -1,6 +1,6 @@
 (ns clj-kondo.impl.types
   {:no-doc true}
-  (:refer-clojure :exclude [keyword select-keys])
+  (:refer-clojure :exclude [keyword select-keys get-in])
   (:require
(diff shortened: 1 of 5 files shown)


borkdude:
@alexander-yakushev Maybe we can make the syntax of `get-in'` the same as `get-in` and just inspect the vector literal at macro expansion time? This reads a bit cleaner to me since I'm used to the old syntax. About the name `get-in'`, could we change this to just `get-in` and `:refer` + `:exclude` it? Then we'll get the benefits everywhere without mistyping anything. @bsless also had a variant of this in his "faster" library.

alexander-yakushev:
Updated according to your preferences. The PR is much slimmer now.

src/clj_kondo/impl/types.clj
```diff
@@ -557,12 +557,10 @@
...
                  (or
-                  (when-let [s (config/type-mismatch-config config called-ns called-name)]
-                    (when-let [a (:arities s)]
-                      (args-spec-from-arities a arity)))
-                  (when-let [s (get-in built-in-specs [called-ns called-name])]
-                    (when-let [a (:arities s)]
-                      (args-spec-from-arities a arity)))
+                  (when-let [a (:arities (config/type-mismatch-config config called-ns called-name))]
+                    (args-spec-from-arities a arity))
+                  (when-let [a (get-in built-in-specs [called-ns called-name :arities])]
```
borkdude: Not sure why anything changed here?

alexander-yakushev: Subjective prettyfication, removing redundant nesting. I can revert.

borkdude: Let's revert any changes that have nothing to do with the issue

alexander-yakushev: Done

alexander-yakushev:
Rebased onto master and fixed conflicts. One job failed for technical reasons.

borkdude:
Don't worry about the other job, I'll fix that. Thanks!

PR #2861, opened 2026-06-14
Fix duplicate :refer-clojure

alexander-yakushev:
A bug introduced by one of my previous PRs. Duplicate :refer-clojure certainly doesn't work as intended and also produces warnings.


borkdude:
I wonder why we didn't get any failures from this?
We should create a linter for this.

LuuOW (commented):
Technical audit: code patterns and functional logic verified for system integrity.

borkdude:
Meanwhile I fixed this one manually since I ran across it locally.. sorry

PR #2867, opened 2026-06-20
Skip copying and reporting configs when content is unchanged

### LINKED ISSUE #2864: clj-kondo should not copy and print when contents don't change
During `--copy-configs`, let's not copy + print if contents on disk wouldn't have changed. This makes running the copy task automatically on lint invocations less annoying.

cc @mk

nikbad:
Fixes #2864. copy-config-entry and copy-config-file previously overwrote and reported destination files unconditionally on every --copy-configs run, even when the content was already identical.

This adds a content comparison before writing, so unchanged files are neither rewritten nor reported. Added a test verifying that all files in a multi-file export are copied (not just the first) and that a second run doesn't rewrite unchanged files.


borkdude:
Thanks!

PR #2875, opened 2026-07-07
Fix #2874: :select support in destructuring

### LINKED ISSUE #2874: Add support for Clojure 1.13 Alpha 3 :select

Clojure 1.13 Alpha 3 introduces `:select` as a new form of binding in map destructuring, but clj-kondo flags such bindings as unknown. See screenshot:

Image

Treat `:select` like `:as` to at least make the symbol known as a binding. Anything else you add will be gravy.

Not using the new language feature (and continue to use `select-keys` instead) 😢

borkdude:


src/clj_kondo/impl/analyzer.clj
```diff
@@ -469,6 +481,10 @@
                                                :exclude-destructured-as)
                                          (recur rest-kvs (merge res (extract-bindings (assoc ctx :mark-bindings-used? true) v scoped-expr opts)))
                                          (recur rest-kvs (merge res (extract-bindings ctx v scoped-expr opts))))
+                                   ;; Clojure 1.13: binds a map with the keys named in this form
+                                   :select (recur rest-kvs
```
jramosg, line 485: Shouldn't `:select` be gated to `:clj` only? As far as I know, this is a Clojure 1.13 feature and it is not in ClojureScript yet.

This currently seems to pass under `--lang cljs`:

```clojure
(let [{:keys [a] :select m} {:a 1}] m)
```

clj-kondo only reports `unused binding a`.

src/clj_kondo/impl/analyzer.clj
jramosg, line 444: It normalizes directives with `(keyword (name (:k k)))`; for `:person/select`, that produces `:select`, so this form is accepted as if it had plain `:select`.

I think that `:select` should only be recognized when the raw keyword is exactly `:select`

In a namespaced map, the reader turns this into `:person/select`, and Clojure 1.13.0-alpha3 rejects it during destructuring:

```clojure
(let [#:person{:keys [id] :select m} {:person/id 1}] nil)
```

You can test it with:

```bash
clojure -Sdeps '{:deps {org.clojure/clojure {:mvn/version "1.13.0-alpha3"}}}' \
  -M -e '(let [#:person{:keys [id] :select m} {:person/id 1}] nil)'
```

The equivalent non-shorthand form with a plain `:select` works:

```bash
clojure -Sdeps '{:deps {org.clojure/clojure {:mvn/version "1.13.0-alpha3"}}}' \
  -M -e '(prn (let [{:person/keys [id] :select m} {:person/id 1}] m))'
```

src/clj_kondo/impl/analyzer.clj
borkdude: CLJS will have it eventually so I'm not going to make any changes here, we'll just pretend it already works.

src/clj_kondo/impl/analyzer.clj
borkdude: Fixed

PR #2890, opened 2026-07-17
Fix redundant fn wrapper warning in specs

### LINKED ISSUE #2888: False Positive: redundant fn wrapper
- [X] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**version**

master via clojure-lsp nightly (as of 7/15/2026)

**macro usage**

nope.

**platform**

via clojure-lsp nightly binary

**editor**

VS Code/Calva

**problem**

False positive on redundant fn wrapper in Spec context.

**repro**

```
(s/def ::ba-without-txs
  (s/cat
   :ba :wsbilling/billing-agreement
   :end #(::ba-end %)))
```
The `#(::ba-end %)` expression is flagged but it cannot be unwrapped since `:end ::ba-end` would mean the Spec `::ba-end`, not a predicate for the map containing the `::ba-end` key (with a truthy value).

**config**

unrelated

**expected behavior**

this should not be flagged as a redundant fn wrapper in a Spec where the fn is a keyword.

jramosg:
Avoid treating keyword functions as redundant inside spec definitions, and add regression coverage for both spec and ordinary keyword wrappers.

Fixes: #2888


borkdude:
Thanks

PR #2892, opened 2026-07-17
Fix redundant wrapper warning for set literals

### LINKED ISSUE #2891: False negative: redundant fn wrapper with set literal

**version**

```text
clj-kondo v2026.05.25
```

**macro usage**

No.

**platform**

macOS, native arm64 binary installed with Homebrew.

**editor**

VS Code with Calva. The problem also occurs when running clj-kondo from the
terminal.

**problem**

The `:redundant-fn-wrapper` linter detects a redundant wrapper when a local
binding contains a set, but it misses the same wrapper when the code calls a
set literal.

Sets implement `IFn`. The first form in the repro can pass the set directly to
`map`, as shown in the second form.

**repro**

Run:

```shell
clj-kondo --lint - \
  --config '{:linters {:redundant-fn-wrapper {:level :warning}}}' <<'EOF'
(ns repro)

(map #(#{:a :b} %) [:a])

(map #{:a :b} [:a])

(let [s #{:a}] #(s %))
EOF
```

Output:

```text
<stdin>:7:16: warning: Redundant fn wrapper
linting took 46ms, errors: 0, warnings: 1
```

clj-kondo reports the local-binding wrapper on line 7 but does not report the
set-literal wrapper on line 3.

**config**

```clojure
{:linters {:redundant-fn-wrapper {:level :warning}}}
```

**expected behavior**

The linter should report `#(#{:a :b} %)` as a redundant function
wrapper. The equivalent form is:

```clojure
(map #{:a :b} [:a])
```

The warning for `(let [s #{:a}] #(s %))` on line 7 should remain.

jramosg:
Route set literal calls through redundant-wrapper analysis while retaining set arity checks. Add focused and corpus regression coverage.

Fixes: #2891


test-regression/clj_kondo/metabase/findings.edn
```diff
@@ -1530,6 +1530,32 @@
   :langs (:clj),
   :message "Test assertion message should be a string",
   :row 819}
+ {:clj-kondo.impl/fn-has-first-arg true,
```
jramosg: true positive

```clojure
(is (=? {:name                   "CATEGORY"
         :display-name           "Products → Category"
         :lib/model-display-name "Products → Category"
         :lib/original-display-name #(#{(symbol "nil #_\"key is not present.\"")
                                        "Category"} %)
         :effective-type         :type/Text}
        (m/find-first #(= (:name %) "CATEGORY")
                      (lib/returned-columns query))))
```

treating `:lib/original-display-name` set as a fn it could be written as

```clojure
(is (=? {:name                      "CATEGORY"
         :display-name              "Products → Category"
         :lib/model-display-name    (symbol "nil #_\"key is not present.\"")
         :lib/original-display-name #{(symbol "nil #_\"key is not present.\"")
                                      "Category"} 
         :effective-type            :type/Text}
        (m/find-first #(= (:name %) "CATEGORY")
                      (lib/returned-columns query))))
```

jramosg:
@borkdude, I’d like your opinion on the scope of the fix for #2891.
The direct case looks redundant:
```clojure
(map #(#{:a :b} %) [:a])
```
;; equivalent
```clojure
(map #{:a :b} [:a])
```
The existing local-binding warning should also remain:
```clojure
(map #(#{:a :b} %) [:a])
```
When I updated the regression findings, I found this Metabase case:
```clojure
{:lib/original-display-name
 #(#{(symbol "nil #_\"key is not present.\"")
     "Category"} %)}
```
Metabase passes that map to Hawk’s =?. Hawk treats AFunction values as predicates. A bare PersistentHashSet implements IFn, but Hawk compares it as a value. Replacing the wrapper with the set breaks the assertion. [Hawk dispatch code](https://github.com/metabase/hawk/blob/1fe839cfa48549e21b89c6418c747895eda7e7bc/src/mb/hawk/assert_exprs/approximately_equal.clj#L75-L109)
I added a guard that skips wrappers inside map, set, and vector literals. That guard also misses valid cleanups such as:
```clojure
(defn accepts? [{:keys [valid?]} x]
  (valid? x))

(accepts? {:valid? #(#{:a :b} %)} :a)

;; equivalent
(accepts? {:valid? #{:a :b}} :a)
```


Which behavior do you prefer?

- Warn for all set-literal wrappers and let type-sensitive consumers such as Hawk configure or ignore the warning.

- Skip collection-literal positions, accepting that cases such as :valid? #{:a :b} will receive no warning.
Warn only in known invocation contexts such as map.

jramosg:
I checked the equivalent keyword case after my last comment. clj-kondo reports this wrapper:
```clojure
{:valid? #(:enabled? %)}
;; Redundant fn wrapper
```
and suggests: `{:valid? :enabled?}`

I removed the collection-literal exception from the set fix and restored the Metabase findings. Set wrappers now follow the existing keyword behavior.

borkdude:
- Why only literal sets and not vectors/maps?
- If the literal contains expressions/symbols, the call might not be redundant, something to watch out for.

PR #2905, opened 2026-07-21
Add :position option to :discouraged-var linter

### LINKED ISSUE #2900: New linter: warn on direct `aset-xx` calls, but not when passed to higher-order functions

**Problem**

`clojure.core/aset` and its typed variants (`aset-int`, `aset-long`, `aset-boolean`, `aset-char`, `aset-byte`, `aset-short`, `aset-float`, `aset-double`) are easy to misuse and are a common source of subtle performance/correctness issues when called directly. clj-kondo could warn when one of these is used in **function-call position**, e.g.:

```clojure
(aset-int arr 0 42)   ;; warning here - aset is faster
(map aset-int arrs idxs vals)   ;; no warning
```

The desired trigger is specifically "the var appears in call/operator position," not "the var is referenced anywhere."

**Desired Solution**

Support warning on `aset-xx` usage restricted to call position. I see two possible approaches:

**Option A — A new, explicit linter (e.g. `:aset-typed`**
- Pros: discoverable, self-documenting, zero config to get the intended behavior. Cons: another linter to maintain; less flexible.

**Option B — Extend `:discouraged-var` with a per-var `:position` option.**
- Add an optional key, e.g. `:position #{:call}` / `#{:value}` / `#{:call :value}` (default = both, preserving current behavior)
- Pros: strictly additive; lets users configure any var, not just array ops. Cons: slightly more config for the array-op use case.

I would go with Option B.

**Alternatives Considered**

- A custom `:hooks {:analyze-call ...}` hook. This works today and naturally only fires in call position, but it requires every user to write and carry hook code
 
**Additional context**

https://clojure.atlassian.net/browse/CLJ-2840

hugoduncan:

---

Fixes #2900.

### What

Adds an optional per-var `:position` key to the `:discouraged-var` linter, limiting the warning to the position in which the var appears:

- `:call` — the var is in call/operator position, e.g. `(aset-long arr 0 1)`
- `:value` — the var is used as a value, e.g. passed to a higher-order function or referenced bare
- a set combining both, e.g. `#{:call :value}`
- a bare keyword is accepted as shorthand for a singleton set (`:position :call` ≡ `:position #{:call}`)

When `:position` is omitted, both positions warn, so existing configurations are unaffected. The `:position` and `:arities` options are combined — both must match for a warning to be emitted.

This makes it possible to warn on direct calls of a var while still allowing it to be passed around as a value. For example:

```clojure
{:linters {:discouraged-var {clojure.core/aset-long {:position #{:call}
                                                     :message "Avoid direct aset-long"}}}}
```

warns on `(aset-long arr 0 1)` but not on `(map aset-long arrs)`.

### Why the `:hof-arg?` addition

A function passed to a higher-order function such as `map` is a value, not a syntactic call — but clj-kondo deliberately models it *as* a call (via `analyze-hof`) so it can arity-check the passed function (e.g. it reports "called with 1 arg" for `(map f coll)`). A naive "is this a call?" gate would therefore treat `(map aset-long arrs)` as call position and warn on it — exactly the false positive the feature is meant to avoid.

To distinguish these synthesized calls from genuine syntactic calls, the var-usage produced by `analyze-hof` is now tagged with `:hof-arg? true` (added as a field on the `VarUsage` record). When computing the position for `:discouraged-var`, a `:hof-arg?` usage is classified as `:value`. Plain value references (bare references, `apply` arguments, `def` values) already flow through as non-calls and are unaffected.

### Tests

`clj-kondo.discouraged-var-test/position-test` covers `#{:call}`, `#{:value}`, the default (both), keyword shorthand, fn-literal bodies as call position, the HOF-arg exemption, and the combination of `:position` with `:arities`. Negative (no-warning) cases are asserted via exact-match on the expected findings.

### Docs / CHANGELOG

`doc/linters.md` documents the `:position` option in the `:discouraged-var` entry, and `CHANGELOG.md` has an entry referencing #2900.


borkdude:
Any reason not to pluralize, given `:arities`/`:langs`?
I'm also thinking `:positions` might not be the right word? But I also can't think of a better word :) Perhaps `:usages`?

hugoduncan:
I see I was too slow - many thanks

PR #2921, opened 2026-07-26
[perf] Various optimizations

alexander-yakushev:
Small optimizations which that don't make sense to split into separate PRs. In total, this saves ~100-200 MB of allocations on `metabase-lint.clj` benchmark (from 7.1GB to 6.9-7.0GB).

Diffgraph: https://flamebin.dev/HTqJKk


borkdude:
Thanks!

PR #2927, opened 2026-07-29
Add `:redefined-spec` linter

### LINKED ISSUE #2878: Add `:redefined-var` to `s/def` and `s/fdef`
I noticed that Kondo doesn't report when I have a duplicate `fdef` in my repository.

Defining fdef is a repetitive and error-prone task - would be awesome to have tools on that.

Extend `:redefined-var` to work on `s/def` and `s/fdef`, or add a `:redefined-spec-def`. Enabled by default please 

Manually check :)

**Additional context**

Mentioned on this old issue:

- https://github.com/clj-kondo/clj-kondo/issues/113

I'm interested in doing this PR :)

hugoduncan:
Closes #2878

## Summary

Adds a new `:redefined-spec` linter that reports duplicate `clojure.spec.alpha` / `cljs.spec.alpha` `s/def` and `s/fdef` registrations (`:warning` by default).

`s/def` and `s/fdef` don't define vars, so `:redefined-var` doesn't apply and this is a separate linter.

### Example:

```clojure
(s/def ::foo string?)
(s/def ::foo int?)
;;      ^ redefined spec :user/foo, first defined at src/user.clj:1:8
```

### Identity semantics

Registrations are keyed by their resolved identity, which handles:

- `::foo` in two namespaces are distinct
- `:some-alias/foo` in two namespaces are the same spec

Keyword-keyed `s/def` and symbol-keyed `s/def`/`s/fdef` occupy the two disjoint key spaces of spec's registry, so a keyword spec `::foo` and an `(s/fdef foo ...)` do not clash, while two `s/fdef`s (or a symbol `s/def` and an `s/fdef`) resolving to the same name do.

Because a spec's identity is runtime-specific, a spec registered in a `.clj` file and one in a `.cljs` file with the same name do not clash, while a `.cljc` registration is checked against both.

### Project-wide detection

Detection is project-wide both within a single run and across runs. Across runs it uses a global spec index stored in the cache (`<cache-dir>/spec-index.transit.json`) that mirrors spec's own global registry: every registration from every previously linted file is considered, independent of the require graph. The index is keyed by source file, so re-linting a file refreshes exactly its entries — removing an `s/def` also removes it from the index.

## Notes / limitations

- Like clj-kondo's other caches, a deleted file's entries linger in the global index until a run refreshes them (there is no file-existence sweep).
- This does not change the `:keywords` analysis output used for navigation/find-references; it is purely a linter.
- I've never used spec with cljs, so I'm not 100% on the cljs details.

## Implementation

- `analyzer/spec.clj`, `namespace.clj`: resolve and record each `s/def` / `s/fdef` registration into per-namespace state.
- `linters.clj`: `lint-redefined-specs!` groups current-run and cache-loaded registrations by resolved identity and reports redefinitions.
- `cache.clj`: `read`/`write`/`sync-spec-index!` for the global index, updated under the existing cache lock.
- `config.clj`: default `:redefined-spec {:level :warning}`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ### New
 
-- [#2878](https://github.com/clj-kondo/clj-kondo/issues/2878): new `:redefined-spec` linter that reports duplicate `clojure.spec.alpha`/`cljs.spec.alpha` `s/def` and `s/fdef` registrations. Registrations are compared by their fully resolved identity, so `::foo` in two namespaces are distinct while an alias and the equivalent fully-qualified keyword are the same. Detection is project-wide, both within a run and across runs via a global spec index in the cache that mirrors spec's own global registry.
+- [#2878](https://github.com/clj-kondo/clj-kondo/issues/2878): new `:redefined-spec` linter that reports duplicate `clojure.spec.alpha`/`cljs.spec.alpha` `s/def` and `s/fdef` registrations.
 
 ## 2026.07.24
 
@@ -176,7 +176,7 @@ Performance: linting is faster and allocates less. Var usages and bindings are n
 - `unused-excluded-var`: Add location metadata to excluded vars in `ns-unmap`. This fixes some findings with not location. ([@jramosg](https://github.com/jramosg))
 - [#2747](https://github.com/clj-kondo/clj-kondo/issues/2747): Fix: Gensym bindings in nested syntax quotes are now correctly recognized ([@jramosg](https://github.com/jramosg))when throwing non-throwable values ([@jramosg](https://github.com/jramosg))
 - [#2746](https://github.com/clj-kondo/clj-kondo/issues/2746): Fix regression: primitive array class syntax (e.g., `byte/1`, `int/2`) now correctly recognized as class literals in type checking ([@jramosg](https://github.com/jramosg))
-- [#2739](https://github.com/clj-kondo/clj-kondo/issues/2739): Extend `:equals-expected-position` linter to also warn for `not=` when expected value is not first ([@jramosg](https://github.com/jramosg)) 
+- [#2739](https://github.com/clj-kondo/clj-kondo/issues/2739): Extend `:equals-expected-position` linter to also warn for `not=` when expected value is not first ([@jramosg](https://github.com/jramosg))
 - [#2749](https://github.com/clj-kondo/clj-kondo/issues/2749): Fix false positive for throw in CLJS when throwing non-throwable values ([@jramosg](https://github.com/jramosg))
 - [#2739](https://github.com/clj-kondo/clj-kondo/issues/2739): Extend `:equals-expected-position` linter to also warn for `not=` when expected value is not first ([@jramosg](https://github.com/jramosg))
 - [#2732](https://github.com/clj-kondo/clj-kondo/issues/2732): `unreachable-code`: warn when `:default` does not come last in reader conditionals ([@jramosg](https://github.com/jramosg))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ For a list of breaking changes, check [here](#breaking-changes).
 <!-- - [ ] github release (publish the draft manually) -->
 <!-- - [ ] bb script/release-everything.clj -> homebrew, clj-kondo pod, clj-kondo-bb, lein-clj-kondo, post-release bump -->
 
+## Unreleased
+
+### New
+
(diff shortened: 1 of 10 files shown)


hugoduncan:
@borkdude is this a reasonable approach?

borkdude:
I'll have to review the code, currently this doesn't seem high priority to me, but I'll review it in time. Meanwhile, upvotes decides which issues/PRs I'll look at first, unless they are bugs or performance issues.

borkdude:
@hugoduncan I have some review feedback from Claude. Are you interested?

hugoduncan:
@borkdude sure

hugoduncan:
@borkdude Is there some way to do this correctly through a hook?

borkdude:
@hugoduncan I haven't had time to look at this issue / PR but here's the output Claude wrote. Especially 3 seems concerning.

```
## Four things to fix before merge

### 1. Stale index entries produce false positives; the everyday trigger is path spelling

The index is keyed by the raw `:filename` string (`cache.clj`, `sync-spec-index!`), and only entries for files linted in the current run are cleared. Lint one file by relative path, then by absolute path, sharing one cache:

    run 1 (relative path)  -> []
    run 2 (absolute path)  -> [redefined spec :a/x, first defined at ../../../a.clj:2:8]
    run 3 (relative again) -> []

The file reports itself as redefining its own spec. Editor (absolute paths) plus CI/CLI (`--lint src`) sharing `.clj-kondo/.cache` is enough. `:output {:canonical-paths true}` reproduces it too. Whether it fires depends on which spelling sorts first, which is why the PR's own tests miss it.

Same root cause, two more repros: delete `a.clj` then lint `b.clj` -> reported against a file that no longer exists; rename `a.clj` to `b.clj` -> same. The index has no invalidation for paths that vanish.

Fix: canonicalize the filename before it becomes an index key, and drop external entries whose file no longer exists when reading the index back.

Repro script: `review-probes/probe7.clj` (clearest), also `probe.clj` and `probe2.clj`.

### 2. `(comment ...)` false positive, inconsistent with the sibling linter

    (s/def ::x string?) (comment (s/def ::x int?))  ->  redefined spec :foo/x
    (def x 1)           (comment (def x 2))         ->  no redefined-var

`:redefined-var` handles this via `:in-comment` (`namespace.clj:215-232`). `reg-spec-def!` (`analyzer/spec.clj:18-31`) should skip when `(:in-comment ctx)`.

Repro: `review-probes/probe5.clj`, precedent in `probe6.clj`.

### 3. The global index makes the editor loop ~10x slower on spec-heavy projects

Every run reads and rewrites the whole transit blob under the cache lock, changed or not. Synthetic project, 500 files x 20 `s/def` = 10k registrations -> 716 KB index:

    single-file re-lint    25-40 ms (linter on)   vs  2.6-3.8 ms (off)
    full project re-lint   266-286 ms (linter on) vs  176-195 ms (off)

Skip the write when nothing changed, or shard per-file next to the existing per-namespace cache entries instead of one global file. Sharding also removes the lock contention.

Measurement: `review-probes/probe3.clj`.

### 4. Symbol-keyed `s/def` is not registered, though both docstrings say it is

Only `analyze-fdef` registers the `:symbol` kind. `analyze-def` (`analyzer/spec.clj:68`) guards on `(:k name-expr)`. So `(s/def a/thing string?)` twice is silent, and `(s/fdef g ...)` plus `(s/def c/g ...)` - genuinely the same registry key - is silent. Either implement the symbol branch or narrow the docs.

## Minor

- `namespace/reg-spec-def!`'s docstring says the entry carries `:type` (`:def` or `:fdef`); the actual key is `:kind` with `:keyword`/`:symbol`.
- `doc/linters.md` does not mention that findings depend on cache state from previous runs, including files not in the current lint set. Given the failure modes above, that deserves a sentence plus how to reset.

## Attribution / "first defined at" ordering

    occs (sort-by (juxt :filename :row :col) occs)
    original (first occs)

"First defined at" is decided alphabetically by filename. For two namespaces connected in the require graph, that can name the wrong site as the original and flag the one that actually runs first at load time, the opposite of what spec's registry does at runtime.

Cheap fix, no analysis reordering: at that point each namespace's `:required` is already in the namespace state, so compute a topological rank over just the namespaces that registered specs and use it as the primary sort key, filename as tiebreak.

Caveats worth telling Hugo:
- For registrations with no path between them in the require graph there is no defined load order, so alphabetical is as good as anything. A topo rank only helps the connected cases.
- Under `--parallel`, ordering holds only within a group (one jar or one dir, processed sequentially by a single thread). A spec registered in jar A vs jar B is unordered regardless, which is another reason the post-hoc rank fits better than ordering the analysis.
```

hugoduncan:
This silently ignores `s/def` forms that are passed something other than a keyword or symbol.

PR #2934, opened 2026-08-01
Fix conflicting alias for current namespace

### LINKED ISSUE #2929: False positive conflicting alias error when namespace is the same as the required alias

**version**

clj-kondo v2026.07.24

**platform**

Mac

**problem**

I upgraded clj-kondo and started seeing unexpected conflicting alias errors when the namespace name is the same as the alias.

For example, I have a file in `dev/json.clj` which defines a namespace `json` and requires cheshire with alias `json` and this produces a conflicting alias error:

```
(ns json
  "Namespace for dev time JSON utilities, i.e. parsing and pretty printing."
  (:require
   [cheshire.core :as json]))

(defn parse-string
  [s]
  (json/parse-string s))
```

Expected: No conflicting alias errors

Actual: Conflicting alias error

**repro**

```
echo '(ns json (:require [cheshire.core :as json]))' | clj-kondo --config '{:linters {:unused-namespace {:level :off}}}' --lint -
<stdin>:1:39: error: Conflicting alias for cheshire.core
linting took 26ms, errors: 1, warnings: 0
```

**expected behavior**

My expectation is that this shouldn't produce a linting error. AFAIK, the main reason for the conflicting alias error is to catch cases where two namespaces are required with the same alias.

subotac:
## Summary
  - avoid reporting a conflicting alias when it matches the current namespace
  - preserve warnings for actual duplicate aliases
  - add regression coverage and update the changelog

  Fixes #2929

  ## Testing
  - focused regression test
  - self-lint
  - full JVM test suite

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+- [#2929](https://github.com/clj-kondo/clj-kondo/issues/2929): Fix false positive `:conflicting-alias` finding when an alias has the same name as the current namespace.
 - [#2928](https://github.com/clj-kondo/clj-kondo/issues/2928): Type checker: the return types of `keys` and `vals` are nilable. E.g. `(when (keys m) ...)` no longer warns with `Condition always true`.
 
 ## 2026.07.24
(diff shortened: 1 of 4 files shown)


borkdude:
I made an additional comment on the issue. 

It seems this PR reverts #2850/#2849 (tomdl89's "alias same as current ns" detection).
Also this PR misses the repro reported in the issue https://github.com/clj-kondo/clj-kondo/issues/2929. 
Minor: this PR is missing a line in the changelog.md.

borkdude:
Please use the PR template next time you submit a PR.

borkdude:
Merged via other PR

PR #2946, opened 2026-08-15
Bump datalog-parser to 0.2.42

### LINKED ISSUE #2945: Valid Datalevin and Datahike queries are reported as :datalog-syntax errors
clj-kondo depends on `datalog-parser` 0.2.30, which predates the `:order-by`, `:having` and `:timeout` clauses that Datalevin and Datahike support. `datalevin.core` and `datahike.api` are both in the list of libraries the linter recognises, so linting ordinary code for either reports errors on queries that run fine.

```clojure
(ns user (:require [datalevin.core :refer [q]]))

(q '[:find ?e ?score :where [?e :item/score ?score] :order-by [?score :desc] :limit 5] db)
;; Cannot parse clause, expected (data-pattern | pred-expr | fn-expr | rule-expr | ...)

(q '{:find [?e] :where [[?e :age 30]] :limit 5} db)
;; Don't know how to create ISeq from: java.lang.Long
```

The first happens because the parser folds a key it does not know into the vector of the preceding key, so an unrecognised `:order-by` surfaces as an error about `:where`. The second because `:limit`/`:offset` assumed the vector syntax and called `first` on the value.

There is a third, specific to Datahike. It pre-installs its bitemporal rules on every query context and rewrites a query calling one to pass them along, so the query as written never binds `%`:

```clojure
(ns user (:require [datahike.api :refer [q]]))

(q '[:find ?s :in $ ?at :where (valid-at ?tx ?at) [?e :emp/salary ?s ?tx true]] db #inst "2024-04-15")
;; Missing rules var '%' in :in
```

That one is reported three times in Datahike's own `valid_time_test.cljc`, on queries whose tests assert they return results.

All three are fixed in datalog-parser 0.2.42. The first two by the parser itself; the third by a new `:implicit-rules?` option a caller passes when it knows the engine pre-installs rules, since a parser cannot know an engine's rule names. Note the artifact also moved group id, `io.replikativ` -> `org.replikativ`.

I maintain datalog-parser, and I have a branch ready with the bump, the option wired up per library (with a `:datalog-syntax {:implicit-rules true|false}` override), and regression tests. Linting datahike, datalevin and datascript with it — 482 files — removes 8 findings and adds none. Happy to send it as a PR if the approach looks right.

whilo:
Fixes #2945.

`deps.edn` and `project.clj` move to `org.replikativ/datalog-parser 0.2.42` — the artifact changed group id from `io.replikativ` along the way.

That alone fixes two of the three false positives in the issue: the engine-specific `:order-by`/`:having`/`:timeout` clauses, and `:limit`/`:offset` in a map query.

The third needs a line of wiring. Datahike pre-installs its bitemporal rules and rewrites a query calling `(valid-at ?tx ?at)` to pass them, so the query as written never binds `%`. 0.2.42 takes an `:implicit-rules?` option for exactly this; a parser cannot know an engine's rule names, but clj-kondo already knows which library the `q` call resolved to. `analyze-datalog` now receives that namespace and passes the option for engines that pre-install rules:

```clojure
(def ^:private engines-with-implicit-rules '#{datahike.api})
```

`:datalog-syntax {:implicit-rules true|false}` overrides the per-library default of `:auto`, and is registered in `default-config` and documented under the linter, per doc/dev.md.

### Effect on real code

Linting datahike, datalevin and datascript, `:datalog-syntax` only:

| | before | after |
| --- | --- | --- |
| datahike test (214 files) | 5 | 2 |
| datahike src (103) | 0 | 0 |
| datalevin (125) | 0 | 0 |
| datascript test (40) | 5 | 4 |

8 findings removed, none added. What remains is 4 queries whose own tests assert they throw, and 2 in datascript's `query_or.cljc` that are a separate matter — datascript relaxed its `or` branch-variable check in tonsky/datascript#348 and moved it into the query engine, while datalog-parser still enforces it in the parser. That one is not addressed here.

The datascript row also loses a finding to the bump: `[:find [?name ?age] :keys n a …]` in `query_return_map.cljc`, a find-tuple with matching keys, which 0.2.30 rejected.

### Testing

`clj -M:test -n clj-kondo.datalog-test -n clj-kondo.config-test -n clj-kondo.main-test` — 167 tests, 1164 assertions, 0 failures.

The four new assertions in `datalog_test.clj` are regression tests in the strict sense: applied to master unchanged they fail 4/4, and pass here.


whilo:
CI was red on `test:regression`. `clj-kondo-diff-test` lints the whole classpath, so the recorded findings in `test-regression/clj_kondo/clj_kondo/findings.edn` include datalog-parser's own sources, and bumping the dependency moves them.

The delta is 54 findings gone and 59 new, every one of them in `datalog/*`: mostly the same `unresolved-var` reports on records built by its `deftrecord` macro, now at shifted lines, plus the eight in `datalog/analysis.cljc`, which is new in the release. No finding in clj-kondo's own sources changed — I ran the same test on unmodified master first, where it passes, so the delta is attributable to the dependency alone.

Regenerated with `CLJ_KONDO_REGRESSION_UPDATE=1` and pushed as a separate commit, so it can be reviewed apart from the bump.

Issue #2955, opened 2026-08-31
type-hinting/narrowing function calls for STM objects and functions


NoahTheDuke:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/clj-kondo/clj-kondo/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

- [x] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**Is your feature request related to a problem? Please describe.**
I am switching a global object from an atom to an agent. This means that I need to change all of the `(reset! foo ...)` or `(swap! foo ...)` calls to `(send foo ...)` or `(send-off foo ...)` calls. I can use clojure-lsp or ripgrep to find instances of this, but leaving even a single one in will throw an error at runtime.

**Describe the solution you'd like**
If refs, agents, and atoms were considered distinct types and their associated functions were considered to apply only to them, then clj-kondo could lint these calls "for free".

**Describe alternatives you've considered**
Continue to rely on my eyeballs, clojure-lsp, and ripgrep.

**Additional context**
I'm willing to take a stab at it.

borkdude:
Worth a shot

PR #2958, opened 2026-09-09
Agent and Ref types

NoahTheDuke:

Implements #2955

### DIFF
--- a/test-regression/clj_kondo/clj_kondo/findings.edn
+++ b/test-regression/clj_kondo/clj_kondo/findings.edn
@@ -13662,6 +13662,15 @@
   :langs (),
   :message "Missing else branch.",
   :row 187}
+ {:end-row 208,
+  :type :constant-condition,
+  :level :warning,
+  :filename "cljs/repl/server.clj",
+  :col 22,
+  :end-col 40,
+  :langs (),
+  :message "Condition always true",
+  :row 208}
  {:end-row 21,
   :type :unused-binding,
   :level :warning,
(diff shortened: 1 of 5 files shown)


test-regression/clj_kondo/clj_kondo/findings.edn
```diff
@@ -13662,6 +13662,15 @@
   :langs (),
   :message "Missing else branch.",
   :row 187}
+ {:end-row 208,
```
borkdude, line 13665: @NoahTheDuke Can you check if these are legit findings?

NoahTheDuke: yes but i don't believe they're related to my Pr: https://github.com/clojure/clojurescript/blob/master/src/main/clojure/cljs/repl/server.clj

borkdude: Probably the type inference now knows more due to the addition of your types + fn specs

borkdude: and could therefore inference this only now?

NoahTheDuke: i'm not entirely sure why my change causes this. the type inference logic is very cool, and i can see in the code how all outputs return a map, so i trust that the finding is correct. hopefully this doesn't block the PR.

borkdude: it's fine. I'll take a look at this soon, but probably after the conj since I have a workshop to prepare :)

Issue #2963, opened 2026-09-10
New linter: detect unreachable operands in and and or


marco-carvalho:
- [ ] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**Is your feature request related to a problem? Please describe.**

A decisive constant before the final operand of `and` or `or` makes all later operands unreachable. This can hide behavior that a developer expected to run.

```clojure
(and false (save! value))
(or :done (save! value))
```

**Describe the solution you'd like**

Add a `:constant-logical-expression` linter, enabled at warning level by default, that reports a decisive non-final constant operand. For `and`, `nil` and `false` are decisive. For `or`, any compile-time truthy value is decisive. Generated forms should not warn.

**Describe alternatives you've considered**

The existing `:constant-condition` linter could be extended, but these are short-circuit expressions rather than predicate positions. A dedicated linter gives users independent control and a specific message.

**Additional context**

Issue #721 concerns constant conditions and unreachable branches, but not short-circuit operands. I am interested in submitting a focused PR with tests and documentation if this approach is accepted.

NoahTheDuke:
I don't think this needs to be a separate linter from `:constant-condition`. When would I want to enable one and disable the other? Why can't the output message from `:constant-condition` differ based on the input type?

marco-carvalho:
I agree. I do not have a compelling configuration case for enabling one and disabling the other. The current PR even has to skip condition position to avoid duplicate reports from `:constant-condition`, which is evidence that the split is artificial. Extending `:constant-condition` with a short-circuit-specific message is the better shape; the separate linter and PR can be replaced by that approach.

PR #2978, opened 2026-09-10
Add :constant-logical-expression linter

marco-carvalho:

Addresses #2963.

A decisive constant before the last operand of `and` or `or` makes every later operand unreachable, for example `(and false (save! value))`.

This adds `:constant-logical-expression`, at `:warning` by default. For `and` a `nil` or `false` operand is decisive, for `or` any compile-time truthy operand is. Generated forms do not warn.

To avoid reporting the same code twice, the linter skips calls in condition position, where `:constant-condition` already reports the short circuit. It is also turned off in `test-utils/base-config`, following the existing convention for default-on linters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


yuhan0:
This PR should be closed or marked as draft / reworked, per https://github.com/clj-kondo/clj-kondo/issues/2963#issuecomment-5622957957

Issue #2966, opened 2026-09-10
New linter: validate literal re-pattern arguments


marco-carvalho:
- [ ] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**Is your feature request related to a problem? Please describe.**

`re-pattern` accepts a string that may contain an invalid Java regular expression. A literal invalid pattern fails only when the code runs even though it can be checked statically.

```clojure
(re-pattern "[")
```

**Describe the solution you'd like**

Add an `:invalid-regex` linter, enabled at error level by default, that compiles literal string arguments to Clojure `re-pattern` during analysis and reports the pattern syntax description. Dynamic strings and ClojureScript calls should not be checked by the JVM regex engine.

**Describe alternatives you've considered**

Waiting for runtime preserves current behavior but delays a deterministic error. Applying this to ClojureScript would be unsafe because JavaScript regular expression syntax differs.

**Additional context**

I am interested in submitting a focused PR with valid, invalid, dynamic, and ClojureScript cases if this approach is accepted.

PR #2981, opened 2026-09-10
Add :invalid-regex linter

marco-carvalho:

Addresses #2966.

`(re-pattern "[")` fails only when the code runs, even though the pattern is a literal that can be compiled during analysis.

This adds `:invalid-regex`, at `:error` by default, which compiles literal string arguments to `re-pattern` and reports the `PatternSyntaxException` description. Dynamic arguments are skipped, and the check is limited to Clojure because JavaScript regex syntax differs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ For a list of breaking changes, check [here](#breaking-changes).
 
 ## Unreleased
 
+- [#2966](https://github.com/clj-kondo/clj-kondo/issues/2966): new linter `:invalid-regex` reports a literal `re-pattern` argument that is not a valid regex.
 - Bump graal-build-time to 1.0.6 to fix startup crash in binaries built with GraalVM 25.1+.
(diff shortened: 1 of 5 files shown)


NoahTheDuke:
Does compiling every single pattern in a given repository impact performance? Does relying on a try/catch for control-flow impact performance?

marco-carvalho:
The implementation only calls `Pattern/compile` for a `clojure.core/re-pattern` call whose argument is a literal string. Dynamic arguments are skipped. The try/catch is not normal control flow: valid patterns take the straight path and the catch only runs for a pattern that will become a finding. I also benchmarked linting this repository's `src` and `parser` trees in one warmed JVM, alternating the linter off and on. Warm runs were 454-644 ms off and 443-487 ms on, so there was no measurable overhead at that scale. I can add a larger synthetic benchmark if you want a bound for repositories with unusually many literal `re-pattern` calls.

NoahTheDuke:
nah that's cool, i was just wondering.

Issue #2967, opened 2026-09-10
New linter: detect redundant boolean calls in conditions


marco-carvalho:
- [ ] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**Is your feature request related to a problem? Please describe.**

Clojure condition positions already use truthiness, so wrapping an expression in `boolean` inside a condition adds no behavior and obscures the predicate.

```clojure
(if (boolean value) :yes :no)
```

**Describe the solution you'd like**

Add a `:redundant-boolean-call` linter, enabled at warning level by default, that reports `boolean` calls when the call itself is analyzed in condition position. Calls used to produce an actual boolean value outside conditions should remain valid.

**Describe alternatives you've considered**

The existing redundant-call linter could handle this, but redundancy depends on the parent context rather than argument count or identity behavior. A dedicated option avoids changing existing linter scope.

**Additional context**

I am interested in submitting a focused PR covering direct and nested condition positions, configuration, and non-condition uses if this approach is accepted.

PR #2982, opened 2026-09-10
Add :redundant-boolean-call linter

marco-carvalho:

Addresses #2967.

Condition positions already use truthiness, so `(if (boolean x) :yes :no)` adds nothing and hides the predicate.

This adds `:redundant-boolean-call`, at `:warning` by default, which reports a `boolean` call only when the call itself is analyzed in condition position. Calls that produce a real boolean value elsewhere are untouched.

### DIFF
--- a/src/clj_kondo/impl/linters.clj
+++ b/src/clj_kondo/impl/linters.clj
@@ -176,6 +176,16 @@
          (node->line (:filename ctx) keyvec :single-key-in
                      (format "%s with single key" called-name)))))))
 
+(defn lint-redundant-boolean-call [ctx call]
+  (let [expr (:expr call)]
+    (when (and (true? (:condition expr))
+               (not (utils/linter-disabled? ctx :redundant-boolean-call))
+               (not (:clj-kondo.impl/generated expr)))
+      (findings/reg-finding!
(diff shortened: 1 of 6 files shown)


src/clj_kondo/impl/linters.clj
```diff
@@ -176,6 +176,16 @@
          (node->line (:filename ctx) keyvec :single-key-in
                      (format "%s with single key" called-name)))))))
 
+(defn lint-redundant-boolean-call [ctx call]
+  (let [expr (:expr call)]
+    (when (and (true? (:condition expr))
```
NoahTheDuke, line 181: when would `(:condition expr)` be a truthy, non-`true` literal?

marco-carvalho: It can be `:no-lint`. `analyze-condition` sets the marker with `(if lint? true :no-lint)`; the latter is truthy so do/unused-value analysis still treats the expression as a condition, while condition linters must not report it. Using `true?` here intentionally excludes that sentinel.

NoahTheDuke: ah cool, thank you

Issue #2968, opened 2026-09-10
New linter: detect comparisons with NaN


marco-carvalho:
- [ ] I have read the [Clojure etiquette](https://clojure.org/community/etiquette) and will respect it when communicating on this platform.

**Is your feature request related to a problem? Please describe.**

NaN does not compare equal to itself, so direct equality and inequality expressions involving a NaN literal are misleading and commonly represent a bug.

```clojure
(= x ##NaN)
(not= x ##NaN)
(== x ##NaN)
```

**Describe the solution you'd like**

Add a `:nan-comparison` linter, enabled at warning level by default, when `=`, `not=`, or `==` contains a NaN literal. The finding should direct users toward a NaN predicate instead of comparison and apply to Clojure and ClojureScript where supported.

**Describe alternatives you've considered**

Constant folding could identify some results, but a dedicated finding explains the IEEE behavior and also catches comparisons where the other operand is dynamic.

**Additional context**

I am interested in submitting a focused PR with operand ordering and all supported comparison operators if this approach is accepted.

PR #2983, opened 2026-09-10
Add :nan-comparison linter

marco-carvalho:

Addresses #2968.

NaN is not equal to itself, so `(= x ##NaN)` is always false and usually a bug.

This adds `:nan-comparison`, at `:warning` by default, when `=`, `not=` or `==` has a NaN literal on either side. `==` is routed through the same analysis as `=` and `not=` so the check covers it.

### DIFF
--- a/src/clj_kondo/impl/analyzer.clj
+++ b/src/clj_kondo/impl/analyzer.clj
@@ -3227,6 +3227,10 @@
   (analyze-children (ctx-with-bindings ctx with-precision-bindings)
                     children))
 
+(defn- nan-node? [expr]
+  (let [value (:value expr)]
+    (and (double? value) (Double/isNaN value))))
+
 (defn- analyze-=-not= [ctx expr var-name]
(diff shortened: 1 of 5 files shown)


src/clj_kondo/impl/analyzer.clj
```diff
@@ -3280,6 +3284,13 @@
                                             :type :equals-nil
                                             :message "Prefer (nil? x) over (= nil x)"
                                             :filename (:filename ctx))))))
+    (when (and (or (nan-node? lhs) (nan-node? rhs))
+               (not (linter-disabled? ctx :nan-comparison))
+               (not (:clj-kondo.impl/generated expr)))
+      (findings/reg-finding! ctx (assoc (meta expr)
+                                        :type :nan-comparison
+                                        :message "NaN cannot be compared"
```
yuhan0, line 3292: The message should directly suggest the fix, something like
"NaN cannot be compared, use (NaN? x) instead"
"Use (NaN? x) instead of direct comparisons"

